PR #16, opened 2021-11-29
V0.1.1

### LINKED ISSUE #15: unique_edge_sizes(): debug and move to functions
* Fix the reference to edge_degree
* Move to the function.py file since this is not a core part of the data structure.

nwlandry:
* Fixed #15. Found the bug in the unique_edge_size function and moved it out of the Hypergraph class.
* Minor updates to documentation


setup.py
```diff
@@ -18,21 +18,21 @@
...
+description = "XGI is a Python library for the representation and analysis of complex systems with group (higher-order) interactions."
 
 install_requires = [
     "networkx>=2.2,<3.0",
-    "numpy>=1.15.0,<2.0",
-    "scipy>=1.1.0,<2.0",
-    "pandas>=0.23",
+    "numpy>=1.19.0,<1.22.0",
+    "scipy>=1.5.0,<1.7.0",
+    "pandas>=0.23.4,<1.0.0",
```
leotrs, line 35: Comment (non-blocking): in the future, it'd be best if the `setup.py` file just reads the contents of `requirements/default.txt` so we don't have to remember to keep them in sync.

xgi/classes/hypergraphviews.py
```diff
@@ -100,6 +99,8 @@ def subhypergraph_view(H, filtered_nodes=None, filtered_edges=None):
     newH._edge_attr = {edge: H._edge_attr[edge] for edge in newH._edge}
 
     # Add the filtered nodes with connections to the remaining edges after filtering
-    newH._node = {node: set(H._node[node]).intersection(newH._edge.keys()) for node in nodes}
+    newH._node = {
+        node: set(H._node[node]).intersection(newH._edge.keys()) for node in nodes
+    }
```
leotrs, line 104: Suggestion: we should look into using [`black`](https://black.readthedocs.io/en/stable/) or some other similar tool.

requirements/default.txt
```diff
@@ -1,5 +1,5 @@
-numpy>=1.19
-scipy>=1.5,!=1.6.1
-pandas>=1.1
-networkx>=2.5
+networkx>=2.2,<3.0
+numpy>=1.19.0,<1.22.0
+scipy>=1.5.0,<1.7.0
+pandas>=0.23.4,<1.0.0
```
leotrs, line 4: Question: is it necessary to upper bound these versions? That's usually done if a newer version of one of these libraries will break our code in some way.

xgi/classes/function.py
```diff
@@ -48,6 +50,28 @@ def degree_histogram(H):
...
+    Returns
+    -------
+    if return_counts:
+        numpy.ndarray, numpy.ndarray
+            Numpy arrays of the unique edge sizes and the number of each size respectively
+    else:
+        numpy.ndarray
+            A numpy array of the unique edge sizes
+    """
+    return np.unique(list(H.edge_size), return_counts=return_counts)
```
leotrs, line 72: Nice catch

leotrs (commented):
Looks good! One thing that comes to mind and we haven't discussed is how often do we want to make new releases. Usually, big projects in constant development have a new release every few months. If the project has slower development, then making a release for every PR might make sense.

requirements/default.txt
nwlandry: I was having trouble getting the documentation to build when I didn't, but I think I might have resolved the issue.

nwlandry: Also to your point, it would be nice to automate the up-versioning and pushing to pip

xgi/classes/hypergraphviews.py
nwlandry: I am using black but possibly the wrong version?

setup.py
nwlandry: I implemented this because it was an easy modification

requirements/default.txt
nwlandry: I ended up needing to upper bound numpy, but that's the only necessary upper bound

nwlandry:
I think I've addressed all the issues that you raised.

xgi/classes/hypergraphviews.py
leotrs: I see! What I meant is we should add a GH action that checks for black compliance on every PR.

leotrs: And we should be very careful whenever updating the version of black since that can mess up the codebase a bit.

requirements/default.txt
leotrs: > I ended up needing to upper bound numpy, but that's the only necessary upper bound

Great! Could you please document what is the problem when building the docs as a new issue? We should try to figure that out.

leotrs (approved):

PR #33, opened 2021-12-18
Fixes

### LINKED ISSUE #24: add "reverse" keyword in the convert/read bipartite edgelist
The default behavior is that the first column specifies the nodes and the second column specifies the edges. Instead of calling this method and then `H.dual()`, it would be nice if there was `reverse=True` to switch the labels. This would affect the following functions:
* read_bipartite_edgelist()
* write_bipartite_edgelist()
* from_bipartite_pandas_dataframe()

### LINKED ISSUE #27: erdos_renyi_hypergraph() : does not have n nodes
As discussed today, trying the random hypergraph generator: 
```python
import xgi
n = 10
m = 100
p = 0.01
H = xgi.erdos_renyi_hypergraph(n, m, p)

H.nodes
# NodeView((1, 3, 6, 7, 9))
```
the hypergraph has 5 nodes, not 10, even though the doc says `n` is the number of nodes. 

This discrepancy between n and the actual number of nodes is apparently expected for small n and p (correct me if I'm wrong).
If so, add a warning to the docstring. Otherwise, fix the bug.

nwlandry:
This issue closes #27 and closes #24. This PR implements the following:

* Fixes the `erdos_renyi_hypergraph()`, `chung_lu_hypergraph()`, and `dcsbm_hypergraph()` methods such that the number of nodes will always match the numberthe user specifies, even if the resulting hypergraph is disconnected.
* Changed the construction method from reading in a Pandas DataFrame to using the `add_node_to_edge()` method for a 2x speedup.
* Added some basic unit tests for the generative models.
* Added the `dual` keyword for the `read_bipartite_edgelist()` method.

### DIFF
--- a/tests/generators/test_nonuniform.py
+++ b/tests/generators/test_nonuniform.py
@@ -5,14 +5,18 @@ from xgi.exception import XGIError
 
 
 def test_erdos_renyi_hypergraph():
-    with pytest.raises(XGIError):
+    with pytest.raises(ValueError):
         H = xgi.erdos_renyi_hypergraph(10, 20, -0.1)
-    with pytest.raises(XGIError):
-        H = xgi.erdos_renyi_hypergraph(10, 20, 0.0)
-    with pytest.raises(XGIError):
-        H = xgi.erdos_renyi_hypergraph(10, 20, 1.0)
-    with pytest.raises(XGIError):
+    with pytest.raises(ValueError):
         H = xgi.erdos_renyi_hypergraph(10, 20, 2.0)
+
+    H = xgi.erdos_renyi_hypergraph(10, 20, 0.0)
+    assert H.number_of_nodes() == 10
+    assert H.number_of_edges() == 0
+    
+    H = xgi.erdos_renyi_hypergraph(10, 20, 1.0)
+    assert H.number_of_nodes() == 10
+    assert H.number_of_edges() == 20
 
     H = xgi.erdos_renyi_hypergraph(10, 20, 0.1)
     assert H.number_of_nodes() == 10
--- a/tests/generators/test_nonuniform.py
+++ b/tests/generators/test_nonuniform.py
@@ -1,0 +1,30 @@
+import xgi
+import pytest
+
+from xgi.exception import XGIError
+
+
+def test_erdos_renyi_hypergraph():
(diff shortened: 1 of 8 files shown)


xgi/readwrite/bipartite.py
```diff
@@ -117,11 +116,12 @@ def read_bipartite_edgelist(
...
+            dual=dual,
         )
 
 
 def parse_bipartite_edgelist(
-    lines, comments="#", delimiter=None, create_using=None, nodetype=None
+    lines, comments="#", delimiter=None, create_using=None, nodetype=None, dual=False
 ):
     """
     A helper function to read a iterable of strings containing a bipartite edge list and
```
maximelucas, line 127: Add description for the parameter `create_using` in docstring.

maximelucas (approved):
Looks good to me, thanks Nicholas.
Disclaimer: my experience with testing is close to zero so better if Leo double checks the tests.

tests/generators/test_nonuniform.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def test_erdos_renyi_hypergraph():
+    with pytest.raises(XGIError):
+        H = xgi.erdos_renyi_hypergraph(10, 20, -0.1)
+    with pytest.raises(XGIError):
+        H = xgi.erdos_renyi_hypergraph(10, 20, 0.0)
+    with pytest.raises(XGIError):
+        H = xgi.erdos_renyi_hypergraph(10, 20, 1.0)
+    with pytest.raises(XGIError):
```
leotrs, line 14: I think we should allow for `p` to be 0.0 or 1.0

leotrs: Also, if `p` is not within 0 and 1, we should perhaps raise `ValueError` and not `XGIError`.

leotrs (approved):
Just a couple of nitpicks :)

xgi/readwrite/bipartite.py
nwlandry: done

tests/generators/test_nonuniform.py
nwlandry: should be fixed now!

maximelucas (approved):
Good to go!

Issue #29, opened 2021-12-16
Add function to remove singleton from edges


maximelucas:
Currently, edges contain singletons (nodes) : 

```python
list(H._edge.values())
# [[0], [3, 4], [4], [4]]
```
it's often useful to have a list of edges that does not contain nodes.
Add a function to get that. 
Should the function output a new external list, of remove singletons from the edges stored internally in the HyperGraph?

leotrs:
My suggestion is to implement the following:
```python
H.edges(order=1)
# -> EdgeView([[0], [3], ...])
H.singleton_edges()
# -> same as above
H.remove_singletons()
# -> modifies H and returns None
```
Similarly, we should do something similar for removing all connected components other than the largest one:
```python
H.connected_components()
# -> list of EdgeView objects
H.connected_components(k)
# -> list of EdgeView objects containing only the k largest components
H.largest_component()
# -> same as H.connected_components(1)
H.remove_components(k=1)
# -> modifies H by removing all but the k (default k=1) largest components and returns None
```

nwlandry:
Fixed with #35

PR #35, opened 2021-12-23
Horss migration

maximelucas:
- Added small functions to the HyperGraph class
- Added generator of random hypergraph
- Added functions related to #29


xgi/generators/nonuniform.py
```diff
@@ -288,3 +290,51 @@ def dcsbm_hypergraph(k1, k2, g1, g2, omega):
...
+    Hypergraph object
+        The generated hypergraph
+
+    """
+
+    if (np.any(np.array(ps) < 0)) or (np.any(np.array(ps) > 1)):
+        raise ValueError("All elements of ps must be between 0 and 1 included.")
+
+    # I first generate a standard ER graph with edges connected with probability p1
+    G = nx.fast_gnp_random_graph(N, ps[0], seed=None)
```
nwlandry, line 322: I agree that for the pairs that this is possibly faster, but I feel like we should use the same generator that you're using for the 3-hyperedges and larger for the whole hypergraph to a) be consistent andb) avoid importing NetworkX for this function only.

xgi/generators/nonuniform.py
```diff
@@ -288,3 +290,51 @@ def dcsbm_hypergraph(k1, k2, g1, g2, omega):
...
+
+    nodes = list(G.nodes())
+    hyperedges = list(G.edges())
+
+    for i, p in enumerate(ps[1:]):
+        d = i + 2  # order (+2 because we started with [1:])
+
+        for hyperedge in combinations(nodes, d + 1):
```
nwlandry, line 330: It looks like you didn't import combinations or NetworkX. Another thing (in the future) is that for sparse hypergraphs, it probably is faster to write a function that goes from an index to a tuple (Similar to line 27 in https://github.com/nwlandry/hypergraph-assortativity/blob/main/UNUSED/GenerativeModels.py)

xgi/generators/nonuniform.py
```diff
@@ -288,3 +290,51 @@ def dcsbm_hypergraph(k1, k2, g1, g2, omega):
...
+            if random.random() <= p:
+                hyperedges.append(hyperedge)
+
+    hyperedges += [[i] for i in nodes]  # add singleton edges
+
+    H = xgi.empty_hypergraph()
```
nwlandry, line 336: I don't think that this will work if you don't import xgi.

xgi/generators/nonuniform.py
```diff
@@ -288,3 +290,51 @@ def dcsbm_hypergraph(k1, k2, g1, g2, omega):
...
+    Parameters
+    ----------
+    N : int
+        Number of nodes
+    ps : list of float
+        List of probabilities (between 0 and 1) to create a
+        hyperedge at each order d between any d+1 nodes. For example,
+        ps[0] is the wiring probability of any edge (2 nodes), ps[1]
+        of any triangles (3 nodes).
+
```
nwlandry, line 310: Perhaps add a reference to the model? One relevant paper might be https://arxiv.org/abs/1703.07686

nwlandry (changes requested):
There are a few imports broken in your Erdos-Renyi function - I might add a few tests. Overall everything looks good - very nice! Some of the imports will be fixed when the merge conflicts are fixed (The main branch has been updated since you created the horss branch) but good to fix these prior to merging nonetheless

xgi/generators/nonuniform.py
nwlandry: I would also add a check for valid p values

xgi/classes/hypergraph.py
```diff
@@ -1151,3 +1147,92 @@ def bunch_iter(nlist, nodes):
...
+        }
+
+    def remove_singleton_edges(self):
+        """Removes all singletons edges from the hypergraph"""
+
+        singleton_ids = [id_ for id_, size in dict(self.edge_size).items() if size == 1]
+        self.remove_edges_from(singleton_ids)
+        return None
+
+    def is_d_uniform(self, return_d=False):
```
leotrs, line 1202: I'm not loving the design of this function. I think it'd be cleaner with a signature like the following:
```python
is_uniform(self)
```
and make it so that if the graph is uniform, it returns `d`, and if it is not uniform, it returns something that evaluates to false (like False, 0, or None). In this way, we always return `d` and we can still use it as a boolean check: `if H.is_uniform()` always works as expected.

leotrs (changes requested):
Just one requested change, the rest looks pretty good.

lordgrilo (changes requested):
No particular comments to the latest commit by Max. The only problem I found was a merge conflict with main due to test functions being present there for non uniform HG and not on horss-migration. I believed I resolved them. This is ok for me. I'm flagging this as request changes so that Nick/leo can check the other changes they requested.

nwlandry:
Changes:
* Added a function to generate non-uniform Erdos-Renyi hypergraph.
* Added the following Hypergraph class methods:
    * max_edge_order()
    * is_possible_order()
    * singleton_edges()
    * remove_singleton_edges()
    * is_uniform()

PR #41, opened 2022-02-04
Refactor core interface

### LINKED ISSUE #40: Interface of NodeView
`NodeView` supports accessing the edge ids that contain a node in three different ways:
```python
H = some_hypergraph()
H.nodes[1] == H.nodes(1) == H.members(1)
# -> True
```
I find this unpythonic. I'd much rather there was a single way of doing things if at all possible. If there was exactly one way of accessing the edges that contain a node, *which would you prefer?* If anyone can argue that we should support two (or more) ways of doing the same thing, I'm all ears.

Related to #20.

leotrs:
This one is pretty big so I suggest reviewing each commit in turn instead of the whole thing at once. This basically implements the behavior discussed here #40. In particular, there is (hopefully, I think), one way of accessing each thing. Among other things:

* `H['attr']` accesses hypergraph attributes.
* `H.nodes()` and `H.edges()` are no longer supported
* `H.nodes[]` and `H.edges[]` now access attributes
* `H.nodes.membership` is now the same as `H.nodes.members`. (I decided to keep members for simplicity and symmetry with `H.edges`)

There are a few kinks to work out, but I thought this PR was already big enough.

**As you review, please let me know if you think of any other test I should add to this PR.**

Also: there's a lot of noise in this PR because I decided to change `*_attrs` to `*_attr` just to be uniform with the variable names in the main `Hypergraph` class.

Closes #40

### DIFF
--- a/tests/classes/test_hypergraph.py
+++ b/tests/classes/test_hypergraph.py
@@ -1,5 +1,6 @@
 import pytest
 import xgi
+from xgi.exception import XGIError
 
 
 def test_constructor(edgelist5, dict5, incidence5, dataframe5):
@@ -169,7 +170,7 @@ def test_add_node_attr(edgelist1):
 
 def test_hypergraph_attr(edgelist1):
     H = xgi.Hypergraph(edgelist1)
-    with pytest.raises(KeyError):
+    with pytest.raises(XGIError):
         H['color']
     H['color'] = 'red'
     assert H['color'] == 'red'
--- a/tests/classes/test_hypergraph.py
+++ b/tests/classes/test_hypergraph.py
@@ -173,3 +173,12 @@ def test_hypergraph_attr(edgelist1):
         H['color']
     H['color'] = 'red'
     assert H['color'] == 'red'
+
+
+def test_members(edgelist1):
+    H = xgi.Hypergraph(edgelist1)
+    assert H.nodes.members(1) == H.nodes.membership(1) == [0]
+    assert H.nodes.members(2) == H.nodes.membership(2) == [0]
+    assert H.nodes.members(3) == H.nodes.membership(3) == [0]
+    assert H.nodes.members(4) == H.nodes.membership(4) == [1]
+    assert H.nodes.members(6) == H.nodes.membership(6) == [2, 3]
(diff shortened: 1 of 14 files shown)


nwlandry:
I simplified the view classes and removed the data views.

xgi/classes/reportviews.py
```diff
@@ -171,104 +148,38 @@ def __getitem__(self, n):
...
-                f"{type(self).__name__} does not support slicing, "
-                f"try list(H.nodes)[{n.start}:{n.stop}:{n.step}]"
-            )
-        elif n not in self._nodes:
-            raise XGIError(f"The node {n} is not in the hypergraph")
-
-        return self._node_attr[n]
+        try:
+            return self._id_attrs[id]
+        except:
```
leotrs, line 153: Comment: it's generally a bad idea to catch every single Exception using a naked `except`. It's always recommended to specify the exception type. In this case it'd be `except KeyError` (I believe).

tests/classes/test_hypergraph.py
```diff
@@ -155,3 +155,12 @@ def test_isolates(edgelist1):
     assert H.isolates() == {4}
     H.remove_isolates()
     assert 4 not in H
+
+
+def test_add_node_atr(edgelist1):
```
maximelucas: atr -> attr in function name?

maximelucas (approved):

xgi/classes/hypergraph.py
```diff
@@ -162,19 +155,9 @@ def __len__(self):
...
-        -------
-        list
-           A list of the edges of which the specified node is a part.
-
-        Notes
-        -----
-        H[n] is the same as H.nodes[n]
-        """
-        return self._node[n]
+    def __getitem__(self, attr):
```
maximelucas, line 158: Do we have a test for this new behaviour?

tests/classes/test_hypergraph.py
```diff
@@ -173,3 +173,12 @@ def test_hypergraph_attr(edgelist1):
         H['color']
     H['color'] = 'red'
     assert H['color'] == 'red'
+
+
+def test_members(edgelist1):
```
maximelucas, line 179: Add a line to test when the node does not exist, and when it's not a member of any edge?
assert H.nodes.membership(0) == ?

H.add_node(0)
assert H.nodes.membership((0) == ?

nwlandry: Fixed. Now it returns an XGIError.

PR #46, opened 2022-02-18
More tests

nwlandry:
Added unit tests for the following files:
* matrix.py (Improved coverage from 77% to 95%)
* hypergraphviews.py (Improved coverage from 17% to 96%)
* function.py (Improved coverage from 60% to 84%)

Overall coverage improved from 61% to 66%.

Tests before:
test_before

Tests after:
test_after

Also made small fixes to the documentation.

### DIFF
--- a/tests/classes/test_function.py
+++ b/tests/classes/test_function.py
@@ -86,9 +86,129 @@ def test_create_empty_copy(edgelist1):
     assert E2.shape == (8, 0)
     for node in E2.nodes:
         assert len(E1.nodes.memberships(node)) == 0
-    assert E2._hypergraph == {"name": "test", "timestamp": "Nov. 20"}
-    for n in H.nodes:
-        assert H.nodes[n]["name"] == attr_dict[n]["name"]
+    assert E2["name"] == "test"
+    assert E2["timestamp"] == "Nov. 20"
+    with pytest.raises(XGIError):
+        author = E2["author"]
+    for n in E2.nodes:
+        assert E2.nodes[n]["name"] == attr_dict[n]["name"]
+
+
+def test_set_node_attributes(edgelist1):
+    attr_dict1 = {
+        1: {"name": "Leonie"},
+        2: {"name": "Ilya"},
+        3: {"name": "Alice"},
+        4: {"name": "Giovanni"},
+        5: {"name": "Heather"},
+        6: {"name": "Juan"},
+        7: {"name": "Nicole"},
+        8: {"name": "Sinan"},
+    }
+
+    attr_dict2 = {
+        1: "Leonie",
+        2: "Ilya",
+        3: "Alice",
+        4: "Giovanni",
+        5: "Heather",
+        6: "Juan",
+        7: "Nicole",
+        8: "Sinan",
+    }
+
+    H1 = xgi.Hypergraph(edgelist1)
+    xgi.set_node_attributes(H1, attr_dict1)
+
+    for n in H1.nodes:
+        assert H1.nodes[n]["name"] == attr_dict1[n]["name"]
+
+    H2 = xgi.Hypergraph(edgelist1)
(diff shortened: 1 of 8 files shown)


tests/linalg/test_matrix.py
```diff
@@ -51,43 +52,55 @@ def test_incidence_matrix(edgelist1, edgelist3, edgelist4):
...
+    assert I4[node_dict1[8], edge_dict1[3]] == 1
+
+    I5, node_dict, edge_dict = xgi.incidence_matrix(
         H1, index=True, weight=lambda node, edge, H: 3
     )
     node_dict4 = {k: v for v, k in node_dict.items()}
     edge_dict4 = {k: v for v, k in edge_dict.items()}
-    assert I4[node_dict4[1], edge_dict4[0]] == 3
+    assert I5[node_dict4[1], edge_dict4[0]] == 3
+    assert np.unique(np.ravel(I5.todense())).tolist() == [0, 3]
```
maximelucas, line 75: break into two lines for readability?

tests/classes/test_function.py
```diff
@@ -86,9 +86,129 @@ def test_create_empty_copy(edgelist1):
...
+    assert E2["name"] == "test"
+    assert E2["timestamp"] == "Nov. 20"
+    with pytest.raises(XGIError):
+        author = E2["author"]
+    for n in E2.nodes:
+        assert E2.nodes[n]["name"] == attr_dict[n]["name"]
+
+
+def test_set_node_attributes(edgelist1):
+    attr_dict1 = {
```
maximelucas, line 98: add test for what happens if trying to set attr for a node not in the hypergraph?
attr_dict = {0 : {"name" : "Nobody"}}
From the function itself it looks like no error is raised, should we raise one? (Same for the set_edge_attr)

tests/classes/test_function.py
```diff
@@ -86,9 +86,129 @@ def test_create_empty_copy(edgelist1):
...
+        3: {"name": "Alice"},
+        4: {"name": "Giovanni"},
+        5: {"name": "Heather"},
+        6: {"name": "Juan"},
+        7: {"name": "Nicole"},
+        8: {"name": "Sinan"},
+    }
+    xgi.set_node_attributes(H1, attr_dict)
+
+    assert xgi.get_node_attributes(H1, "name") == {id : data["name"] for id, data in attr_dict.items()}
```
maximelucas: This still works if we set the name of a subset of the nodes I think, right?
LIke attr_dict = { 1: {"name": "Leonie"}}

tests/classes/test_function.py
```diff
@@ -86,9 +86,129 @@ def test_create_empty_copy(edgelist1):
...
+    assert xgi.get_node_attributes(H1, "weight") == dict()
+    
+
+def test_set_edge_attributes(edgelist1):
+    H1 = xgi.Hypergraph(edgelist1)
+    attr_dict1 = {
+        0: {"weight": 1},
+        1: {"weight": 2},
+        2: {"weight": 3},
```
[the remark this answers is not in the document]
nwlandry: Done.

tests/linalg/test_matrix.py
nwlandry: Done.

tests/classes/test_function.py
nwlandry: I'm torn on this, but my inclination is that it's okay if it fails silently. Like if you had a giant list of attribute data some of which corresponded to nodes in the hypergraph and others which didn't (For instance the Project 90 network data set), you would potentially have to remove the attributes connected to non-existent nodes for it to run.

maximelucas: Okay, fair enough.

maximelucas (approved):

PR #50, opened 2022-02-23
Frozen hypergraphs

### LINKED ISSUE #31: Create a frozen hypergraph object
We have discussed implementing a frozen object that would allow every statistic to be cached:
```python
H = some_hyper_graph()

H.adj_tensor()    # compute the tensor
# ...more code here...
H.adj_tensor()    # has to recompute

H.freeze()
H.adj_tensor()    # since H is frozen, it computes the tensor once and *caches it*
# ...more code here...
H.adj_tensor() # no need to recompute!
```

The point is that after `H` is frozen, we can essentially cache everything!
```python
H.freeze()
H.max_degree() # compute and cache
H.degree_histogram() # compute and cache
# etc
```

If at any point the user wants to modify `H` again, the cache is flushed,
```python
H.thaw() # everything that was cached is erased
H.add_edge([1,2,3])
H.max_degree() # need to recompute
H.freeze()
H.max_degree() # compute and cache
```

leotrs:
Welp, here it is. The results speak for themselves:

```python

>>> import xgi
>>> H = xgi.FrozenHypergraph([[0, 1, 2], [1, 2, 3], [2, 3, 4], [4]])
>>> H.is_uniform()
2

>>> %timeit H.is_uniform()
2.78 µs ± 48.8 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

>>> H.freeze()

>>> %timeit H.is_uniform()
296 ns ± 1.68 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

This particular method is already 10x faster on this particular (small!) hypergraph.

A couple of notes:
1. I think the implementation is an abomination and I'm ashamed of myself for pushing this publicly.
2. Ok not so much, but honestly the thing is not pretty and I can already tell it'll be a mess to debug.
3. I've only implemented one method for now to get yall's opinions before merging.

PLEASE DO NOT MERGE THIS YET. Feel free to review and leave comments.

Closes #31.

### DIFF
--- a/xgi/classes/__init__.py
+++ b/xgi/classes/__init__.py
@@ -1,4 +1,5 @@
 from .hypergraph import Hypergraph
+from .frozen import FrozenHypergraph
 from .function import *
 from xgi.classes import hypergraphviews
 from xgi.classes import reportviews
--- a/xgi/classes/frozen.py
+++ b/xgi/classes/frozen.py
@@ -1,0 +1,46 @@
+"""Class for a hypergraph that can be frozen."""
+
+from functools import lru_cache
+
+from xgi.classes import Hypergraph
+
+__all__ = ["FrozenHypergraph"]
+
+
+def cache_when_frozen(method):
+
+    @lru_cache
+    def cached(self, *args, **kwargs):
+        return method(self, *args, **kwargs)
+
+    def wrapper(self, *args, **kwargs):
+        if self.frozen:
+            return cached(self, *args, **kwargs)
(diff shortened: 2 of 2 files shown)


leotrs:
Wait I just realized the output of `is_uniform` is wrong LOL it should be `False`....

leotrs:
I think this approach is not great and will keep thinking about this. I'm deleting this branch for now.

PR #51, opened 2022-02-25
Unit tests for the read/write module

nwlandry:
Added unit tests for the following files in the readwrite module.
* edgelist.py: Increased coverage from 14% to 94%. Also changed API to remove support for weighted/attributed edgelists.
* bipartite.py: Increased coverage from 19% to 89%. Added the ability to specify edgetype.
* incidence.py: Increased coverage from 0% to 100%.
* json.py: Increased coverage from 28% to 75%. Also changed the API and JSON structure.

Overall coverage increased from 66% to 75%.

Tests before:
tests_before

Tests after:
tests_after


leotrs (approved):
All good! It's always great to have more tests. I found only a couple of nitpics, but in the interest of efficiency I won't point them out. Sorry the review took so long.

Issue #47, opened 2022-02-18
Add random seed option in the random hypergraph generators


maximelucas:
For reproducibility, it would be nice to have a "random_seed" argument that can be set in the generators.

iaciac:
We could do the same as in networkx right? https://networkx.org/documentation/stable/reference/randomness.html#randomness

maximelucas:
Yea agreed. It seems they use this decorator `@py_random_state(2)` and generate a random number with `seed.random()`.
The decorator is implemented here: https://github.com/networkx/networkx/blob/main/networkx/utils/decorators.py.

leotrs:
@maximelucas if #56 fully resolved this issue please close it :)

PR #56, opened 2022-03-14
Random seed

maximelucas:
Added a `random_seed` parameter to `random_hypergraph()`, the networkx way.
To solve #47.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -138,5 +138,8 @@ dmypy.json
 # Cython debug symbols
 cython_debug/
 
+# Macos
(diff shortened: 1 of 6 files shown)


leotrs (approved):
I don't particularly love the idea of copy/pasting an entire file... but I guess it's fine for now?

nwlandry:
Looks good! Do you have an idea of the speed comparisons for the decorator vs. the old method? I assume it's the same?

nwlandry (approved):

PR #59, opened 2022-03-15
Equine migration

### LINKED ISSUE #1: Add support for simplicial complexes
Because one can infer all the sub-interactions given a simplex, perhaps it is beneficial to store a simplicial complex using it's maximal edges to save space and perhaps use to develop efficient algorithms.

maximelucas:
- created the `SimplicialComplex` class:
  - uses "simplices" and not "edges" in the naming of the add/remove methods (inherited "edge" methods raise an error)
  - a simplex is stored as a frozenset to ensure uniqueness and undirectedness
  - `add_simplex`-like methods also add all subfaces 
  - `remove_simplex`-like methods also remove "supfaces", i.e. simplices of higher orders containing the removed simplex
- added to linalg functions:
  - added `order` and `weighted` arguments to the `incidence_matrix` and `adjacency_matrix` functions
  - added a `intersection_profile` function
  - added a `laplacian` function with argument `order` 
- added 2 generators of random (maximal) simplcial complexes, and toy star-clique generator

I dont' have tests yet for most of those, but it feels big enough to review already. 

Closes #1


tests/generators/test_classic.py
```diff
@@ -1,6 +1,21 @@
...
+        H = xgi.star_clique(-1, 7, 3)
+    with pytest.raises(ValueError):
+        H = xgi.star_clique(6, -1, 3)
+    with pytest.raises(ValueError):
+        H = xgi.star_clique(6, 7, -1)
+    with pytest.raises(ValueError):
+        H = xgi.star_clique(6, 7, 7)
+
+    H = xgi.star_clique(6, 7, 3)
+    assert H.num_nodes == 13
```
leotrs, line 21: Suggestion: this test seems a bit weak - can you please add a few more asserts? Like for the number of simplices, maximal order, etc.

tests/classes/test_simplicialcomplex.py
```diff
@@ -0,0 +1,3 @@
+import pytest
+import xgi
+from xgi.exception import XGIError
```
leotrs, line 3: We should probably try to add a couple of tests before merging :grin:

xgi/classes/hypergraph.py
```diff
@@ -1152,6 +1154,35 @@ def remove_isolates(self, ignore_singletons=True):
...
+        remove_isolates
+        remove_singleton_edges
+        """
+        
+        edges = [tuple(e) for e in self._edge.values()]
+        edges_unique = set(edges) 
+        
+        self._edge.clear() # remove all edges
+        self._edge_attr.clear()
+        self.add_edges_from(edges_unique) # re-add only unique ones
```
leotrs, line 1172: Suggestion:
```suggestion
        edges_unique = {tuple(e) for e in self._edge.values()}
        self._edge.clear()
        self._edge_attr.clear()
        self.add_edges_from(edges_unique)
```

leotrs: Also, I believe this is going to destroy all attributes and never add them back again! You'd have to filter the attributes we need to keep:
```python
...
attrs_unique = {id: self._edge_attr[id] for id in edge_unique}
..
self.add_edges_from(edges_unique, attrs_unique)
```

leotrs: Having said that, we really need a test for this that keeps track of the attributes correctly...

xgi/classes/simplicialcomplex.py
```diff
@@ -0,0 +1,332 @@
...
+    #             if size>=3 : # because we already checked nodes 
+    #                 for face in combinations(edge, size-1) : # check if all subfaces are present
+
+    #                     face = tuple(sorted(face))
+    #                     if face not in edges : 
+    #                         print(f"{face} not in SimplicialComplex (first found)")
+    #                         return False
+            
+    #         else : 
+    #             return True
```
leotrs, line 49: Pro-tip: leaving commented out code is usually frowned upon -- that's what version control is for! :)

xgi/generators/classic.py
```diff
@@ -45,3 +45,62 @@ def empty_hypergraph(create_using=None, default=Hypergraph):
...
+    nodes_star = range(n_star)
+    nodes_clique = range(n_star, n_star + n_clique)
+    nodes = list(nodes_star) + list(nodes_clique)
+
+    H = empty_hypergraph()
+    H.add_nodes_from(nodes)
+
+    # add star edges (center of the star is 0-th node)
+    for i in range(1, n_star):
+        H.add_edge((nodes_star[0], nodes_star[i]))
```
leotrs, line 96:
```suggestion
H.add_edges_from([(nodes_star[0], nodes_star[i]) for i in range(1, n_star)])
```

xgi/generators/classic.py
```diff
@@ -45,3 +45,62 @@ def empty_hypergraph(create_using=None, default=Hypergraph):
...
+
+    # connect clique and star by adding last star leg
+    H.add_edge((nodes_star[0], nodes_clique[0]))
+
+    # add clique hyperedges up to order d_max
+    for d in range(1, d_max):
+        for hyperedge in combinations(nodes_clique, d + 1):
+            H.add_edge(hyperedge)
```
leotrs, line 104:
```suggestion
H.add_edges_from([e for d in range(1, d_max) for e in combinations(nodes_clique, d + 1)])
```

xgi/classes/simplicialcomplex.py
```diff
@@ -0,0 +1,332 @@
...
+            Each edge id given in the list or container will be removed
+            from the hypergraph.
+
+        See Also
+        --------
+        remove_simplex_id : remove a single simplex by ID
+
+        Notes
+        -----
+        Will fail silently if an edge in ebunch is not in the simplicial complex.
```
leotrs, line 298: It seems inconsistent to me that `remove_simplex_ids_from` fails silently but `remove_simplex_id` raises an exception. I think the two of them should have the same behavior. Perhaps both raise a warning (not exception)?

leotrs (changes requested):

xgi/classes/simplicialcomplex.py
```diff
@@ -0,0 +1,332 @@
...
+        raise XGIError("Cannot add_edges_from to SimplicialComplex, use add_simplices_from instead")
+
+    def add_weighted_edges_from(self, ebunch_to_add, weight="weight", **attr):
+        raise XGIError("Cannot add_weighted_edges_from to SimplicialComplex, use add_weighted_simplices_from instead")
+
+    def remove_edge(self, id):
+        raise XGIError("Cannot remove_edge to SimplicialComplex, use remove_simplex instead")
+
+    def remove_edges_from(self, ebunch):
+        raise XGIError("Cannot remove_edges_from to SimplicialComplex, use remove_simplices_from instead")
```
iaciac, line 85: These are a bit hilarious, but surely useful :)

xgi/generators/nonuniform.py
```diff
@@ -285,3 +286,106 @@ def random_hypergraph(N, ps):
...
+    simplices = []  
+
+    for i, p in enumerate(ps):
+        d = i + 1  # order, ps[0] is prob of edges (d=1)
+
+        for simplex in combinations(nodes, d + 1):
+            if random.random() <= p:
+                simplices.append(simplex)
+
+    simplices += [[i] for i in nodes]  # add singleton edges
```
iaciac, line 341: This can be the initial one, instead of the empty one.

xgi/generators/nonuniform.py
```diff
@@ -285,3 +286,106 @@ def random_hypergraph(N, ps):
...
+
+    S = xgi.Simplicialcomplex()
+    S.add_nodes_from(nodes)
+    S.add_edges_from(simplices)
+
+    return H
+
+def random_maximal_simplicial_complex_d2(N, p, seed=None) : 
+    """Generate maximal simplicial complex from graph,
+    up to order 2, by filling all empty triangles.
```
iaciac, line 351:
```suggestion
    """Generate a maximal simplicial complex (up to order 2) from a $G_{N,p}$ Erdős-Rényi random graph by filling all empty triangles with 2-simplices.
```

xgi/generators/nonuniform.py
```diff
@@ -285,3 +286,106 @@ def random_hypergraph(N, ps):
...
+    
+    G = nx.fast_gnp_random_graph(N, p, seed=seed)
+    
+    nodes = G.nodes()
+    edges = list(G.edges()) 
+    
+    # compute all triangles to fill
+    all_cliques = list(nx.enumerate_all_cliques(G))
+    triad_cliques = [tuple(x) for x in all_cliques if len(x)==3]
+    triad_cliques = triad_cliques
```
iaciac, line 383: ```triad_cliques = triad_cliques```
What is this for?

maximelucas: Oops. Had a sorting function there, but removed it without removing the line.

xgi/classes/simplicialcomplex.py
maximelucas: Forgot it there :)

xgi/generators/classic.py
maximelucas: Done

tests/generators/test_classic.py
maximelucas: Done

xgi/classes/hypergraph.py
maximelucas: Very good point, I realised that when writing it, and then never solved it. I wrote it originally to use it in the SimplicialComplex class, but in the end we use frozensets so we don't need it anymore. Deleting for now.

xgi/classes/simplicialcomplex.py
maximelucas: Fair point. I actually copied the behaviour we had for those in the Hypergraph class. I think Nicholas' idea of not raising a exception when removing a bunch that you get from a dataset, and there might be many that are not in the Hypergraph, but raising many warnings/exceptions might be a pain. But when you remove a single one I guess you're expected to know which one it is. I don't have a strong opinion. What do you think?

tests/classes/test_simplicialcomplex.py
maximelucas: I'll to add some in the coming days, if anyone feels inspired to write some too, feel free 😁

maximelucas: Okay, 17 hours later.. done.

leotrs (approved):

xgi/classes/simplicialcomplex.py
iaciac: I see the problem. Then I think this solution makes sense. Writing it down explicitly (as it is at the moment in the note) seems fair.

xgi/generators/classic.py
```diff
@@ -45,3 +45,61 @@ def empty_hypergraph(create_using=None, default=Hypergraph):
         # try create_using as constructor
         H = create_using()
     return H
+
+
+def star_clique(n_star, n_clique, d_max):
+    """Generate a star-clique structure
+
+    That is a star network and a clique network,
+    connected by one pairwise edge. In the clique
```
iaciac, line 54:
```suggestion
    connected by one pairwise edge connecting the centre of the star to the clique.
```

leotrs (approved):

xgi/classes/simplicialcomplex.py
maximelucas: Let's leave it for now then!

PR #60, opened 2022-03-16
Docs n' stuff

### LINKED ISSUE #53: Add tutorial/example notebooks to test suite

leotrs:
I made a pass on the documentation. The main changes are the following:

1. The contents of "about" page have been moved to the landing page.
1. We are now using `autodoc` and `autosummary` to autogenerate documentation and files.
1. Now every package, module, and class lives in its own webpage. Before we had all clases/functions/exceptions of a module inside a single webpage and we were linking to modules directly (bypassing packages). So now the organization of the documentation reflects more closely the organization of the codebase.
1. The landing page contains links to each of the package webpages, each of which contains a list of their submodules. The documentation page of each submodule contains a list of all the contained classes.
1. Since reaching e.g. the main documentation page for say the `Hypergraph` class now takes a few clicks, I added a new "quick links" section to the landing page.
1. I added stub (empty) files for a few tutorials I want to write. They are also now linked in the landing page.

Closes #53.


leotrs:
BTW this is not working because I'm dumb and I didn't pay attention to the hypergraph view class. Marking as draft for now..

leotrs:
Alright this is big but mostly documentation changes. I also took the liberty to profile/refactor some methods and add some tests. Sorry for the mess..

xgi/algorithms/connected.py
```diff
@@ -35,11 +35,12 @@ def is_connected(H):
...
     -------
-        >>> import xgi
-        >>> n = 1000
-        >>> ps = [0.01, 0.001]
-        >>> H = xgi.random_hypergraph(n, ps)
-        >>> print(xgi.is_connected(H))
+    >>> import xgi
+    >>> n = 1000
+    >>> ps = [0.01, 0.001]
+    >>> H = xgi.random_hypergraph(n, ps)
```
nwlandry, line 41: Wondering if we want to use the seed option that @maximelucas implemented so we can assure consistent results.

xgi/classes/hypergraph.py
```diff
@@ -351,33 +367,20 @@ def remove_nodes_from(self, nodes):
...
-                del self._node_attr[n]
-                for edge in edge_neighbors:
-                    self._edge[edge].remove(n)  # remove all edges n-u in graph
-                    # delete empty edges
-                    if len(self._edge[edge]) == 0:
-                        del self._edge[edge]
-                        del self._edge_attr[edge]
-            except KeyError as e:
-                pass
+            self.remove_node(n)
```
nwlandry, line 379: This used to fail silently, so maybe in the for loop have
```
for n in nodes:
    if n in self._node:
        self.remove_node(n)
```

xgi/classes/hypergraph.py
```diff
@@ -652,18 +642,13 @@ def remove_edges_from(self, ebunch):
...
         """
         for id in ebunch:
-            try:
-                for node in self.edges.members(id):
-                    self._node[node].remove(id)
-                del self._edge[id]
-                del self._edge_attr[id]
-            except:
-                pass
+            self.remove_edge(id)
```
nwlandry, line 648: same comment as for `remove_nodes_from`

docs/source/conf.py
```diff
@@ -37,6 +37,11 @@
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ["_templates"]
 
+# copybutton options
```
nwlandry, line 40: Can you update the author and copyright?

nwlandry (changes requested):
Minor changes. This all looks great. Can you either add header docstrings for the readwrite, algorithms, and generators, and linear algebra or make an issue so we can take care of it later?

docs/source/conf.py
leotrs: Done

leotrs:
> Minor changes. This all looks great. Can you either add header docstrings for the readwrite, algorithms, and generators, and linear algebra or make an issue so we can take care of it later?

Done!

xgi/classes/hypergraph.py
nwlandry: That makes sense. I like the idea of throwing a warning in an `else` statement so that the method still runs, but the user is aware that there might be missing nodes. I'm fine with whatever you decide, though.

nwlandry (approved):
This all looks good. I'm approving this PR because I'm fine with whatever you decide to do about `remove_nodes_from` and `remove_edges_from`.

leotrs:
Done. Now this is the behavior:
```python
H = xgi.Hypergraph()
H.remove_node(10)
XGIError: The node 10 is not in the graph.
H.remove_nodes_from([10])
UserWarning: Node 10 not in hypergraph
```
I guess I can see how this would be helpful, though it will perhaps be changed at some point, especially if/when #81 gets merged. But let's leave it like this for now

PR #66, opened 2022-03-23
Return an empty array rather than a 1x1 zero array when appropriate

### LINKED ISSUE #61: adjacency_matrix error when no edges (of order d)
Running 
```python
import xgi
H = xgi.Hypergraph()
H.add_edges_from([[0, 1, 2], [1, 2, 3], [2, 3, 4]]) #edgelist6

A, node_dict = xgi.adjacency_matrix(H, order=1, index=True)
```
returns
```
File ~/Dropbox (ISI Foundation)/WORK/SCIENCE/xgi/xgi/linalg/matrix.py:136, in adjacency_matrix(H, order, s, weighted, index)
    133     I = incidence_matrix(H, index=False, order=order)
    135 A = I.dot(I.T)
--> 136 A.setdiag(0)
    138 if not weighted:
    139     A = (A >= s) * 1

AttributeError: 'numpy.float64' object has no attribute 'setdiag'
```

That happens because, there is no edge of order 1 in the hypergraph. In that case, right now, we have
```python 
xgi.incidence_matrix(H, order=1, index=True)
# array([0])
```
so that `I.dot(I.T) == 0` which is an int which has no diagonal to set. 

To fix this, two options:
1. raise an error in incidence_matrix when there is no edge or no nodes, instead of returning [0].
2. have a `if else` in `adjacency_matrix` to return an NxN array of zeros if `I == [0]`. Probably need to add `sparse` argument there too then to deal with both cases coming from `incidence_matrix`.

What do you think? Do we want to return [0] for the incidence matrix when there are no edges/nodes or not?

leotrs:
Fixes #61.

### DIFF
--- a/xgi/linalg/matrix.py
+++ b/xgi/linalg/matrix.py
@@ -47,11 +47,14 @@ def incidence_matrix(
         >>> ps = [0.1, 0.01]
         >>> H = xgi.random_hypergraph(N, ps)
         >>> I = xgi.incidence_matrix(H)
-    """
 
+    """
     edge_ids = H.edges
     if order is not None:
         edge_ids = [id_ for id_, edge in H._edge.items() if len(edge) == order + 1]
+    if not edge_ids:
+        return (np.array([]), {}, {}) if index else np.array([])
+
     node_ids = H.nodes
     num_edges = len(edge_ids)
     num_nodes = len(node_ids)
@@ -59,7 +62,7 @@ def incidence_matrix(
     node_dict = dict(zip(node_ids, range(num_nodes)))
     edge_dict = dict(zip(edge_ids, range(num_edges)))
 
-    if len(node_dict) != 0 and len(edge_dict) != 0:
+    if node_dict and edge_dict:
 
         if index:
             rowdict = {v: k for k, v in node_dict.items()}
@@ -67,9 +70,9 @@ def incidence_matrix(
 
         if sparse:
             # Create csr sparse matrix
-            rows = list()
-            cols = list()
-            data = list()
+            rows = []
+            cols = []
+            data = []
             for edge in edge_ids:
                 members = H.edges.members(edge)
                 for node in members:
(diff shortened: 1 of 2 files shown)


xgi/linalg/matrix.py
```diff
@@ -125,13 +128,16 @@ def adjacency_matrix(H, order=None, s=1, weighted=False, index=False):
...
-    """
 
+    """
     if index:
         I, rowdict, _ = incidence_matrix(H, index=True, order=order)
     else:
         I = incidence_matrix(H, index=False, order=order)
 
+    if I.shape == (0,):
+        return (np.array([]), {}) if index else np.array([])
```
maximelucas, line 139: This works if we have no nodes, but when we have nodes but no edges (of order d), we want to return np.zeros((N,N)).

```
if I.shape == (0,):
    if not rowdict:
        A = np.array([])
    if not coldict:
        A = np.zeros((H.num_nodes, H.num_nodes))
    return (A, {}) if index else A 

```

xgi/linalg/matrix.py
```diff
@@ -248,9 +254,12 @@ def laplacian(H, order=1, rescale_per_node=False, index=False):
     .. [1] Lucas, M., Cencetti, G., & Battiston, F. (2020).
         Multiorder Laplacian for synchronization in higher-order networks.
         Physical Review Research, 2(3), 033410.
-    """
 
+    """
     A, row_dict = adjacency_matrix(H, order=order, weighted=True, index=True)
+    if A.shape == (0,):
```
maximelucas, line 260: similar problem here i think

xgi/linalg/matrix.py
leotrs: Good catch!

xgi/linalg/matrix.py
leotrs: In this case I think it's actually fine, because even when the adj matrix is all zeros, you can compute the Laplacian in the same way as before, so you only need to check whether it's empty.

maximelucas:
I realise `clique_motif_matrix` might need the same update (but not `intersection_profile`?).

leotrs:
@maximelucas done!

maximelucas (approved):

PR #68, opened 2022-03-23
Add weights to layout + little fix

iaciac:
1. New `weighted_barycenter_spring_layout` similar to the `barycenter_spring_layout`, but where weights are added in order to keep the hyperedges closer together. 
2. Little fix on the numbering of the phantom nodes in the `barycenter_spring_layout` that allows for node labels different from {0,...,N-1}.


xgi/drawing/layout.py
```diff
@@ -99,7 +100,7 @@ def barycenter_spring_layout(H, return_phantom_graph=False):
         G.add_edge(i, j)
     
     # Adding phantom nodes and connections therein
-    phantom_node_id = H.number_of_nodes() #the first id is N (the first node is 0)
+    phantom_node_id = max(H.nodes)+1
```
maximelucas, line 103: H.num_nodes

xgi/drawing/layout.py
```diff
@@ -115,6 +116,58 @@ def barycenter_spring_layout(H, return_phantom_graph=False):
...
+    
+    # Adding real nodes
+    G.add_nodes_from(list(H.nodes))
+    
+    # Adding links (edges composed by two nodes only, for which we don't use phantom nodes)
+    for i, j in H.edges_of_order(1).values():
+        G.add_edge(i, j, weight=1*5)
+    
+    # Adding phantom nodes and connections therein
+    phantom_node_id = max(H.nodes)+1
```
maximelucas, line 155: H.num_nodes

xgi/drawing/layout.py
iaciac: What if nodes are not numbered from 0 to N-1? Let's say your nodes are {1,2,4}, so the number of nodes is 3, then H.num_nodes+1 (the ID of the phantom node) will be 4 which is already in use.
I'm open to better solutions :)

maximelucas:
In `weighted_barycenter_spring_layout`, the default weight 5 is hardcoded and used twice. 
Do we want to have a variable for it? Do we want to include it as an argument with default value 5?

Also, now the scaling factor for weights of hyperedges is `d`, we could pass it as an argument to allow flexibility for the user?

xgi/drawing/layout.py
maximelucas: Ah yea good point, my bad :) Is it possible that an edge already ID `H.num_nodes+1' and that it clashes? I don't remember how we deal with IDs from nodes and edges. In any case, forget my suggestion hehe.

maximelucas:
In `random_layout`, the example includes `H = nx.stupid_hypergraph()`. I guess it should be xgi. not nx.? Same for the next line of the example. Also, if we don't actually have that function, I would maybe replace with the `xgi.random_hypregraph` that we often used in the examples. 

You use `@random_state(3)` to deal with the seed. I guess it works but it's different from what we used in the random hypergraph generators. Better to choose one or the other for consistency I guess.

leotrs (approved):
Looks good to me! We'll have to remember to have some notebooks with examples and so on.

maximelucas:
Looks great!

maximelucas (approved):

PR #69, opened 2022-03-23
Filter edges by order and nodes by degree

leotrs:
This is an addition to the view interface. Essentially, now we can do the following:
```python
H = xgi.Hypergraph([[1, 2, 3], [3, 4], [4, 5, 6]])

# filter nodes by degree
H.nodes
# -> NodeView((1, 2, 3, 4, 5, 6))
H.nodes(degree=2)     # awesome new stuff
# -> NodeView((3, 4))

# filter edges by order
H.edges
# -> EdgeView((0, 1, 2))
H.edges(order=2)     # whoa
# -> EdgeView((0, 2))

# awesome stuff happens
1 in H.edges
# -> True
1 in H.edges(order=2)     # magic
# -> False
```

Using `H.nodes(degree)` or `H.edges(order)` return view objects that contain only a subset of nodes or edges. These are still instances of the same old view classes. The difference is that now `IDView` contains a list of IDs that it keeps track of. So it will always contain the entire dictionary and attributes of all nodes/edges, just like before, but now whenever it reads from them, it uses the provided subset of IDs that it is keeping track of. This is probably going to be a pain in the ass later on because whenever we add or change functionality of `IDView`, we will need to remember to use the subset of ids, and not the keys of the stored dictionary which would be more natural.

One technical comment is that I had to implement the class method `IDView.from_view` to create a new view from a different view. This is because the constructor of `NodeView` and `EdgeView` take only a hypergraph, so we didn't have a way of creating a view from a view.

One drawback of the current implementation is that the views no longer preserve order. In particular, I had to change all the tests that use views from lists to sets. So that made the diff a little messy.

Finally, in the process of writing this I found a bug in how json hypergraphs were being read. The specified nodetype was not being respected. This is now fixed.


leotrs:
If anyone can review this it'd be great :) @nwlandry @maximelucas @iaciac

maximelucas:
Good job. The ability to filter edges and nodes like that is great.
I'm only worried about it not preserving order. Do you think it could made to preserve it?

The order is not fundamental in the sense that changing it does not make things wrong, but it's quite practical to keep a consistent order when, for example, you want to check visually if your functions does what you think it does. Especially for hypergraphs that are not super small. For example if you need check that correct edges were removed, and they are are in a completely different order.

xgi/classes/reportviews.py
```diff
@@ -149,18 +168,14 @@ def __getitem__(self, id):
...
 
         """
+        if id not in self._ids:
+            raise XGIError(f"The ID {id} is not in this view")
+
         try:
-            return self._id_attrs[id]
+            return self._id_attr[id]
         except KeyError as e:
             raise XGIError(f"The ID {id} is not in the hypergraph") from e
```
nwlandry, line 177: I would also add a TypeError for when a user specifies an unhashable id

xgi/classes/reportviews.py
```diff
@@ -194,8 +209,26 @@ def __repr__(self):
...
 
+    def __call__(self, size):
+        """Filter the results by size."""
+        bunch = [id for id in self._id_dict if len(self._id_dict[id]) == size]
+        return self.from_view(self, bunch)
+
+    @classmethod
+    def from_view(cls, view, bunch=None):
+        """Create a view from another view.
+
```
nwlandry, line 220: Could you add the parameters in the docstring?

leotrs:
@maximelucas you're right, I've reverted the changes and now it preserves order. I updated the tests as well. It will now be slower to check for membership but I think we can afford that for the time being.

xgi/classes/reportviews.py
leotrs: done!

xgi/classes/reportviews.py
leotrs: Done! I'll write more extensive docs in the docs branch

nwlandry (approved):
I'm approving with minor changes requested. If you could also remove `has_edge_id` from Hypergraph.py that would be great.

xgi/classes/simplicialcomplex.py
```diff
@@ -275,13 +272,10 @@ def remove_simplex_ids_from(self, ebunch):
...
-                for node in self.edges.members(id):
-                    self._node[node].remove(id)
-                del self._edge[id]
-                del self._edge_attr[id]
-            except:
-                pass
+            for node in self.edges.members(id):
+                self._node[node].remove(id)
+            del self._edge[id]
+            del self._edge_attr[id]
```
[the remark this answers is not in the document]
leotrs: In what for loop?

maximelucas (approved):

maximelucas:
Wait, why is `has_edge_id` unnecessary now?

leotrs:
> Wait, why is `has_edge_id` unnecessary now?

It was never necessary I think. Or perhaps it was necessary before we reworked the views. Now you can just do `e in H.edges`.

PR #70, opened 2022-03-23
A way to get edge members (not edge ids)

### LINKED ISSUE #28: Add function to get list of edges (not edge ids)
Currently, 

```python
H.edges 
# EdgeView([3, 2, 4, 9])
```
outputs edge ids, not edges as list of member nodes. 
It's often useful to access edges as list of member nodes.
One current way is ` list(H._edge.values()) ` but it's long and hidden.

Should we have a two views? 

- EdgeIdView (to replace current EdgeView): for ids
```python
H.edges_id
# EdgeIdView([3, 2, 4, 9])
```

- EdgeView:for list of member nodes
```python
H.edges
# EdgeView([[0], [3, 4], [4], [4]])
```

leotrs:
Fixes #28.

```python
H = xgi.Hypergraph([[1, 2, 3], [3, 4], [4, 5, 6]])
H.edges.members()
# -> [[1, 2, 3], [3, 4], [4, 5, 6]]
H.edges.members(dtype=dict)
# -> {0: [1, 2, 3], 1: [3, 4], 2: [4, 5, 6]}
```

### DIFF
--- a/tests/classes/test_reportviews.py
+++ b/tests/classes/test_reportviews.py
@@ -1,0 +1,10 @@
+import pytest
+
+import xgi
+from xgi.exception import XGIError
+
+
+def test_edge_members(edgelist3):
+    H = xgi.Hypergraph(edgelist3)
+    assert H.edges.members() == [[1, 2, 3], [3, 4], [4, 5, 6]]
+    assert H.edges.members(dtype=dict) == {0: [1, 2, 3], 1: [3, 4], 2: [4, 5, 6]}
--- a/xgi/classes/reportviews.py
+++ b/xgi/classes/reportviews.py
@@ -372,7 +372,7 @@ class EdgeView(IDView):
     def __init__(self, hypergraph):
         super(EdgeView, self).__init__(hypergraph._edge, hypergraph._edge_attr)
 
-    def members(self, e):
+    def members(self, e=None, dtype=list):
         """Get the nodes that are members of an edge.
 
         Given an edge ID, this method returns the node IDs
@@ -394,6 +394,14 @@ class EdgeView(IDView):
             Returns an error if the user tries passing in a slice or if
             the edge ID does not exist in the hypergraph.
         """
+        if e is None:
+            if dtype is dict:
+                return self._ids.copy()
+            elif dtype is list:
+                return [v for _, v in self._ids.items()]
+            else:
+                raise XGIError(f"Unrecognized dtype {dtype}")
+
         try:
             return self._ids[e]
         except KeyError:


xgi/classes/reportviews.py
```diff
@@ -372,7 +372,7 @@ class EdgeView(IDView):
     def __init__(self, hypergraph):
         super(EdgeView, self).__init__(hypergraph._edge, hypergraph._edge_attr)
 
-    def members(self, e):
+    def members(self, e=None, dtype=list):
         """Get the nodes that are members of an edge.
 
         Given an edge ID, this method returns the node IDs
```
maximelucas, line 378: add dtype to doc

xgi/classes/reportviews.py
```diff
@@ -372,7 +372,7 @@ class EdgeView(IDView):
     def __init__(self, hypergraph):
         super(EdgeView, self).__init__(hypergraph._edge, hypergraph._edge_attr)
 
-    def members(self, e):
+    def members(self, e=None, dtype=list):
```
maximelucas, line 375: add dtype to doc

leotrs:
> ```python
> H = xgi.Hypergraph([[1, 2, 3], [3, 4], [4, 5, 6]])
> 
> H.edge_ids
> # -> EdgeIDView((0, 1, 2))
> H.edges
> # -> EdgeView([[1, 2, 3], [3, 4], [4, 5, 6]])
> H.edges(return_dict=True)
> # -> {0: [1, 2, 3], 1: [3, 4], 2: [4, 5, 6]}
> ```

The first one is currently `H.edges`. The second one (after this PR) is currently `H.edges.members()`. I agree the that having something shorter would be nice.. but OTOH I do like the fact that the current `edges` and `edges.members` implementation emphasizes the fact that edges are usually treated as IDs, and not as collections of nodes. So it makes sense to me to use the shorter `H.edges` to expose the default version (the IDs) and the longer `H.edges.members()` to expose the non-default version (collections of nodes).

There is still no way of  getting the dict version, other than accessing directly `H._edges`, which is probably a bad idea. Maybe we can implement something like `H.edges.todict()` that returns a copy of the internal dictionary?

maximelucas:
Sure I agree, it's just a matter of how to call them. And I agree that your implementation makes sense :)
I'd say let's just merge for now and see at the next meeting, if people want to implement the other way later on or just keep this one.

xgi/classes/reportviews.py
```diff
@@ -394,6 +394,14 @@ def members(self, e):
             Returns an error if the user tries passing in a slice or if
             the edge ID does not exist in the hypergraph.
         """
+        if e is None:
+            if dtype is dict:
+                return self._ids.copy()
+            elif dtype is list:
+                return [v for _, v in self._ids.items()]
```
nwlandry, line 401: What about making this `list(self._ids.values())`?

nwlandry:
Thoughts on allowing users to specify an iterable of ids, e.g., `H.edges.members([0, 1])`?

xgi/classes/reportviews.py
leotrs: Done

tests/classes/test_reportviews.py
```diff
@@ -0,0 +1,10 @@
+import pytest
+
+import xgi
+from xgi.exception import XGIError
+
+
+def test_edge_members(edgelist3):
+    H = xgi.Hypergraph(edgelist3)
+    assert H.edges.members() == [[1, 2, 3], [3, 4], [4, 5, 6]]
+    assert H.edges.members(dtype=dict) == {0: [1, 2, 3], 1: [3, 4], 2: [4, 5, 6]}
```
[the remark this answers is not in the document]
leotrs: Yep, you're right.. but I think I'd rather keep this for now rather than deal with that ... (I think we need to normalize that too soon)

xgi/classes/reportviews.py
leotrs: Done

tests/classes/test_reportviews.py
```diff
@@ -0,0 +1,10 @@
+import pytest
+
+import xgi
+from xgi.exception import XGIError
+
+
+def test_edge_members(edgelist3):
+    H = xgi.Hypergraph(edgelist3)
+    assert H.edges.members() == [[1, 2, 3], [3, 4], [4, 5, 6]]
```
nwlandry, line 9: I would add an assert to check a single ID, i.e. `H.edges.members(0)` and check that an Error is thrown in the `None` case and the single id case.

nwlandry (approved):
I've approved this because my suggested changes were minor.

tests/classes/test_reportviews.py
leotrs: I added the new test case for members(0). The case of members(None) is actually the default, which outputs all of them so it shouldn't give an error-1

PR #73, opened 2022-04-08
fix: wrong dimension in sparse incidence matrix #72

### LINKED ISSUE #72: Wrong dimension in sparse incidence matrix
The incidence matrix should always have dimension `(num_nodes, num_edges)` or `(num_nodes, num_edges_of_order)`.

But when `sparse=True`, dimension 0 is wrong when some nodes are disconnected (overall, or at a given order when it is specified). This makes all the other functions using the incidence matrix wrong (adjacency, laplacian, etc.).

MWE:
```python
nodes = range(5)
edges = [[0,1], [0,2], [1,2,3]]

H = xgi.empty_hypergraph()
H.add_nodes_from(nodes)
H.add_edges_from(edges)

xgi.incidence_matrix(H, sparse=False).shape
## --> (5, 3) # good
xgi.incidence_matrix(H, sparse=True).shape
## --> (4, 3) # not good
```

This is because when `sparse=True`, the sparse matrix is built by iterating over the members of edges (of order d):

https://github.com/ComplexGroupInteractions/xgi/blob/388114d72d3ba03ea7c04ced52532762f3588ef8/xgi/linalg/matrix.py#L71-L82

maximelucas:
- fixed #72
- added tests to check that sparse and non-sparse incidence matrices are always the same


leotrs (approved):

nwlandry:
Everything looks good! I just formatted in black. I tried increasing the scipy ceiling but it still doesn't work :(

nwlandry (approved):
Nice work!

PR #75, opened 2022-04-11
fix: #74 wrong dimension for degree when no neighbours

### LINKED ISSUE #74: Wrong dim of linalg degree when no neighbours
Running 
```python
import xgi

edges = [[0,1], [0,2], [1,2,3]]
H = xgi.Hypergraph(edges)
```

yields 
```python
xgi.degree(H, order=3).shape
## --> (0,)
```
but it should be `(4,)`, the number of nodes. 

This then yields an error when running `xgi.laplacian(H, order=3)`, because it conflicts with the (correct) dimension of the adjacency matrix:

```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Input In [16], in <cell line: 1>()
----> 1 xgi.laplacian(H, order=3).shape

File ~/Dropbox (ISI Foundation)/WORK/SCIENCE/xgi/xgi/linalg/matrix.py:274, in laplacian(H, order, rescale_per_node, index)
    270     return (np.array([]), {}) if index else np.array([])
    272 K = degree(H, order=order, index=False)
--> 274 L = order * np.diag(np.ravel(K)) - A  # ravel needed to convert sparse matrix
    275 L = np.asarray(L)
    277 if rescale_per_node:

ValueError: operands could not be broadcast together with shapes (0,0) (4,4)
```

maximelucas:
- fixed #74 
- made the linalg `degree` accessible


leotrs:
Oh and let's change `degree` to `degree_matrix`?

leotrs (approved):

PR #77, opened 2022-04-11
Case study

maximelucas:
- added `multiorder_laplacian` function
- added first case study notebook


xgi/linalg/matrix.py
```diff
@@ -280,6 +281,64 @@ def laplacian(H, order=1, rescale_per_node=False, index=False):
...
+        Multiorder Laplacian for synchronization in higher-order networks.
+        Physical Review Research, 2(3), 033410.
+
+    """
+
+    Ls = [laplacian(H, order=i, rescale_per_node=rescale_per_node) for i in orders]
+    Ks = [degree_matrix(H, order=i) for i in orders]
+
+    L_multi = 0
+    for L, K, w in zip(Ls, Ks, weights):
```
leotrs, line 331: The documentation says that `weights` can be a single number or a list, but if it is a number then this `zip` will fail.

xgi/linalg/matrix.py
```diff
@@ -280,6 +281,64 @@ def laplacian(H, order=1, rescale_per_node=False, index=False):
...
+        L_multi += L * w / np.mean(K)
+
+    _, rowdict, _ = incidence_matrix(H, index=True)  # get index
+
+    if index:
+        return L_multi, rowdict
+    else:
+        return L_multi
```
leotrs, line 339: Perform the computation only if necessay:
```suggestion
    if index:
        _, rowdict, _ = incidence_matrix(H, index=True)
        return L_multi, rowdict
    else:
        return L_multi
```

leotrs (changes requested):
Notebook looks good. There's just one issue about the input types. If you have time, it'd be great to add a test though I know testing matrices is a PITA....

xgi/linalg/matrix.py
maximelucas: Done

leotrs (approved):

PR #81, opened 2022-04-12
Add IDDict which takes care of data validation and raising appropriate errors

leotrs:
Currently in several places of the code we have a version of the following code:
```python
if node not in self._node:
    raise XGIError("Node not in hypergraph")
```
Of course, this is not done every time it should be done, and the error message is always different. Sometimes the node is checked not only for membership but also it is checked that it does not equal None. But this practice is inconsistent. So I decided to abstract this out into a class `IDDict` that performs the data validation and then just acts as a usual dictionary. Added some tests to that effect.

I think the code node is cleaner and in the future we won't need to remember about data validation anymore. Or we can just modify the `IDDict` class instead of having to modify multiple try/except blocks.

### DIFF
--- a/tests/classes/test_hypergraph.py
+++ b/tests/classes/test_hypergraph.py
@@ -1,6 +1,6 @@
 import pytest
 import xgi
-from xgi.exception import XGIError
+from xgi.exception import XGIError, IDNotFound
 
 
 def test_constructor(edgelist5, dict5, incidence5, dataframe5):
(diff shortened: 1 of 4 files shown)


leotrs:
For now I think it's fine to leave it there since Hypergraph is the only class that is using it. Depending on how #82  turns out, it will be moved for sure.

maximelucas:
Approving now but might be worth having a second opinion to be sure.

maximelucas (approved):

nwlandry:
Looks great!

nwlandry (approved):

PR #84, opened 2022-04-13
Docs update

leotrs:
Cleaned up most of the Hypergraph methods and the View classes.


leotrs:
When reviewing please keep in mind writing documentation is always a work in progress :) I know there's still a ton to do, but I decided to PR now because it was getting too big.

xgi/classes/hypergraph.py
```diff
@@ -1154,20 +1154,6 @@ def max_edge_order(self):
             d_max = 0 if self._node else None
         return d_max
 
-    def edges_of_order(self, d):
```
maximelucas, line 1119: Nice but should replace the instances where it is used in the the two .py files in`drawing`.

leotrs: Good catch, done!

xgi/classes/hypergraph.py
```diff
@@ -769,7 +807,7 @@ def get_edge_data(self, id, default=None):
             return default
 
     def degree(self, nbunch=None, weight=None, order=None, dtype="dict"):
-        """A NodeDegreeView for the Hypergraph as H.degree or H.degree().
+        """A NodeDegreeView for the Hypergraph.
```
nwlandry, line 810: I think this is called `DegreeView` now

tests/classes/test_hypergraph.py
```diff
@@ -218,3 +219,48 @@ def test_egonet(edgelist3):
...
+        assert 0 in H.edges
+        assert [1, 2, 3] in H.edges.members()
+        assert [1, 2, 3] == H.edges.members(0)
+        assert H.edges.members(dtype=dict) == {0: [1, 2, 3]}
+
+    H = xgi.Hypergraph()
+    for edge in [[], set(), iter([])]:
+        with pytest.raises(XGIError):
+            H.add_edge(edge)
+
```
nwlandry, line 238: Maybe add a test for user-specified IDs and attributes? i.e.
```
H = xgi.Hypergraph()
H.add_edge([0, 1, 2], "test", color="blue")
assert "test" in H.edges
assert H.edges["test"] == "blue"
```

nwlandry: Sorry, disregard this. I see the tests below.

xgi/classes/reportviews.py
```diff
@@ -465,60 +388,57 @@ def __init__(self, hypergraph, bunch=None):
...
 
         Parameters
         ----------
         e : hashable
-            edge ID
-        dtype : list
-            spectify the type of the return value
+            Edge ID.
+        dtype : type, default list
+            Spectify the type of the return value.
```
nwlandry, line 402: "Specify"

xgi/classes/reportviews.py
```diff
@@ -465,60 +388,57 @@ def __init__(self, hypergraph, bunch=None):
...
-        except TypeError:
+        except KeyError as e:
+            raise XGIError(f"The edge ID {e} is not in the hypergraph") from e
+        except TypeError as e:
             if isinstance(e, slice):
                 raise XGIError(
                     f"{type(self).__name__} does not support slicing, "
                     f"try list(H.edges)[{e.start}:{e.stop}:{e.step}]"
-                )
+                ) from e
```
nwlandry, line 438: Wondering if on line 438 outside the if-statement we should have
```
raise XGIError("The edge ID {e} is not a hashable.") from e
```
and perhaps even remove the prior if statement?

nwlandry (approved):
Everything looks great. Just one or two minor changes, so as before, I'm approving to streamline the process.

xgi/classes/reportviews.py
maximelucas: Done

PR #86, opened 2022-04-15
Functions for loading XGI-DATA datasets

nwlandry:
* Added `open_file` decorator for loading file paths or a string of a loaded file
* Updated the json.py file to (1) have the `open_file` decorator and (2) utilize the add/remove methods instead of modifying the Hypergraph dictionaries directly.
* Added functionality to load datasets directly from the xgi-data repository


nwlandry:
I also
* modified the Github Actions script to not run unnecessary tests.
* added seaborn to the tutorials requirements and requests to the default requirements
* added the `open_file` decorator to all read/write functions

tests/utils/test_utilities.py
```diff
@@ -0,0 +1,29 @@
+from xgi.utils import XGICounter, get_dual, load_xgi_data
+
+
+def test_load_xgi_data():
```
leotrs, line 4: I'm hesitant to make an HTTP request part of the test suite. Perhaps we could mark it as [slow](https://docs.pytest.org/en/6.2.x/mark.html).

tests/utils/test_utilities.py
```diff
@@ -0,0 +1,29 @@
...
+    assert dual[2] == [0]
+    assert dual[3] == [0]
+    assert dual[4] == [1]
+    assert dual[5] == [2]
+    assert dual[6] == [2, 3]
+    assert dual[7] == [3]
+    assert dual[8] == [3]
+
+
+def test_xgi_counter():
```
leotrs, line 26: Can we replace XGICounter with python's [count](https://docs.python.org/3/library/itertools.html#itertools.count)? If so, we can leave that for a separate PR but we should as much as possible make use of what's already in the standard library.

leotrs (commented):
Unfortunately the docstring of the open_file decorator didn't explain to me very well what the decorator is doing or why is it needed. Could you add a little context? What problem is this solving and why is this the best way of solving it?

tests/utils/test_utilities.py
nwlandry: Done.

tests/utils/test_utilities.py
nwlandry: I agree! I'm not familiar with `count` until now, but I will make an issue to deal with later.

nwlandry: Actually, I'm not sure that it does the same thing. `XGICounter` increments a counter every time that it is called, which doesn't seem to be the same as `count` from what I could tell.

nwlandry:
This is now being replaced by #88 so I'm closing this.

Issue #23, opened 2021-12-10
Modify degree_counts function


nwlandry:
Change API such that
1. degree_counts returns what degree_histogram currently returns, and
2. degree_histogram returns the same data structures as np.unique(i.e. one array with the heights and one array with the degree bins.)

PR #87, opened 2022-04-18
feat(api)!: Changed api of degree_histogram and added degree_counts based on #23

tlarock:
Addresses #23.

I refactored the existing `degree_histogram` code into a function called `degree_counts`. I did not modify the behavior of `degree_counts` from the original `degree_histogram`.

I then wrote a new `degree_histogram` function that returns a `tuple` of 2 `list`s, where the first list corresponds to the observed degrees (bin centers for the histogram) in ascending order, and the second tuple is the number of nodes with the degree in the same index of the first list (heights of the histogram). Note that, unlike the behavior of `degree_counts`, I do not return entries with count 0, as there is no need to infer the degree from the list since I return the list of observed degrees explicitly. These lists can be passed directly to `matplotlib.pyplot.[bar | plot]` for visualization.

BREAKING CHANGE: Existing code using the old version of `degree_histogram` will now fail because the new `degree_histogram` returns a 2 tuple, not a list of degree values. Code using `degree_histogram` should be modified to use `degree_counts`.

Edit: The original issue (#23) asks to mimic the `np.unique` return - this pull request does achieve that, though it returns a tuple of lists, not an `np.array`. `numpy` is not being used in this file so I did not add the dependency and just used tuple of lists instead.

### DIFF
--- a/tests/classes/test_function.py
+++ b/tests/classes/test_function.py
@@ -4,14 +4,14 @@ import xgi
 from xgi.exception import IDNotFound, XGIError
 
 
-def test_degree_histogram(edgelist1, edgelist2, edgelist3):
+def test_degree_counts(edgelist1, edgelist2, edgelist3):
     H1 = xgi.Hypergraph(edgelist1)
     H2 = xgi.Hypergraph(edgelist2)
     H3 = xgi.Hypergraph(edgelist3)
 
-    assert xgi.degree_histogram(H1) == [0, 7, 1]
-    assert xgi.degree_histogram(H2) == [0, 5, 1]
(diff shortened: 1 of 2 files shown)


leotrs (approved):
Thank you Timothy for your contribution!

nwlandry:
Thanks for contribution! Looks great! Only question I have is if the functions should accept the `order` and `weight` parameters that `.degree()` now accepts.

leotrs:
TBH I'm happy merging this as it is and then worrying about extra parameters when the new super secret `nodestat` functionality is in place ;)

nwlandry:
Sounds good!

PR #83, opened 2022-04-12
Myst notebooks

nwlandry:
* Added jupytext as a dependency in the tutorials requirements file.
* Converted the tutorials to MyST (markdown) files
* Removed the Jupyter notebooks
* Provided a brief description for how to convert between markdown and Jupyter notebooks.

PR #88, opened 2022-04-18
Function to get datasets from xgi-data

nwlandry:
* Updated list of requirements (added seaborn in `tutorials.txt` and requests in `default.txt`)
* Fixed `read_hypergraph_json` to handle isolates.
* Added a function `load_xgi_data()` to get datasets from the [xgi-data respository](https://github.com/ComplexGroupInteractions/xgi-data).
* Added `webtest` and `slow` pytests markers in pytest.ini and added pytest.ini to MANIFEST.in


nwlandry:
Modified the Github actions trigger as suggested in #83.

leotrs (approved):
Still think we should replace XGICounter with itertools.count, but we can leave that for a different PR. This looks much better than the previous one!

PR #90, opened 2022-04-20
Docs

maximelucas:
- added new SimplicialComplex constructors and generators (by Gio)
- added tests for the above new functions
- cleaned SimplicialComplex docs


maximelucas:
Eh it looks messy because it shows all the commits I merged from main...

xgi/generators/nonuniform.py
```diff
@@ -418,15 +424,24 @@ def random_flag_complex(N, p, seed=None, max_order=2):
...
     -----
     Computing all cliques quickly becomes heavy for large networks.
 
+    Example
+    -------
+    >>> import xgi
+    >>> N = 100
+    >>> p = 0.05
+    >>> S = xgi.random_flag_complex(N, p, max_order=2)
+
```
leotrs, line 444: I'd get rid of this example for two reasons: 1) it doesn't actually illustrate any functionality, it just shows a function call, and 2) it slows down the doctest run.

leotrs (changes requested):

xgi/generators/nonuniform.py
maximelucas: Done.

PR #96, opened 2022-04-21
Organize Hypergraph documentation

leotrs:
Split the methods in separate tables, so it is more clear what are the broad capabilities of the class.

Also reordered the autosummary table of the reportviews module.


leotrs:
After this we should delete this branch and create a new docs branch so the long commit list doesn't keep appearing...

maximelucas:
I think it's great! Could add the same method categories for the SImplicialComplex class.

maximelucas (approved):

PR #106, opened 2022-04-29
Refactor subhypergraph methods

leotrs:
* Remove unnecessary duplicated functions generic_subhypergraph_view and subhypergraph_view. The latter can already do what the former did by just passing None as arguments.
* Remove the subhypergraph methods from Hypergraph that were just thin wrappers around subhypergraph_view. Not only were these thin wrappers but the creation of subhypergraph views is something that (IMHO) is better located outside the main class and into the main library. Accordingly, we can now create subhypergraphs by calling `xgi.subhypergraph(...)`.
* Rename subhypergraph_view as subhypergraph. 
* Remove the nbunch_iter method that is no longer needed. It was only being used in one of the wrapper functions. I think a method like this could be helpful in the future, but for now I don't see the need for it. We can revisit later on.


tests/classes/test_hypergraphviews.py
```diff
@@ -7,7 +7,7 @@
 def test_generic_hypergraph_view(edgelist1):
```
nwlandry, line 7: Rename this function to match the new function name.

nwlandry (approved):
This all looks good. My only thought is that in the subhypergraph function, if we should transition to using `add_nodes_from` and `add_edges_from` instead of directly accessing the dictionaries. This could just become an open issue, though, if you feel that that's best. Because I only requested a small change, I'm approving it.

nwlandry:
Also, I think the docs need to be modified to reflect the change you made.

leotrs:
> This all looks good. My only thought is that in the subhypergraph function, if we should transition to using `add_nodes_from` and `add_edges_from` instead of directly accessing the dictionaries. This could just become an open issue, though, if you feel that that's best. Because I only requested a small change, I'm approving it.

I'm actually not sure about that. We should profile both options and determine which one is faster.

PR #107, opened 2022-05-05
Remove get_edge_data and has_node

### LINKED ISSUE #78: There are three different ways of accessing edge data
`H.edge[id]`, `xgi.get_edge_attributes`, and `H.get_edge_data`. All of these should be handled via the same interface, perhaps a `AttrView` object that we used to have.

leotrs:
The documentation of `Hypergraph.get_edge_data(x, val)` claims that it is the same as `H.edges[x]` with the sole exception that if `x` is not an edge, it fails more nicely, i.e. it returns the fallback value `val` instead of raising an exception. However, this functionality is already possible in pure python by calling `H.edges.get(x, val)`. This is not only more pythonic but also shorter.

In order to implement this change, I had to delete the `get_edge_data` method as well as change the type of `IDNotFound` exception from `Exception` to `KeyError`. That made it possible for all methods implemented automatically by `IDDict(Mapping)` to work perfectly. (More concretely, for `Mapping` to implement `get` correctly, it was expecting `__getitem__` to raise a `KeyError` or one of its descendants, rather than an `XGIError` as was previously).

The rest of the diff is due to some cleanup in how certain exceptions are handled, and the updating of relevant tests. 

EDIT: I have updated this PR to also remove `Hypergraph.has_node(x)`. This is identical to `x in H` and `x in H.nodes`. There is no reason to have THREE different ways to do the exact same thing... I have left `Hypergraph.has_edge` intact for now because it is unclear whether it is exactly identical to anything else currently.

Closes #78


nwlandry (approved):
Looks good! It's especially slick how the IDDict is seamlessly handling a lot of the logic now, which is great.

PR #110, opened 2022-05-07
Refactor members

nwlandry:
Improved the speed of the `members()` method in the `EdgeView` class by refactoring. Before refactoring, in the following list comprehension
`[H.edges.members(e) for e in H.edges]`
the `members()` method took 1.254s for 10885 function calls on the [email-enron dataset](https://github.com/ComplexGroupInteractions/xgi-data/blob/main/data/email-Enron/README_email-Enron.md). After refactoring, the same 10885 function calls took 0.014s. This refactor did not change the speed of `H.edges.members()`.

### DIFF
--- a/tests/classes/test_reportviews.py
+++ b/tests/classes/test_reportviews.py
@@ -2,7 +2,7 @@ import numpy as np
 import pytest
 
 import xgi
-from xgi.exception import XGIError
+from xgi.exception import IDNotFound, XGIError
 
 
 def test_edge_order(edgelist3):
@@ -116,6 +116,17 @@ def test_edge_members(edgelist3):
     assert H.edges.members(0) == [1, 2, 3]
     assert H.edges.members() == [[1, 2, 3], [3, 4], [4, 5, 6]]
     assert H.edges.members(dtype=dict) == {0: [1, 2, 3], 1: [3, 4], 2: [4, 5, 6]}
+    with pytest.raises(XGIError) as exception_info:
+        H.edges.members(dtype=np.array)
+    assert "Unrecognized dtype" in str(exception_info.value)
+
+    with pytest.raises(XGIError) as exception_info:
+        H.edges.members(slice(1, 4, 1))
+    assert "try list(H.edges)[1:4:1]" in str(exception_info.value)
+
+    with pytest.raises(IDNotFound) as exception_info:
+        H.edges.members("test")
+    assert "Item test not in this view" == str(exception_info.value)
 
 
 def test_bunch_view(edgelist1):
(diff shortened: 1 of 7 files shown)


tests/classes/test_reportviews.py
```diff
@@ -116,6 +116,17 @@ def test_edge_members(edgelist3):
...
+        H.edges.members(dtype=np.array)
+    assert "Unrecognized dtype" in str(exception_info.value)
+
+    with pytest.raises(XGIError) as exception_info:
+        H.edges.members(slice(1, 4, 1))
+    assert "try list(H.edges)[1:4:1]" in str(exception_info.value)
+
+    with pytest.raises(IDNotFound) as exception_info:
+        H.edges.members("test")
+    assert "Item test not in this view" == str(exception_info.value)
```
leotrs, line 129: I'm not sure what I think about these ones. What is the benefit of testing that "Item test not in this view" is exactly the message attached to the exception versus just testing that the type of the exception is correct?

xgi/classes/reportviews.py
```diff
@@ -413,25 +413,23 @@ def members(self, e=None, dtype=list):
...
-            raise IDNotFound(f"Item {e} not in this view")
-
         try:
             return self._id_dict[e].copy()
-        except TypeError as e:
+        except TypeError:
             if isinstance(e, slice):
                 raise XGIError(
                     f"{type(self).__name__} does not support slicing, "
                     f"try list(H.edges)[{e.start}:{e.stop}:{e.step}]"
```
leotrs, line 422: Do we realistically expect users to call `members` with a slice as an argument? I wouldn't expect this to be a frequent occurrence, so I vote we delete this check.

xgi/classes/reportviews.py
```diff
@@ -413,25 +413,23 @@ def members(self, e=None, dtype=list):
...
+                )
+        except IDNotFound:
+            if e is None:
+                if dtype is dict:
+                    return {key: self._id_dict[key] for key in self._ids}
+                elif dtype is list:
+                    return [self._id_dict[key] for key in self._ids]
+                else:
+                    raise XGIError(f"Unrecognized dtype {dtype}")
+            raise IDNotFound(f"Item {e} not in this view")
```
leotrs, line 432: Perfect!

xgi/classes/reportviews.py
nwlandry: Good call. Now everything (list, dict, slice, etc.) is caught as a TypeError, which I do think is important to handle.

tests/classes/test_reportviews.py
leotrs: Usually one relies on the type of the exception to test what kind of error was raised. In he first two cases, the actual exception type raise should be ValueError and not really XGIError, and testing for that would be better than testing for the actual message I think.

leotrs (approved):

PR #112, opened 2022-05-09
refactor: improve speed of accessing nodes and edges

nwlandry:
* accessing `ids` and `id_dict` now by memory instead of making an unnecessary new copy.
* Removed lazy view creation because it decreased the performance of accessing nodes and edges.

Running, for example `[H.edges.members(e) for e in H.edges(e)]` for the [email-enron](https://github.com/ComplexGroupInteractions/xgi-data/blob/main/data/email-Enron/README_email-Enron.md) dataset, the profile before is
IDView_old
and after is
IDView_new

### DIFF
--- a/benchmarks/benchmark.ipynb
+++ b/benchmarks/benchmark.ipynb
@@ -2,7 +2,7 @@
  "cells": [
   {
    "cell_type": "code",
-   "execution_count": 11,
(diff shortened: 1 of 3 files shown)


leotrs (approved):
I'm surprised this change makes such a difference!

PR #113, opened 2022-05-13
Fix views

leotrs:
Refactor and improve `Hypergraph.add_edges_from`. It now supports different input formats. See the new docstring for details. The resulting function is long and a bit hard to follow but the main benefit is that it never assumes the input data is a list, only an iterable, so this function will work with huge amounts of data passed in. Additionally, added a bunch of new tests and modified some functions across the codebase to use `add_edges_from` and `add_nodes_from` instead of manipulating `H._nodes` and `H._edges` directly.

I also added a property `ids` to `IDView` to fix a bug where it would not update after the underlying hypergraph structure changed.

NOTE: Please ignore the name of this branch - I had started it to do something different.


tests/classes/test_hypergraph.py
```diff
@@ -9,32 +9,36 @@ def test_constructor(edgelist5, dict5, incidence5, dataframe5):
     H_dict = xgi.Hypergraph(dict5)
     H_mat = xgi.Hypergraph(incidence5)
     H_df = xgi.Hypergraph(dataframe5)
+    H_hg = xgi.Hypergraph(H_list)
```
maximelucas, line 12: Wait, what's the reason to allow create a Hypegraph from a Hypergraph again? 
Something about the subhypergraphs needing to be Hypergraphs?

tests/classes/test_hypergraph.py
```diff
@@ -267,3 +271,143 @@ def test_add_node_to_edge():
...
+    H.add_edges_from(edges)
+    assert list(H.edges.members()) == [list(e) for e in edges]
+
+    edges = [["foo", "bar"], ["bar", "baz"], ["foo", "bar", "baz"]]
+    H = xgi.Hypergraph()
+    H.add_edges_from(edges)
+    assert list(H.nodes) == ["foo", "bar", "baz"]
+    assert list(H.edges.members()) == [list(e) for e in edges]
+
+    edges = [["a", "b"], ["b", "c"], ["c", "d", "e"]]
```
maximelucas, line 303: is this not covered by the foor bar baz case above?

xgi/classes/hypergraph.py
```diff
@@ -512,10 +513,20 @@ def add_edges_from(self, ebunch_to_add, **attr):
 
         Parameters
         ----------
-        ebunch_to_add : iterable of edges
-            Each edge given in the iterable will be added to the
-            graph. Each edge must be given as as an iterable of nodes
-            or an iterable with the last entry as a dictionary.
+        ebunch_to_add : iterable
+
+            An iterable of edges, in one of the following formats.
```
maximelucas, line 498: These formats refer to each edge in the iterable of edges that is the input, not the whole input, right? I would make that a tad more explicit. Maybe
"An iterable of edges, where each edge must be in one of the following formats (basically, an iterable of nodes with extra options):

xgi/classes/hypergraph.py
```diff
@@ -512,10 +513,20 @@ def add_edges_from(self, ebunch_to_add, **attr):
...
+
+            * Format 1: Iterable of hashables (id1, id2, ...), or
```
maximelucas: It's more important to know it's node ids (which I didn't find clear at first), rather than that it must be hashable. Also this (id1, id2, ...) is the same as `members` in the other formats I think. Maybe we can make those things more explicit?

Maybe here just 
"* Format 1: Iterable of hashables nodes IDs (id1, id2, ...), or

xgi/classes/hypergraph.py
```diff
@@ -512,10 +513,20 @@ def add_edges_from(self, ebunch_to_add, **attr):
...
+            * Format 2: 2-tuple (members, id), or
+            * Format 3: 2-tuple (members, attr), or
+            * Format 4: 3-tuple (members, id, attr),
+
+            where `members` is an iterable of node IDs, each `id` is a hashable to use
```
maximelucas, line 506: and here 
"where `members` is an iterable of node IDs (just like format 1), each `id` is a hashable to use..."

I would maybe even use `edge_id` in formats 2 and 4 to make it clear

xgi/classes/hypergraph.py
```diff
@@ -512,10 +513,20 @@ def add_edges_from(self, ebunch_to_add, **attr):
...
+            as edge ID, and `attr` is a dict of attributes. The second and third formats
+            are unambiguous because `attr` dicts are not hashable, while `id`s must be.
+            In all except the first format, each element of `ebunch_to_add` must have
+            the same length.
```
maximelucas: add "(i.e 2 or 3)"

xgi/classes/hypergraph.py
```diff
@@ -531,27 +542,53 @@ def add_edges_from(self, ebunch_to_add, **attr):
...
+        except StopIteration:
+            return
+        try:
+            first_elem = list(first_edge)[0]
+        except TypeError:
+            first_elem = None
+
+        format1, format2, format3, format4 = False, False, False, False
+        if isinstance(first_elem, Iterable):
+            if len(first_edge) == 2 and hasattr(first_edge[1], "__hash__"):
```
maximelucas: maybe easier to understand/read if checking if `dict` rather than hash?

xgi/classes/hypergraph.py
```diff
@@ -513,20 +513,21 @@ def add_edges_from(self, ebunch_to_add, **attr):
 
         Parameters
         ----------
-        ebunch_to_add : iterable
+        ebunch_to_add : Iterable
 
             An iterable of edges, in one of the following formats.
 
-            * Format 1: Iterable of hashables (id1, id2, ...), or
+            * Format 1: Iterable (edge1_members, edge2_members), or
```
maximelucas, line 500: Wait, this should a single edge, so this is now wrong?
Should be (node_id1, node_id1, ...) or (edge_members) I guess.

xgi/classes/hypergraph.py
```diff
@@ -513,20 +513,21 @@ def add_edges_from(self, ebunch_to_add, **attr):
...
             * Format 2: 2-tuple (members, id), or
             * Format 3: 2-tuple (members, attr), or
             * Format 4: 3-tuple (members, id, attr),
+            * Format 5: a dict of the form {id: members}
```
maximelucas, line 504: are format 2 and 5 not redundant?

xgi/classes/reportviews.py
```diff
@@ -84,8 +84,15 @@ def __init__(self, id_dict, id_attr, ids=None):
 
     @property
     def ids(self):
-        """The ids in this view."""
-        return list(self._id_dict.keys()) if self._ids is None else self._ids
+        """The ids in this view.
+
+        Notes
+        -----
+        DO NOT use this property for membership check: instead of `x in view.ids` always
```
maximelucas, line 91: why?

xgi/classes/hypergraph.py
```diff
@@ -98,6 +98,9 @@ def __init__(self, incoming_data=None, **attr):
         self._edge = self._hyperedge_dict_factory()
         self._edge_attr = self._hyperedge_attr_dict_factory()
 
+        self.nodes = NodeView(self)
```
maximelucas, line 101: wait, what does this change?
It seems we loose the explanation from the dosctrings of the nodes/edges property we had? See lines deleted below on the left.

maximelucas:
Thanks for this Leo. I just left a bunch of comments mainly just to try and clarify some of the docstrings.

tests/classes/test_hypergraph.py
leotrs: I had to add this one because some of the formats accepted by `add_edges_from` differentiate between some of the elements having three elements (covered by "foo" "bar" "baz") and one element (covered by this test).

tests/classes/test_hypergraph.py
leotrs: Uuuuuh, I swear I had a good reason when I implemented it, but can't think of it anymore?

xgi/classes/hypergraph.py
leotrs: Good catch, I updated the docs.

xgi/classes/hypergraph.py
leotrs: Good catch, I updated the docs.

xgi/classes/hypergraph.py
leotrs: Good catch, I updated the docs.

xgi/classes/hypergraph.py
leotrs: This is checking for the first edge, no the entire thing :)  BTW I updated to check against `Hashable` rather than `__hash__` (bc there are objects that implement __hash__ but are not hashable...)

xgi/classes/hypergraph.py
leotrs: Yes they are! But I thought adding the dict would be nice since it is the format returned by `H.edges.members(dtype=dict)`. So you can easily add edges from another hypergrpah.

xgi/classes/reportviews.py
leotrs: Because this is slower. I updated the docs to reflect this fact.

xgi/classes/hypergraph.py
leotrs: Done!

xgi/classes/hypergraph.py
leotrs: It was wrong before, I fixed it now.

xgi/classes/hypergraph.py
```diff
@@ -512,10 +513,20 @@ def add_edges_from(self, ebunch_to_add, **attr):
...
+            * Format 2: 2-tuple (members, id), or
+            * Format 3: 2-tuple (members, attr), or
+            * Format 4: 3-tuple (members, id, attr),
+
+            where `members` is an iterable of node IDs, each `id` is a hashable to use
+            as edge ID, and `attr` is a dict of attributes. The second and third formats
+            are unambiguous because `attr` dicts are not hashable, while `id`s must be.
+            In all except the first format, each element of `ebunch_to_add` must have
+            the same length.
         attr : keyword arguments, optional
```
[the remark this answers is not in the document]
leotrs: Yes, one is per-edge and the other one applies to all edges. I have updated the docs and added examples and tests.

xgi/classes/hypergraph.py
leotrs: After talking to @nwlandry extensively about this, we realized that 1) the views were not working, 2) we don't need the lazy initialization. This PR fixes both problems. Since the views don't need to be lazy, we can just create them once and for all in the constructor.

I have added a one-line documentation for now. My next PR is going to continue changing some things around so I'd rather wait until then to add more documentation so I don't have to write everything twice :)

PR #118, opened 2022-06-03
Move `egonet()` and improve `neighbors()`

nwlandry:
* Moved the `egonet()` function from the Hypergraph class to the function module.
* Added the ability to get edge neighbors as well as node neighbors with the `neighbors()` function.

### DIFF
--- a/tests/classes/test_hypergraph.py
+++ b/tests/classes/test_hypergraph.py
@@ -97,21 +97,6 @@ def test_dual(edgelist1, edgelist2, edgelist4):
     assert (D3.num_nodes, D3.num_edges) == (3, 5)
 
 
-def test_max_edge_order(edgelist1, edgelist4, edgelist5):
-    H0 = xgi.empty_hypergraph()
-    H1 = xgi.empty_hypergraph()
-    H1.add_nodes_from(range(5))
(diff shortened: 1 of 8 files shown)


tests/classes/test_hypergraph.py
```diff
@@ -80,6 +80,12 @@ def test_neighbors(edgelist1, edgelist2):
     assert H2.neighbors(4) == {3, 5, 6}
     assert H2.neighbors(1) == {2}
 
+    # test edge neighbors
+    assert H1.neighbors(0, edges=True) == set()
+    assert H1.neighbors(1, edges=True) == set()
+    assert H1.neighbors(2, edges=True) == {3}
+    assert H1.neighbors(3, edges=True) == {2}
```
leotrs, line 87: What do you think about `H.nodes.neighbors` and `H.edges.neighbors` instead?

nwlandry: I like this idea! I guess that this should get implemented on the `feat_nodestats` branch?

leotrs: Yep!

leotrs (approved):

PR #120, opened 2022-06-07
stats package

leotrs:
This PR adds the `stats` package which implements `NodeStat`, `EdgeStat` and related functionality. For an overview of all the new features, see the [tutorial](https://github.com/ComplexGroupInteractions/xgi/blob/feat_nodestats/tutorials/Tutorial%206%20-%20Statistics.ipynb).

The main changes are all within the newly created `xgi/stats/` package. Other classes affected are `Hypergraph`, `NodeView`, and `EdgeView`. Additionally, `DegreeView` and `EdgeSizeView` are deleted.

This is a big diff, but it can basically be broken in three: changes to the docs (docstrings and *rst files), new tests (this PR adds 51 new tests collected by pytest), and the actual implementation, which is mostly within `xgi/stats/`, `xgi/classes/hypergraph.py` and `xgi/classes/reportviews.py`.

In terms of documentation, I believe we now have the minimum required to make sense of the new features, but a lot more can be done for sure. To avoid repetition, in some places I just added a direct link to the tutorial.

On a technical note. The way the `xgi/stats/` package is implemented is the following. `xgi/stats/__init__.py` contains all the major classes and decorators. `xgi/stats/nodestats.py` contains only the functions that actually compute node statistics (like degree clustering, etc). Similarly, `xgi/stats/edgestats.py` contains only the functions that compute edge statistics. I don't like having so much code in the `__init__.py` module, but it is necessary for the current implementation that `nodestats.py` and `edgestats.py` ONLY contain functions that compute the corresponding statistics.


leotrs:
Please do NOT review/merge unless #121 has been merged.

docs/source/api/stats/.#xgi.stats.MultiEdgeStat.rst
```diff
@@ -0,0 +1 @@
+leo@syenite.46863:1654177428
```
nwlandry, line 1: delete this

docs/source/api/stats/xgi.stats.EdgeStat.rst
```diff
@@ -0,0 +1,35 @@
+﻿xgi.stats.EdgeStat
```
nwlandry, line 1: There seem to be a few duplicate file names here

tests/generators/test_uniform.py
```diff
@@ -15,4 +15,4 @@ def test_uniform_configuration_model_hypergraph():
         m = 3
         k = {1: 1, 2: 6}
         H = xgi.uniform_hypergraph_configuration_model(k, m)
-    assert sum([deg for _, deg in H.degree()]) % m == 0
+    assert sum([deg for _, deg in H.degree().items()]) % m == 0
```
nwlandry, line 18: Change this to `assert sum(H.nodes.degree.aslist()) % m == 0`

xgi/generators/nonuniform.py
```diff
@@ -443,3 +445,19 @@ def random_flag_complex(N, p, max_order=2, seed=None):
     S.add_simplices_from(max_cliques, max_order=max_order)
 
     return S
+
+
+@np_random_state("seed")
+def watts_strogatz_hypergraph(n, d, k, l, p, seed=None):
+    H = lattice(n, d, k, l)
```
nwlandry, line 452: Is there a reference for this and the lattice hypergraph? Both look great!

xgi/stats/nodestats.py
```diff
@@ -0,0 +1,230 @@
...
+>>> H = xgi.Hypergraph([[1, 2, 3], [2, 3, 4, 5], [3, 4, 5]])
+>>> H.degree()
+{1: 1, 2: 2, 3: 3, 4: 2, 5: 2}
+>>> H.nodes.degree.asdict()
+{1: 1, 2: 2, 3: 3, 4: 2, 5: 2}
+
+"""
+
+import xgi
+from itertools import combinations
```
nwlandry, line 22: I think this is an unused import?

xgi/generators/nonuniform.py
leotrs: Nope, I was just trying to quickly add something to try out something on WS-like hypergraphs... I kinda want to keep them for now, though undocumented because I haven't really tested these at all.

docs/source/api/stats/xgi.stats.EdgeStat.rst
leotrs: All of `nodestats`, `NodeStat`, and `nodestat` are valid names implemented in this PR: the module, the class, and the decorator. All three names are imported into the `xgi.stats` package, so sphinx creates all three files.

I think it's fine, at least documentation-wise, though these three names are too similar to each other that they may eventually cause a problem with end-users?

leotrs: Open to suggestions here

PR #122, opened 2022-06-14
Add assortativity module

nwlandry:
Adding the computation of the dynamical assortativity and the assortativity values defined in Phil Chodrow's "Configuration Models of Random Hypergraphs"

### DIFF
--- a/xgi/algorithms/__init__.py
+++ b/xgi/algorithms/__init__.py
@@ -1,4 +1,3 @@
 from . import assortativity, connected
-
 from .assortativity import *
 from .connected import *
--- a/xgi/algorithms/__init__.py
+++ b/xgi/algorithms/__init__.py
@@ -1,4 +1,4 @@
-from xgi.algorithms import assortativity, connected
+from . import assortativity, connected
 
 from .assortativity import *
 from .connected import *
--- a/xgi/algorithms/__init__.py
+++ b/xgi/algorithms/__init__.py
@@ -1,2 +1,4 @@
-from . import connected
+from xgi.algorithms import assortativity, connected
+
+from .assortativity import *
 from .connected import *
--- a/xgi/algorithms/assortativity.py
+++ b/xgi/algorithms/assortativity.py
@@ -141,6 +141,7 @@ def choose_degrees(e, k, kind="uniform"):
             return degs
 
         elif kind == "top-bottom":
+            # this selects the largest and smallest degrees in one line
             degs = sorted([k[i] for i in e])[:: len(e) - 1]
             random.shuffle(degs)
             return degs
--- a/xgi/algorithms/assortativity.py
+++ b/xgi/algorithms/assortativity.py
@@ -1,0 +1,66 @@
+import numpy as np
+import xgi
+from xgi.exception import XGIError
+from itertools import combinations
+import random
+
+__all__ = [
+    "dynamical_assortativity",
+    "top_2_assortativity",
+    "top_bottom_assortativity",
+    "uniform_assortativity",
+]
+
+def dynamical_assortativity(H):
+    """ Gets the dynamical assortativity of a uniform hypergraph.
+
+    Parameters
+    ----------
+    H : xgi.Hypergraph
+        Hypergraph of interest
+
+    Returns
+    -------
+    _type_
+        _description_
+
+    Raises
+    ------
+    XGIError
+        _description_
+    """
+    if not xgi.is_uniform(H):
+        raise XGIError("Hypergraph must be uniform!")
+    
+    k1 = np.mean([H.degree(n) for n in H.nodes])
+    k2 = np.mean([H.degree(n)**2 for n in H.nodes])
+
+    kk1 = np.mean([H.degree(n1)*H.degree(n2) for e in H.edges for n1, n2 in combinations(H.edges.members(e), 2)])
+    return kk1*k1**2/k2**2 - 1
+
+
+def top_2_assortativity(H):
+    degrees = dict(H.degree())
+    k1 = [max((degrees[n] for n in H.edges.members(e))) for e in H.edges]
+    try:
+        k2 = [sorted((degrees[n] for n in H.edges.members(e)))[-2] for e in H.edges]
+    except IndexError:
+        raise XGIError("Hypergraph must not have singleton edges.")
+    return np.corrcoef(k1, k2)[0, 1]
+
+
+def top_bottom_assortativity(H):
+    degrees = dict(H.degree())
+    k1 = [max((degrees[n] for n in H.edges.members(e))) for e in H.edges]
+    k2 = [min((degrees[n] for n in H.edges.members(e))) for e in H.edges]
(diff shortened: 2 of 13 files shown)


xgi/algorithms/__init__.py
```diff
@@ -1,2 +1,4 @@
-from . import connected
+from xgi.algorithms import assortativity, connected
```
leotrs, line 1:
```suggestion
from . import assortativity, connected
```

xgi/algorithms/assortativity.py
```diff
@@ -0,0 +1,156 @@
...
+    """
+    if not xgi.is_uniform(H):
+        raise XGIError("Hypergraph must be uniform!")
+
+    if H.num_nodes == 0 or H.num_edges == 0:
+        raise XGIError("Hypergraph must contain nodes and edges!")
+
+    degs = H.degree()
+    k1 = np.mean([degs[n] for n in H.nodes])
+    k2 = np.mean([degs[n] ** 2 for n in H.nodes])
```
leotrs, line 47: Nitpick: this makes two passes over the set of nodes, but one is possible, something along these lines
```suggestion
    k1, k2 = np.array([(degs[n], degs[n]**2) for n in H.nodes]).mean(axis=1)
```

Not sure about the `axis=1`, but you get the idea.

xgi/algorithms/assortativity.py
```diff
@@ -0,0 +1,156 @@
...
+    Journal of Complex Networks 2020.
+    DOI: 10.1093/comnet/cnaa018
+    """
+    degs = H.degree()
+    if exact:
+        k1k2 = [
+            choose_degrees(e, degs, type)
+            for e in H.edges
+            if len(H.edges.members(e)) > 1
+        ]
```
leotrs, line 93: Just out of curiosity (and if you have time), could you please check whether the following is faster?
```python
[choose_degrees(e, degs, type) for e in H.edges.filterby("order", 1, "geq")]
```

xgi/algorithms/assortativity.py
```diff
@@ -0,0 +1,156 @@
...
+    else:
+        edges = list([e for e in H.edges if len(H.edges.members(e)) > 1])
```
leotrs, line 95:
```suggestion
        edges = [e for e in H.edges if len(H.edges.members(e)) > 1]
```

xgi/algorithms/assortativity.py
```diff
@@ -0,0 +1,156 @@
...
+        k1k2 = list()
+
+        samples = 0
+        while samples < num_samples:
+            e = random.choice(edges)
+            k1k2.append(choose_degrees(H.edges.members(e), degs, type))
+            samples += 1
```
leotrs, line 102:
```suggestion
        k1k2 = [choose_degrees(H.edges.members(random.choice(edges)), degs, type) for _ in range(num_samples)]
```

xgi/algorithms/assortativity.py
```diff
@@ -0,0 +1,156 @@
...
+    """Choose the degrees of two nodes in a hyperedge.
+
+    Parameters
+    ----------
+    e : iterable
+        the members in a hyperedge
+    k : dict
+        the degrees where keys are node IDs and values are degrees
+    type : str, default: "uniform"
+        the type of degree assortativity
```
leotrs, line 116: List the other options ("top-2" and "top-bottom"). Also, instead of "type", I'd prefer something like "kind", since "type" is a reserved keyword in python.

xgi/algorithms/assortativity.py
```diff
@@ -0,0 +1,156 @@
...
+    Journal of Complex Networks 2020.
+    DOI: 10.1093/comnet/cnaa018
+    """
+    if len(e) > 1:
+        if type == "uniform":
+            i = np.random.randint(len(e))
+            j = i
+            while i == j:
+                j = np.random.randint(len(e))
+            return np.array([k[e[i]], k[e[j]]])
```
leotrs, line 141: Docstring says this returns a tuple, while this line returns an array.

xgi/algorithms/assortativity.py
```diff
@@ -0,0 +1,156 @@
...
+
+        elif type == "top-2":
+            degs = sorted([k[i] for i in e])[-2:]
+            random.shuffle(degs)
+            return degs
+
+        elif type == "top-bottom":
+            degs = sorted([k[i] for i in e])[:: len(e) - 1]
```
leotrs, line 149: I *think* `reversed(sorted(...))` is faster than using `[::-1]`, but I might be wrong?

leotrs (changes requested):

xgi/algorithms/assortativity.py
nwlandry: The method that is currently in place is ~25% faster than that.

nwlandry: I could replace it for readability though

xgi/algorithms/assortativity.py
nwlandry: good catch!

xgi/algorithms/assortativity.py
leotrs: :sob: :sob: :sob: no, leave it, I'll have to profile the stats package..

xgi/algorithms/assortativity.py
nwlandry: Good point. I changed this for a 1.5x speedup for k1 and a 4.5x speedup for k2. I think the list comprehension was the slowest part so that's what I removed.

xgi/algorithms/assortativity.py
nwlandry: okay, good point. I like "kind".

xgi/algorithms/assortativity.py
nwlandry: good call. fixed!

xgi/algorithms/assortativity.py
nwlandry: this was to select the first and last elements but let me double check that

nwlandry: and I believe that l[::-1] has different behavior than l[::n-1]. (super weird though)

nwlandry: okay, confirmed that that's why I did that. it's to select the min and max elements in one line

xgi/algorithms/assortativity.py
nwlandry: Nice. much more pythonic.

xgi/algorithms/assortativity.py
leotrs: Hmmm okay, maybe leave a comment somehwere to that effect? It wasn't clear to me what this line was doing.

nwlandry: done.

leotrs (approved):

nwlandry:
Added the computation of the dynamical, top-2, top-bottom, and uniform degree assortativity.

PR #124, opened 2022-06-17
Refactor edges

nwlandry:
Before:
before

After:
after

### DIFF
--- a/xgi/classes/hypergraph.py
+++ b/xgi/classes/hypergraph.py
@@ -41,6 +41,44 @@ class IDDict(dict):
         except KeyError as e:
             raise IDNotFound(f"ID {item} not found") from e
 
+from collections.abc import MutableMapping
+
+
+class IDDict2(MutableMapping):
+    """A dictionary that applies an arbitrary key-altering
+       function before accessing the keys"""
+
+    def __init__(self, *args, **kwargs):
+        self.store = dict()
+        self.update(dict(*args, **kwargs))
+
+    def __getitem__(self, item):
+        try:
+            return self.store[self._keytransform(item)]
+        except KeyError as e:
+            raise IDNotFound(f"ID {item} not found") from e
+
+    def __setitem__(self, item, value):
+        try:
+            self.store[self._keytransform(item)] = value
+        except KeyError as e:
+            raise IDNotFound(f"ID {item} not found") from e
+
+    def __delitem__(self, item):
+        try:
+            del self.store[self._keytransform(item)]
+        except KeyError as e:
+            raise IDNotFound(f"ID {item} not found") from e
+
+    def __iter__(self):
+        return iter(self.store)
+    
+    def __len__(self):
+        return len(self.store)
+
+    def _keytransform(self, key):
+        return key
+
 
 class Hypergraph:
     r"""A hypergraph is a collection of subsets of a set of *nodes* or *vertices*.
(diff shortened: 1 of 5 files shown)


xgi/classes/hypergraph.py
```diff
@@ -506,7 +506,7 @@ def add_edges_from(self, ebunch_to_add, **attr):
             i.e. you cannot mix different formats.  The iterables containing edge
             members cannot be strings.
 
-        attr : \*\*kwargs, optional
+        attr : **kwargs, optional
```
leotrs, line 509: I made this change because without the backward slashes, Sphinx issues a warning when building the docs.

xgi/classes/reportviews.py
```diff
@@ -391,14 +391,4 @@ def members(self, e=None, dtype=list):
...
-            return self._id_dict[e].copy()
-        except IDNotFound:
-            if e is None:
-                if dtype is dict:
-                    return {key: self._id_dict[key] for key in self._ids}
-                elif dtype is list:
-                    return [self._id_dict[key] for key in self._ids]
-                else:
-                    raise XGIError(f"Unrecognized dtype {dtype}")
-            raise IDNotFound(f"Item {e} not in this view")
```
leotrs, line 404: This is all repeated from above, no? This duplication looks like it might have been the product of a botched merge/rebase (of which I am notoriously guilty). Good catch.

xgi/classes/reportviews.py
```diff
@@ -391,14 +391,4 @@ def members(self, e=None, dtype=list):
...
+        return self._id_dict[e].copy()
```
leotrs, line 394: I'm p sure it's safe to remove this `copy` call. The problem is that we do not want the returned value to be mutable (or, if it is, we do not want it to mutate the underlying structure). For example, some version the following code can be added to the test suite to make sure that it is safe to do so:
```python
H = xgi.Hypergraph([[1,2,3]])
m = H.edges.members(0)
m.append(10)
assert H.edges.members(0) == [1, 2, 3]
```

xgi/classes/hypergraph.py
leotrs: Can you run `make html` from `docs/` to make sure it's not just me?

leotrs (commented):
Good catch there!

xgi/classes/reportviews.py
nwlandry: Thanks! Yeah, I couldn't see a case where it would ever get used, so I removed.

xgi/classes/reportviews.py
nwlandry: If it's okay, I was thinking of adding as an issue and dealing with this a bit later. As is, with the changes that I made, we are currently ~2-2.5x slower than a stripped down dict and we added the copy() method because mutating the returned edge will affect the underlying Hypergraph dictionaries.

xgi/classes/hypergraph.py
nwlandry: yep, you're right!

leotrs (approved):

xgi/classes/reportviews.py
```diff
@@ -74,11 +74,7 @@ def __init__(self, H, id_dict, id_attr, dispatcher, ids=None):
         self._dispatcher = dispatcher
         self._id_dict = id_dict
         self._id_attr = id_attr
-
-        if id_dict is None:
-            self._ids = None
-        else:
-            self._ids = ids
+        self._ids = ids
```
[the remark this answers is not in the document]
nwlandry: Yeah, but the only place it seemed like it was used was in __contains__ and that was handled with an if statement. So, what I did was to keep the if-statement, modify it slightly and remove the if statement from __contains__

PR #125, opened 2022-06-19
add operations for attributes

nwlandry:
This adds additional operations for node/edge attributes.

### DIFF
--- a/xgi/classes/reportviews.py
+++ b/xgi/classes/reportviews.py
@@ -237,7 +237,22 @@ class IDView(Mapping, Set):
         <https://github.com/ComplexGroupInteractions/xgi/blob/main/tutorials/Tutorial%206%20-%20Statistics.ipynb>`_.
 
         """
-        bunch = [idx for idx in self if self._id_attr[idx][attr] == val]
+        if mode == "eq":
+            bunch = [idx for idx in self if self._id_attr[idx][attr] == val]
+        elif mode == "neq":
+            bunch = [idx for idx in self if self._id_attr[idx][attr] != val]
+        elif mode == "lt":
+            bunch = [idx for idx in self if self._id_attr[idx][attr] < val]
+        elif mode == "gt":
+            bunch = [idx for idx in self if self._id_attr[idx][attr] > val]
+        elif mode == "leq":
+            bunch = [idx for idx in self if self._id_attr[idx][attr] <= val]
+        elif mode == "geq":
+            bunch = [idx for idx in self if self._id_attr[idx][attr] >= val]
+        elif mode == "between":
+            bunch = [idx for idx in self if val[0] <= self._id_attr[idx][attr] <= val[1]]
+        else:
+            raise ValueError(f"Unrecognized mode {mode}")
         return type(self).from_view(self, bunch)
 
     @classmethod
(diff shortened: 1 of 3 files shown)


maximelucas:
Looks nice! Safer with a few tests?

xgi/classes/reportviews.py
```diff
@@ -237,7 +237,24 @@ def filterby_attr(self, attr, val, mode="eq"):
         <https://github.com/ComplexGroupInteractions/xgi/blob/main/tutorials/Tutorial%206%20-%20Statistics.ipynb>`_.
```
leotrs, line 238: Notes
--------
Beware of using comparison modes ("lt", "gt", "leq", "geq") when the attribute is a string. For example, the string comparison `'10' < '9'` evaluates to `True`.

xgi/classes/reportviews.py
```diff
@@ -237,7 +237,22 @@ def filterby_attr(self, attr, val, mode="eq"):
...
+        elif mode == "gt":
+            bunch = [idx for idx in self if self._id_attr[idx][attr] > val]
+        elif mode == "leq":
+            bunch = [idx for idx in self if self._id_attr[idx][attr] <= val]
+        elif mode == "geq":
+            bunch = [idx for idx in self if self._id_attr[idx][attr] >= val]
+        elif mode == "between":
+            bunch = [idx for idx in self if val[0] <= self._id_attr[idx][attr] <= val[1]]
+        else:
+            raise ValueError(f"Unrecognized mode {mode}")
```
maximelucas, line 257: Would specify which modes are known:

raise ValueError(f"Unrecognized mode {mode}. mode must be one of ["eq", "neq", "lt", "gt", "leq", "geq", "between"].")

xgi/classes/reportviews.py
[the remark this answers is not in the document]
nwlandry: The two ways I can think to deal with this are 1) exclude these IDs and 2) specify a default value.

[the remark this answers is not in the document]
nwlandry: I've implemented 1) for now.

xgi/classes/reportviews.py
nwlandry: Good call. Fixed here and also in `filterby`

xgi/classes/reportviews.py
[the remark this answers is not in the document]
leotrs: We should handle it in the same way as `H.nodes.attrs` handles it, via the `missing` keyword. you don't have to implement it rn, but please open an issue :)

xgi/classes/reportviews.py
```diff
@@ -225,36 +225,72 @@ def filterby(self, stat, val, mode="eq"):
...
             bunch = [node for node in self if val[0] <= values[node] <= val[1]]
         else:
-            raise ValueError(f"Unrecognized mode {mode}")
+            raise ValueError(
+                f"Unrecognized mode {mode}. mode must be one of 'eq', 'neq', 'lt', 'gt', 'leq', 'geq', or 'between'."
+            )
         return type(self).from_view(self, bunch)
 
-    def filterby_attr(self, attr, val, mode="eq"):
+    def filterby_attr(self, attr, val, mode="eq", default=None):
```
leotrs, line 233:
```suggestion
    def filterby_attr(self, attr, val, mode="eq", missing=None):
```

leotrs (approved):
Just change `default` to `missing` to it matches `H.nodes.attrs` (see `def attrs` in the `nodestats.py` module) and then you can merge.

xgi/classes/reportviews.py
nwlandry: I liked your suggestion to match the interface so I implemented it now instead of adding it as an issue

PR #127, opened 2022-06-20
Function to relabel nodes and edges

nwlandry:
This implements a function very similar to `networkx.convert_node_labels_to_integers`. but for both node and edge IDs.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -161,3 +161,11 @@ def hyperwithattrs(edgelist4, attr1, attr2, attr3, attr4, attr5):
         ]
     )
     return H
+
+
+@pytest.fixture
+def hypergraph1():
+    H = xgi.Hypergraph()
+    H.add_nodes_from(["a", "b", "c"])
+    H.add_edges_from({"e1": ["a", "b"], "e2": ["a", "b", "c"], "e3": ["c"]})
+    return H
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -174,6 +174,6 @@ def hypergraph1():
 @pytest.fixture
 def hypergraph2():
     H = xgi.Hypergraph()
-    H.add_nodes_from([0, "b", "c"])
+    H.add_nodes_from(["b", "c", 0])
     H.add_edges_from({"e1": [0, "b"], "e2": [0, "c"], "e3": [0, "b", "c"]})
     return H
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -169,3 +169,11 @@ def hypergraph1():
     H.add_nodes_from(["a", "b", "c"])
     H.add_edges_from({"e1": ["a", "b"], "e2": ["a", "b", "c"], "e3": ["c"]})
     return H
+
+
+@pytest.fixture
+def hypergraph2():
+    H = xgi.Hypergraph()
+    H.add_nodes_from([0, "b", "c"])
+    H.add_edges_from({"e1": [0, "b"], "e2": [0, "c"], "e3": [0, "b", "c"]})
+    return H
--- a/tests/utils/test_utils.py
+++ b/tests/utils/test_utils.py
@@ -27,24 +27,44 @@ def test_get_dual(dict5):
     assert dual[8] == [3]
 
 
-def test_convert_labels_to_integers(hypergraph1):
-    H = convert_labels_to_integers(hypergraph1)
+def test_convert_labels_to_integers(hypergraph1, hypergraph2):
+    H1 = convert_labels_to_integers(hypergraph1)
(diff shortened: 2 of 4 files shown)


tests/utils/test_utils.py
```diff
@@ -25,3 +25,26 @@ def test_get_dual(dict5):
     assert dual[6] == [2, 3]
     assert dual[7] == [3]
     assert dual[8] == [3]
+
+
+def test_convert_labels_to_integers(hypergraph1):
```
leotrs, line 30: Can you please add another test where the original hypergraph has nodes/edges labeled by non-consecutive integers? And/Or, a test where the original hypergraph has a single node labeled `0` but all other nodes are labeled by strings.

xgi/utils/utilities.py
```diff
@@ -73,3 +74,51 @@ def load_xgi_data(dataset, nodetype=None, edgetype=None):
...
+        temp_H._node[id] = temp_H._node.pop(node)
+        temp_H._node_attr[id] = temp_H._node_attr.pop(node)
+        temp_H._node_attr[id].update({"label": node})
+        node_mapping[node] = id
+
+    for edge in list(temp_H.edges):
+        id = next(edge_id)
+        temp_H._edge[id] = temp_H._edge.pop(edge)
+        temp_H._edge_attr[id] = temp_H._edge_attr.pop(edge)
+        temp_H._edge_attr[id].update({"label": edge})
```
leotrs, line 115: Why this instead of `temp_H._edge_attr[id]["label"] = edge`?

xgi/utils/utilities.py
```diff
@@ -73,3 +74,51 @@ def load_xgi_data(dataset, nodetype=None, edgetype=None):
...
+    the old "label" values (if they exist) will be overwritten.
+    """
+    node_id = count()
+    edge_id = count()
+    node_mapping = dict()
+    edge_mapping = dict()
+    temp_H = H.copy()
+    for node in list(temp_H.nodes):
+        id = next(node_id)
+        temp_H._node[id] = temp_H._node.pop(node)
```
leotrs, line 106: I'm pretty sure this is buggy. For example, if the original hypergraph has a node with id `0` (but other nodes have ids that are, say, strings), the first time this line is executed, the edges incident to `0` will be overwritten and can never be recovered.

My suggestion is to first build the `node_mapping` and `edge_mapping` dictionaries WITHOUT modifying the network, and then using those mappings to do the relabeling in a single loop. In this loop, you are trying to build the `node_mapping` at the same time as modifying the network, and that's what's causing the problem.

tests/utils/test_utils.py
leotrs: This will catch the bug I point out later.

leotrs (changes requested):

xgi/utils/utilities.py
nwlandry: Good call.

xgi/utils/utilities.py
nwlandry: This is a good suggestion. Done.

xgi/utils/utilities.py
```diff
@@ -73,3 +74,51 @@ def load_xgi_data(dataset, nodetype=None, edgetype=None):
     r = requests.get(index[dataset]["url"])
 
     return convert.dict_to_hypergraph(r.json(), nodetype=nodetype, edgetype=edgetype)
+
+
+def convert_labels_to_integers(H):
```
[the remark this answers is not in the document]
nwlandry: Done.

tests/utils/test_utils.py
nwlandry: Done.

tests/conftest.py
```diff
@@ -169,3 +169,11 @@ def hypergraph1():
     H.add_nodes_from(["a", "b", "c"])
     H.add_edges_from({"e1": ["a", "b"], "e2": ["a", "b", "c"], "e3": ["c"]})
     return H
+
+
+@pytest.fixture
+def hypergraph2():
+    H = xgi.Hypergraph()
+    H.add_nodes_from([0, "b", "c"])
```
leotrs, line 177: I think the right test would be to add `["b", "c", 0]` i.e. the `0` last, so the (old) code would have overwritten `H._node[0] = H._node["b"]` and lost the neighbors of `0`. In this test, `0` is added first so it just gets relabeled to itself and there is no bug.

leotrs: The new code looks better tho and I can't spot any bugs if any.

leotrs (approved):

leotrs (approved):

Issue #98, opened 2022-04-22
The current "egonet" method is wrong


leotrs:
The current method returns all edges a node belongs to. This is not called the egonet, it should be called something else. The egonet includes edges among nodes that are neighbors of the specified node.

nwlandry:
What if we call this `edge_neighborhood`? or something like that?

I think a way that one could get the `egonet` might be something like `xgi.subhypergraph(H, H.neighbors(n) + {n})`

leotrs:
I think what the current egonet does can simply be retrieved by `H.nodes.memberships(n)`? So no need to call it something else I think.

> I think a way that one could get the egonet might be something like xgi.subhypergraph(H, H.neighbors(n) + {n})

This seems reasonable! Sometimes the egonet of `n` contains `n` and sometimes it doesn't, so the `+{n}` at the end should be conditional on some user-supplied parameter.

PR #129, opened 2022-06-28
Rename egonet and move the neighbors method to the IDView class

nwlandry:
I tried implementing egonet as described in #98 but using `subhypergraph` in the implementation creates a circular import (because `subhypergraph` uses `freeze`)

### DIFF
--- a/xgi/classes/reportviews.py
+++ b/xgi/classes/reportviews.py
@@ -38,7 +38,7 @@ class IDView(Mapping, Set):
 
     """
 
-    __slots__ = ("_dispatcher", "_id_dict", "_id_attr", "_ids")
+    __slots__ = ("_dispatcher", "_id_dict", "_id_attr", "_bi_id_dict", "_bi_id_attr", "_ids")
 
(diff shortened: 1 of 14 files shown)


xgi/classes/reportviews.py
```diff
@@ -38,7 +38,14 @@ class IDView(Mapping, Set):
...
     """
 
-    __slots__ = ("_dispatcher", "_id_dict", "_id_attr", "_ids")
+    __slots__ = (
+        "_dispatcher",
+        "_id_dict",
+        "_id_attr",
+        "_bi_id_dict",
+        "_bi_id_attr",
+        "_ids",
```
leotrs, line 47: At this point we might want to consider (in the future) just passing the entire network...

leotrs (approved):

xgi/classes/reportviews.py
nwlandry: haha, good point. In the future perhaps the whole network and an indicator whether it's for nodes or edges?

Issue #138, opened 2022-07-19
Allow the colors parameters in the `draw` function to accept dictionary or array of colors


mcontisc:
So far, the colors parameters in the `draw` function accept only one string value. However, I would find useful to have the same format as the networkx function where a practitioner can specify ideally a different color for every node/edge.

leotrs:
Hi Martina! Thank you very much for reporting this issue. You are right, this would be a desirable feature. Let me take a look!

leotrs:
@mcontisc please check out PR #139! Let me know if the image below looks like what you had in mind. 
![2022-07-20-084651_1518x633_scrot](https://user-images.githubusercontent.com/1096704/179915519-cb23b475-7121-40d0-b1ff-e2112272e612.png)

leotrs:
@mcontisc Thanks for opening this issue. It has now been implemented. 

We are always looking for collaborators, so please feel free to keep opening issues, or better yet, joining us in our Zulip channel you can find [here](https://xgi.readthedocs.io/en/latest/about.html#contributing).

mcontisc:
Thank you for the timely and friendly support :) 
I will keep an eye on the library and have in mind the channel.

PR #139, opened 2022-07-20
Fix 138: allow xgi.draw to set color of individual nodes or edges

leotrs:
See #138. Also see attached image, ran from this branch.
2022-07-20-084651_1518x633_scrot

### DIFF
--- a/xgi/drawing/xgi_pylab.py
+++ b/xgi/drawing/xgi_pylab.py
@@ -35,8 +35,7 @@ def draw(
     node_lw=1,
     node_size=0.03,
 ):
-    """
-    Draw hypergraph or simplicial complex.
+    """Draw hypergraph or simplicial complex.
 
     Parameters
     ----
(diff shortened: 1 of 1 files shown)


mcontisc:
This works, yes. However, it still gives problem when you pass an array because the access for arrays is also through node id and not through node position.

nwlandry:
How about accepting a list or array and mapping indices to nodes like this?
```
# colors is an input array
nodes = list(H.nodes)
color_dict = {nodes[i]: colors[i] for i in range(H.num_nodes)}
```

a more succinct way to do that would be `color_dict = dict(zip(H.nodes, colors))`.

leotrs:
@nwlandry I actually don't like that one because it would effectively be like accepting a de-facto second node ID. The user can already access the node addition order by just looping over H.nodes.

leotrs:
Done! 
2022-07-21-103153_808x440_scrot

mcontisc:
Yes, I agree :) Thanks a lot!
Out of curiosity, this is not possible for the hyperedges right? I mean, for coloring them we can only use a colormap. Correct?

leotrs:
For now, yes. It would be nice to have a way to override the color assigned to a hyperedge via some other parameter, cc @iaciac.

nwlandry:
Looks great!! I think it's ready to merge! I can add the coloring individual edges as an issue.

Issue #140, opened 2022-07-21
Color by individual edges


nwlandry:
It would be nice to have a way to override the color assigned to a hyperedge via some other parameter. (From discussion in #139)

nwlandry:
Fixed with #142

PR #142, opened 2022-07-22
Fix drawing

### LINKED ISSUE #137: Changing plot size changes node aspect ratio in `xgi.draw()`

nwlandry:
This fixes #137 and #140. Among other things, user can specify
* node size, node line color, node face color, node line width, edge line width, and edge color as dictionaries, iterables, or strings.
* edge face color as a colormap, dictionary, iterable or string.

The `draw` function is now modularized into `draw_xgi_nodes`, `draw_xgi_hyperedges`, and `draw_xgi_complexes` as well as a few helper functions. In addition, nodes are now drawn with `scatter` not `Circle` objects.

### DIFF
--- a/tests/generators/test_classic.py
+++ b/tests/generators/test_classic.py
@@ -10,8 +10,12 @@ def test_empty_hypergraph():
     assert (H.num_nodes, H.num_edges) == (0, 0)
 
 
-def test_star_clique():
+def test_empty_hypergraph():
+    SC = xgi.empty_simplicial_complex()
+    assert (SC.num_nodes, SC.num_edges) == (0, 0)
 
+
+def test_star_clique():
     with pytest.raises(ValueError):
         H = xgi.star_clique(-1, 7, 3)
     with pytest.raises(ValueError):
--- a/tutorials/Tutorial 5 - Plotting.ipynb
+++ b/tutorials/Tutorial 5 - Plotting.ipynb
@@ -106,13 +106,13 @@
     "cmap = plt.cm.Blues\n",
     "\n",
     "ax = plt.subplot(1,2,1)\n",
-    "xgi.draw(H, pos, cmap=cmap, ax=ax)\n",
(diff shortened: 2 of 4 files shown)


nwlandry:
One small issue is that I get a 
```
'c' argument looks like a single numeric RGB or RGBA sequence, which should be avoided as value-mapping will have precedence in case its length matches with 'x' & 'y'. Please use a 2-D array with a single row if you really want to specify the same RGB or RGBA value for all points.
```
error which I had to fix temporarily with the following code as a hack:
```
from matplotlib.axes._axes import _log as matplotlib_axes_logger
matplotlib_axes_logger.setLevel('ERROR')
``

leotrs:
Haven't looked at the code yet but can you share what are the reasons why nodes are now drawn with scatter and not circles?

leotrs:
> error which I had to fix temporarily with the following code as a hack:

why silence the warning instead of following the suggestion? It looks like we have to pass `np.array(["c"]).reshape(1, 1)` instead of just "c".

nwlandry:
> Haven't looked at the code yet but can you share what are the reasons why nodes are now drawn with scatter and not circles?

The reason is because it was the easiest way to fix #137.

nwlandry:
> > error which I had to fix temporarily with the following code as a hack:
> 
> why silence the warning instead of following the suggestion? It looks like we have to pass `np.array(["c"]).reshape(1, 1)` instead of just "c".

Agreed. The issue is that this is only an issue for colors with one specific datatype and I'm not sure the best way to handle this. Perhaps my `_arg_to_dict()` function is doing too much?

leotrs:
Haven't looked at the code yet but could you share the reasons why nodes
are now drawn with scatter instead of circles?

nwlandry:
The reason why is because scatter preserves the aspect ratio of the markers regardless of the figure size without any extra code.

PR #143, opened 2022-07-26
Added ability to draw NodeStats and EdgeStats

nwlandry:
Users can now specify NodeStats or EdgeStats as drawing parameters in addition to specifying them directly with a single value, an iterable or a dictionary. Also fixed `CEC_centrality` so that it outputs a dict of floats, not a dict of 1D numpy arrays.

### DIFF
--- a/xgi/drawing/xgi_pylab.py
+++ b/xgi/drawing/xgi_pylab.py
@@ -56,32 +56,45 @@ def draw(
     ax : matplotlib.pyplot.axes (default=None)
 
     edge_lc : color (str, dict, or iterable, default='black')
-    Color of the edges (dyadic links and borders of the hyperedges).  If str, use the
-    same color for all edges.  If a dict, must contain (edge_id: color_str) pairs.  If
-    iterable, assume the colors are specified in the same order as the edges are found
-    in H.edges.
+        Color of the edges (dyadic links and borders of the hyperedges).  If str, use the
+        same color for all edges.  If a dict, must contain (edge_id: color_str) pairs.  If
+        iterable, assume the colors are specified in the same order as the edges are found
+        in H.edges.
 
     edge_lw :  float (default=1.5)
-    Line width of edges of order 1 (dyadic links).
+        Line width of edges of order 1 (dyadic links).
 
     edge_fc : str, 4-tuple, ListedColormap, LinearSegmentedColormap, or dict of 4-tuples or strings
-    Color of hyperedges
+        Color of hyperedges
 
     node_fc : color (str, dict, or iterable, default='white')
-    Color of the nodes.  If str, use the same color for all nodes.  If a dict, must
-    contain (node_id: color_str) pairs.  If other iterable, assume the colors are
-    specified in the same order as the nodes are found in H.nodes.
+        Color of the nodes.  If str, use the same color for all nodes.  If a dict, must
+        contain (node_id: color_str) pairs.  If other iterable, assume the colors are
+        specified in the same order as the nodes are found in H.nodes.
 
     node_ec : color (dict or str, default='black')
-    Color of node borders.  If str, use the same color for all nodes.  If a dict, must
-    contain (node_id: color_str) pairs.  If other iterable, assume the colors are
-    specified in the same order as the nodes are found in H.nodes.
+        Color of node borders.  If str, use the same color for all nodes.  If a dict, must
+        contain (node_id: color_str) pairs.  If other iterable, assume the colors are
+        specified in the same order as the nodes are found in H.nodes.
 
     node_lw : float (default=1.0)
-    Line width of the node borders in pixels.
+        Line width of the node borders in pixels.
 
     node_size : float (default=0.03)
-    Radius of the nodes in pixels
+        Radius of the nodes in pixels
+
+    **kwargs : optional args
+        alternate default values. Values that can be overwritten are the following:
+        * min_node_size
+        * max_node_size
+        * min_edge_linewidth
+        * max_edge_linewidth
+        * min_node_linewidth
+        * max_node_linewidth
+        * node_colormap
+        * node_outline_colormap
+        * edge_face_colormap
+        * edge_outline_colormap
 
     Examples
     --------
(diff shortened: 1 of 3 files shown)


xgi/drawing/xgi_pylab.py
```diff
@@ -14,6 +14,8 @@
 from matplotlib import cm
 from matplotlib.colors import LinearSegmentedColormap, ListedColormap
 
+from xgi.stats import EdgeStat, NodeStat
```
leotrs, line 17: Question: did you try a relative import and it didn't work, or did you not try a relative import? :)

xgi/drawing/xgi_pylab.py
```diff
@@ -54,29 +56,45 @@ def draw(
...
+        Radius of the nodes in pixels
+
+    **kwargs : optional args
+        alternate default values. Values that can be overwritten are the following:
+        * min_node_size
+        * max_node_size
+        * min_edge_linewidth
+        * max_edge_linewidth
+        * min_node_linewidth
+        * max_node_linewidth
```
leotrs, line 93: Nitpick: since the other parameter is called `edge_lw`, this should be `min_edge_lw` and so on.

leotrs (commented):
Looking good! We really need to figure out how to test plotting tho...

xgi/drawing/xgi_pylab.py
```diff
@@ -54,29 +56,45 @@ def draw(
     ax : matplotlib.pyplot.axes (default=None)
 
     edge_lc : color (str, dict, or iterable, default='black')
-    Color of the edges (dyadic links and borders of the hyperedges).  If str, use the
-    same color for all edges.  If a dict, must contain (edge_id: color_str) pairs.  If
-    iterable, assume the colors are specified in the same order as the edges are found
-    in H.edges.
+        Color of the edges (dyadic links and borders of the hyperedges).  If str, use the
```
maximelucas, line 59: I get the naming of edge_lc, edge_lw, and edge_fc, but at first I found it a bit confusing. Maybe we should at least explicitly expand and refer to the abbrevations (lc, fc, lw) in the docsting? Maybe also allow for a long version of these arguments?
Ideas that I don't think are great (but can't find betterones):
"(Line) color of the edges" or "Color of the (line) edges". 

Or change the names?
edge_dyadic_c, edge_dyadic_lw, edge_c
or
edge_c, edge_lw, hyperedge_c

maximelucas: Also, does edge_lc actually color "borders of the hyperedges"? It's not so obvious from the viz examples. Or hyperedges have a much thiner lw that is not argument.

xgi/drawing/xgi_pylab.py
```diff
@@ -152,28 +193,32 @@ def draw_xgi_nodes(ax, H, pos, node_fc, node_ec, node_lw, node_size, zorder):
...
+        Pairwise edge widths
     edge_fc : str, 4-tuple, ListedColormap, LinearSegmentedColormap, or dict of 4-tuples or strings
-        color of hyperedges
+        Color of hyperedges
+    settings : dict
+        Default parameters
     """
-    edge_lc = _arg_to_dict(edge_lc, H.edges)
-    edge_lw = _arg_to_dict(edge_lw, H.edges)
+    edge_lc = _color_arg_to_dict(edge_lc, H.edges, settings["edge_outline_colormap"])
```
maximelucas, line 216: Maybe first check that it is not a simplicial complex
if isinstance(H, SimplicalComplex):
    raise ValueError("Use draw_xgi_complexes() instead")

xgi/drawing/xgi_pylab.py
```diff
@@ -205,29 +250,38 @@ def draw_xgi_hyperedges(ax, H, pos, edge_lc, edge_lw, edge_fc, d_max):
             ax.add_patch(obj)
 
 
-def draw_xgi_complexes(ax, SC, pos, edge_lc, edge_lw, edge_fc):
+def draw_xgi_complexes(ax, SC, pos, edge_lc, edge_lw, edge_fc, settings):
```
maximelucas, line 253: "draw_xgi_simplices" instead?
also push "ax" to 3rd place to match draw()

xgi/drawing/xgi_pylab.py
```diff
@@ -152,28 +193,32 @@ def draw_xgi_nodes(ax, H, pos, node_fc, node_ec, node_lw, node_size, zorder):
         )
 
 
-def draw_xgi_hyperedges(ax, H, pos, edge_lc, edge_lw, edge_fc, d_max):
+def draw_xgi_hyperedges(ax, H, pos, edge_lc, edge_lw, edge_fc, d_max, settings):
```
maximelucas, line 196: Push "ax" to 3rd place to match draw()

xgi/drawing/xgi_pylab.py
```diff
@@ -54,29 +56,45 @@ def draw(
     ax : matplotlib.pyplot.axes (default=None)
```
maximelucas, line 56: Just above: cmap is described in the docstring but it is not an argument of draw() anymore

xgi/drawing/xgi_pylab.py
maximelucas: Describe possible keys for settings?
It's described in draw(), but people might want to use draw_xgi_hyperedges() and draw_xgi_nodes() independently.
Also, right now, I have to have all keys in settings, right? Want if I only want to change one, say the cmap.

maximelucas:
All in all great, just lots of small detail questions!

xgi/drawing/xgi_pylab.py
nwlandry: Fixed, thanks!

xgi/drawing/xgi_pylab.py
nwlandry: good call!

xgi/drawing/xgi_pylab.py
nwlandry: Done.

xgi/drawing/xgi_pylab.py
nwlandry: Good idea. Fixed!

xgi/drawing/xgi_pylab.py
nwlandry: fixed.

xgi/drawing/xgi_pylab.py
nwlandry: Good question. I think I'm in favor of edge_lc only applying to dyadic links. And in the reference to the names, what about changing `edge_lc` to `linkcolor` or `dyadcolor`, `edge_lw` to `edge_linewidth`, etc. and actually expand the names? It's a balance between readability and brevity and I'm not sure what is best.

maximelucas: I agree with edge_lc only applying to dyadic links. I checked in the code, right now `edge_lc` is also used for the border of hyperedges, but with fixed small width of 0.5. To draw hypergraphs, I would maybe replace and use `edge_fc` instead. (We could even have an extra argument specifically for that.. 😛 )

Re. naming, yea why not, then
- `dyad_color`and `dyad_lw` for dyads
- `edge_fc` (and potentially `edge_lc`) for other hyperedges
It would introduce the word dyad to XGI, which might not be the best, but I don't have a better idea to distinguish pairwise edges and others. 

Re. expanding, I'm not sure what is best either. Maybe keep the abbreviated version for the actual argument, but expand in the docstring explanation:
- `edge_fc : Face color of the edges (of more than 2 nodes)` 

(matplotlib allows both lw and linewidth as argument, we could have the same?)

nwlandry: Done. Good point.

PR #144, opened 2022-07-28
Make filterby() accept NodeStat and EdgeStat objects 

### LINKED ISSUE #135: Cannot filterby using a nodestat
```python
n = H.nodes

# get nodes with degree equal to 2
n.filterby("degree", 2)

# compute degree-of-a-given-order
n.degree(order=2).asdict()

# this SHOULD get nodes with degree-of-a-given-order equal to 2, but right now gives an error
n.filterby(n.degree(order=2), 2)
# -> TypeError
```

leotrs:
Previously, it only accepted strings that were the name of valid NodeStat or EdgeStat, so this is a minor change.

Also fixed a bug in how the order of an edge was being computed.

Also added docs.

Fixes #135


nwlandry (approved):
Looks great!

PR #145, opened 2022-07-28
Refactor Hypergraph.copy()

### LINKED ISSUE #128: Hypergraph.copy needs to deepcopy hypergraph attributes
```python
>>> import xgi
>>> H = xgi.Hypergraph()
>>> H['foo'] = 1
>>> H._hypergraph
{'foo': 1}
>>> K = H.copy()
>>> K._hypergraph
{'foo': 1}
>>> K['foo'] = 1000
>>> K._hypergraph
{'foo': 1000}
>>> H._hypergraph
{'foo': 1000} # woops
```

leotrs:
Multiple things were Wrong with `copy()`:

1. It was using `subhypergraph()` which returns a frozen hypergraph (yikes!). I'm guilty of this one.
2. See #128 
3. A naive implementation of `copy()` would have had a bug when using `add_edge`. I fixed it in the current PR. See comment starting on line 944 in `hypergraph.py`.

Fixes #128.

### DIFF
--- a/xgi/classes/hypergraph.py
+++ b/xgi/classes/hypergraph.py
@@ -950,7 +950,8 @@ class Hypergraph:
 
         # Then, we set the start at one plus the maximum edge ID that is an integer,
         # because count() only yields integer IDs.
-        copy._edge_uid = count(start=max(edges_with_int_id) + 1)
+        start = max(edges_with_int_id) + 1 if edges_with_int_id else 0
+        copy._edge_uid = count(start=start)
 
         return copy
 
--- a/xgi/classes/hypergraph.py
+++ b/xgi/classes/hypergraph.py
@@ -910,8 +910,49 @@ class Hypergraph:
         H : Hypergraph
             A copy of the hypergraph.
 
+        Notes
+        -----
+
+        There is no guarantee that performing similar operations on a hypergraph and its
+        copy after the copy is made will yield the same results.  For example,
+
+        >>> import xgi
+        >>> H = xgi.Hypergraph([[1, 2, 3], [4], [5, 6], [6, 7, 8]])
+        >>> H.add_edge([1, 3, 5], id=10)
+        >>> K = H.copy()
+        >>> H.add_edge([2, 4]); K.add_edge([2, 4]);
+        >>> list(H.edges) == list(K.edges)
+        False
+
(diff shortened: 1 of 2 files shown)


xgi/classes/hypergraph.py
```diff
@@ -910,8 +910,50 @@ def copy(self):
...
+        EdgeView((0, 1, 2, 3, 10, 11))
+
         """
-        return subhypergraph(self)
+        copy = self.__class__()
+        nn = self.nodes
+        copy.add_nodes_from((n, deepcopy(attr)) for n, attr in nn.items())
+        ee = self.edges
+        copy.add_edges_from(
+            (ee.members(e), e, deepcopy(attr)) for e, attr in ee.items()
```
nwlandry, line 940: This all looks good, but I'm wondering if might be faster to call members once:
```
copy.add_edges_from((e, id, deepcopy(self.edges[id])) for id, e in ee.members(dtype=dict))
```

xgi/classes/hypergraph.py
```diff
@@ -910,8 +910,50 @@ def copy(self):
...
+        )
+        copy._hypergraph = deepcopy(self._hypergraph)
+
+        # If we don't set the start of copy._edge_uid correctly, it will start at 0,
+        # which will overwrite any existing edges when calling add_edge().  First, we
+        # use the somewhat convoluted float(e).is_integer() instead of using
+        # isinstance(e, int) because there exist integer-like numeric types (such as
+        # np.int32) which fail the isinstance() check.
+        edges_with_int_id = [e for e in self.edges if float(e).is_integer()]
```
nwlandry, line 949: This is really clever and I definitely wouldn't have thought of it. I wonder if we should be handling this in the standard `add_edge` method as well.

nwlandry (approved):
Looks good. Some minor comments, possibly some future fixes.

xgi/classes/hypergraph.py
leotrs: For more context, read this thread.

xgi/classes/hypergraph.py
leotrs: Good catch, fixed!

nwlandry (approved):

Issue #94, opened 2022-04-21
Move some methods away from the base classes


leotrs:
The base classes are already bloated as it is, so we should try to do some cleanup. I propose the base classes only keep methods that directly modify the underlying structure. All other methods should go to auxiliary classes. For example:

These should move to `NodeView`:
* isolates
* neighbors

These should move to `EdgeView`:
* singleton_edges (and rename to singletons)
* max_edge_order (and rename to max_order)
* duplicate_edges (and rename to duplicates)
* is_possible_order (do we really need this method?)

For example, instead of `H.neighbors`, we would do `H.nodes.neighbors`. Yes this is longer but I think it's worth it. Maintenance, documentation, and end-user code will be better.

These methods should be removed completely:
* get_edge_data (data can be accessed via View objects)
* has_edge (do `n in H.edges` instead)
* has_node (do `n in H.nodes` instead)

Issue #36, opened 2022-01-13
Implement "cleanup" method


leotrs:
~Make sure the documentation highlights the fact that the user should remember to call `remove_isolates` and other related functions.~
Implement a `cleanup` method that performs various common clean up tasks, see comment down thread.

leotrs:
#39 implements `remove_isolates`

leotrs:
Other ideas for 2 `cleanup` method: remove isolates, extract largest component, remove duplicate/multi edges, relabel the node ids.

nwlandry:
So you're saying that we have a cleanup method that does these four steps? I like that idea and also we should add the possibility to collapse loopy hyperedges.

leotrs:
> So you're saying that we have a cleanup method that does these four steps?

We don't have it yet, but I want to implement it.

> I like that idea and also we should add the possibility to collapse loopy hyperedges.

Good catch!

leotrs:
For now I think it'd be nice to have something like the following:
```
class Hypergraph:
    ...
    def cleanup(self):
        self.remove_isolates()
        self.remove_singletons()
        self.collapse_loopy_edges()    # not implemented yet
        self.remove_duplicates()     # not implemented yet
        self.relabel_nodes()     # not implemented yet
    ...
```

What other operations should `cleanup` perform?

leotrs:
Closed by #161

PR #146, opened 2022-07-28
Cleanup hypergraph methods

leotrs:
Continued progress on #94 

1. `H.isolates` is now `H.nodes.isolates` and returns a `NodeView`.
2. `H.singleton_edges` is now `H.edges.singletons` and returns an `EdgeView`.

There are two related convenience methods, `H.remove_isolates` and `H.remove_singleton_edges` that I have left for now, though I think they should both go in the future in favor of a general cleanup method (see #36).

### DIFF
--- a/xgi/classes/reportviews.py
+++ b/xgi/classes/reportviews.py
@@ -429,6 +429,38 @@ class NodeView(IDView):
         """
         return self._id_dict.copy() if n is None else self._id_dict[n].copy()
 
+    def isolates(self, ignore_singletons=True):
+        """Nodes that belong to no edges.
+
+        When ignore_singletons is True (default), a node is considered isolated from the
+        rest of the hypergraph when it is included in no edges of size two or more.  In
+        particular, whether the node is part of any singleton edges is irrelevant to
+        determine whether it is isolated.
+
+        When ignore_singletons is False, a node is isolated only when it is a member of
+        exactly zero edges, including singletons.
+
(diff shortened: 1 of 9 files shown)


nwlandry:
Two thoughts:
1) We already have `remove_nodes_from` and `remove_edges_from` so it wouldn't be too terrible to get rid of `remove_isolates` and `remove_singleton_edges`, since we could just do `H.remove_nodes_from(H.nodes.isolates)` and `H.remove_edges_from(H.edges.singletons)`
2) I would update the docs to match the changed location of these methods.

nwlandry (changes requested):
See comment.

xgi/classes/reportviews.py
```diff
@@ -497,3 +533,17 @@ def members(self, e=None, dtype=list):
             raise IDNotFound(f'ID "{e}" not in this view')
 
         return self._id_dict[e].copy()
+
+    def singletons(self):
```
nwlandry, line 537: would it be worth it to make this a property?

leotrs: I'm leaning toward keeping it as a method. Reason being that properties are expected to be fast and it's unclear to me how fast this is. Also, some users might expect a property to be set-able and this one definitely won't be.

leotrs:
Done!

nwlandry (approved):

PR #147, opened 2022-08-02
Add unit tests

nwlandry:
Added some unit tests and modified the api when needed.


tests/readwrite/test_bipartite_edgelist.py
```diff
@@ -1,8 +1,10 @@
 import tempfile
+from typing import Type
```
leotrs, line 2: Where is this being used?

xgi/classes/hypergraph.py
```diff
@@ -24,16 +24,16 @@ class IDDict(dict):
...
 
     def __setitem__(self, item, value):
         if item is None:
             raise XGIError("None cannot be a node or edge")
         try:
             return dict.__setitem__(self, item, value)
-        except KeyError as e:
-            raise IDNotFound(f"ID {item} not found") from e
+        except TypeError:
+            raise TypeError(f"ID {item} not a valid type")
```
leotrs, line 36: Why the change?

xgi/readwrite/json.py
```diff
@@ -80,10 +80,8 @@ def read_hypergraph_json(path, nodetype=None, edgetype=None):
             if nodetype is not None:
                 try:
                     id = nodetype(id)
-                except Exception as e:
-                    raise TypeError(
-                        f"Failed to convert edge IDs to type {nodetype}."
-                    ) from e
+                except ValueError:
+                    raise TypeError(f"Failed to convert edge IDs to type {nodetype}.")
```
leotrs, line 84: Why not `from e` anymore?

xgi/drawing/xgi_pylab.py
```diff
@@ -193,12 +191,8 @@ def draw_xgi_nodes(
...
     """
     # Note Iterable covers lists, tuples, ranges, generators, np.ndarrays, etc
-    node_fc = _color_arg_to_dict(
-        node_fc, H.nodes, settings["node_fc_cmap"]
-    )
-    node_ec = _color_arg_to_dict(
-        node_ec, H.nodes, settings["node_lc_cmap"]
-    )
+    node_fc = _color_arg_to_dict(node_fc, H.nodes, settings["node_fc_cmap"])
+    node_ec = _color_arg_to_dict(node_ec, H.nodes, settings["node_lc_cmap"])
```
leotrs, line 195: Did you change black settings? Or were these lines previously not formatted with black? We should make sure we all have the same black settings.

xgi/linalg/matrix.py
```diff
@@ -59,48 +59,41 @@ def incidence_matrix(
     node_dict = dict(zip(node_ids, range(num_nodes)))
     edge_dict = dict(zip(edge_ids, range(num_edges)))
 
-    if node_dict and edge_dict:
```
leotrs, line 62: As far as I can tell, the main change in this block is the deletion of this `if` statement. Could you please explain why it is no longer necessary?

leotrs (changes requested):
Looks good, just minor questions/nitpicks.

tests/readwrite/test_bipartite_edgelist.py
nwlandry: Good call - it's not.

tests/readwrite/test_edgelist.py
```diff
@@ -50,6 +50,8 @@ def test_parse_edgelist():
         [1, 4, 7, 8],
         [2, 3],
     ]
+    with pytest.raises(TypeError):
+        xgi.parse_edgelist(["test 2", "2 3 4", "test 4 7 8", "2 3"], nodetype=int)
```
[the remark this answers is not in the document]
nwlandry: Done.

xgi/classes/hypergraph.py
nwlandry: I guess this is a question of best practices. Even if `e` is not used, should we still pass it in?

xgi/drawing/xgi_pylab.py
nwlandry: I updated the black version in the developer requirements a few months ago, so maybe that's it?

xgi/readwrite/json.py
nwlandry: See other comment.

xgi/linalg/matrix.py
nwlandry: Every test that I wrote couldn't access that part of the code because when the hypergraph was empty, it would only hit the if-statement at the beginning. This moves all the handling of empty hypergraphs to the beginning and I've changed the beginning if-statement to now handle no nodes as well (before it just handled no edges)

xgi/classes/hypergraph.py
leotrs: But it is being used! `raise ... from e` is super important since it prints a longer stacktrace.

xgi/linalg/matrix.py
leotrs: That makes sense, thanks.

xgi/classes/hypergraph.py
nwlandry: Okay, I reverted it back.

leotrs (approved):

PR #150, opened 2022-08-18
Cleanup hypergraph methods

### LINKED ISSUE #93: Implement method that returns nodes with the same memberships
Analogous to `duplicate_edges`, which returns edges with the same members

### LINKED ISSUE #94: Move some methods away from the base classes

leotrs:
Final PR regarding #94 .

1. `Hypergraph.has_edge` has become `IDView.lookup`.
2. `Hypergraph.duplicate_edges` has been become `IDView.duplicates`.
3. I had to move `utilities.convert_labels_to_integer` to `function.convert_labels_to_integer` due to relative import weirdness.
4. A `IDView` behaves like a list when coercing it to bool (for example we can do `if view: ...`, and `view` will evaluate to `False` if it is empty). This was already implemented, but I added a test for it.

Closes #94.
Closes #93.

### DIFF
--- a/docs/source/api/classes/xgi.classes.hypergraph.Hypergraph.rst
+++ b/docs/source/api/classes/xgi.classes.hypergraph.Hypergraph.rst
@@ -46,11 +46,3 @@
 
       ~Hypergraph.copy
       ~Hypergraph.dual
-
-
-   .. rubric:: Methods that query nodes and edges
-
-   .. autosummary::
-      :nosignatures:
-
-      ~Hypergraph.has_edge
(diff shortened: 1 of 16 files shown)


nwlandry (approved):
This all looks good to me!

PR #151, opened 2022-08-18
Implement computation of moments for NodeStat and EdgeStat

### LINKED ISSUE #149: How to compute the second moment of a NodeStat?
Currently we have `H.nodes.degree.mean()` and `H.nodes.degree.std()`, but we should also have a convenience method to compute the second moment (centered or uncentered) since this is so important in network science.

leotrs:
Closes #149.

I decided to add the convenience function `moment2` since the often-used "second moment of the degree distribution" is actually **NOT** the statistical second moment. The former is uncentered, the latter is centered. So the default value of `center` in `IDView.moment` is set to `True`, which is NOT what the network scientist would expect. So instead of doing `H.nodes.degree.moment(2, False)`, we can just do `H.nodes.degree.moment2()`.

### DIFF
--- a/docs/source/api/stats/xgi.stats.EdgeStat.rst
+++ b/docs/source/api/stats/xgi.stats.EdgeStat.rst
@@ -21,13 +21,15 @@
    .. autosummary::
       :nosignatures:
 
-      ~NodeStat.asdict
-      ~NodeStat.aslist
-      ~NodeStat.asnumpy
-      ~NodeStat.aspandas
-      ~NodeStat.max
-      ~NodeStat.mean
-      ~NodeStat.median
-      ~NodeStat.min
-      ~NodeStat.std
-      ~NodeStat.var
+      ~EdgeStat.asdict
+      ~EdgeStat.aslist
+      ~EdgeStat.asnumpy
+      ~EdgeStat.aspandas
+      ~EdgeStat.max
+      ~EdgeStat.mean
+      ~EdgeStat.median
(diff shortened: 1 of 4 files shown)


nwlandry:
Is the computation of `moment2` different than `var`? Also, I think that the moment method should be uncentered by default. Other than that, I think this looks great!!

leotrs:
I think the moment method should be centered by default. This is the standard meaning of "method" in statistics. This is precisely the main reason for having `moment2` default to uncentered!

Perhaps we can ask @maximelucas to break the tie :)

maximelucas:
For network scientists, I'm not sure whether moments are expected to be centered or not. 
For statistics, wikipedia says a "moment" is by default "raw" (uncentered), but can also be "central" or "standardised". 
`scipy.stats`'s `moment` is actual the central moment, as explicitly said in the docs.

Just because we need to discuss this, I think we should choose a solution that is explicit. 
I would do the following:
- `moment()`: not centered
- `central_moment()"`: centered

Or, have a single method `moment()` that is uncentered by default. And have the docs explicitly explain all this.  
Oops, not exactly choosing between your two options :) 
What do you think?

leotrs:
LMAO I must have read the wikipedia page wrong. Ok let's do uncentered/raw by default. This also means that `moment2` is useless so I'll remove it. Thanks @maximelucas @nwlandry

leotrs:
Closing this in favor of #155

PR #153, opened 2022-08-27
Add unit tests

nwlandry:
This adds some unit tests for the convert module, the function module, and the classic generators module to increase overall test coverage and fixes for minor bugs encountered while writing tests.  This also adds documentation to Read The Docs for the drawing module.


xgi/classes/function.py
```diff
@@ -424,13 +425,13 @@ def set_node_attributes(H, values, name=None):
...
     if name is not None:  # `values` must not be a dict of dict
-        try:  # `values` is a dict
+        if isinstance(values, dict):  # `values` is a dict
             for n, v in values.items():
                 try:
                     H._node_attr[n][name] = v
                 except KeyError:
-                    pass
-        except AttributeError:  # `values` is a constant
+                    warn(f"Node {n} does not exist!")
```
leotrs, line 433: Question: I'm confused as to when this ever gets executed - I thought the IDDict should catch all instances of missing node IDs?

xgi/convert.py
```diff
@@ -87,6 +87,9 @@ def convert_to_hypergraph(data, create_using=None):
     ):
         from_incidence_matrix(data, create_using)
 
+    else:
+        raise XGIError("Not able to convert to hypergraph!")
```
leotrs, line 91: Suggestion: if possible, please add a reason why XGI failed rather than just saying that it failed.

xgi/convert.py
```diff
@@ -161,6 +164,8 @@ def convert_to_simplicial_complex(data, create_using=None):
         raise XGIError(
             "Not implemented: construction of a SimplicialComplex from incidence matrix"
         )
+    else:
+        raise XGIError("Not able to convert to simplicial complex!")
```
leotrs, line 168: See prev comment.

leotrs (changes requested):
Thanks for this, adding tests is hugely important. Just a couple of comments.

xgi/convert.py
nwlandry: Done.

xgi/classes/function.py
nwlandry: Good point. Fixed.

xgi/convert.py
nwlandry: fixed.

leotrs (approved):

PR #154, opened 2022-08-27
Fix bug in `remove_node_from_edge()`

nwlandry:
Fixes the error in this example:
```
H = xgi.Hypergraph()
H.add_edge([1, 2])
H.edges.members()
# [[1, 2]]
H.remove_node_from_edge(0, 2)
```
raises the error
`XGIError: Node 2 not in edge 0`


leotrs:
Thanks. This looks like the product of a bad merge conflict fix.

PR #155, opened 2022-08-29
Fix 149

### LINKED ISSUE #149: How to compute the second moment of a NodeStat?

leotrs:
Second try of #151. See that thread for discussion.

Closes #149 .


nwlandry (approved):
Looks great!

PR #156, opened 2022-08-29
Implement weak and strong node removal.

### LINKED ISSUE #76: Implement weak removal of edges
A weak removal of a node from a hypergraph means removing the node from each edge it belongs to and then removing the node itself from the node set. A strong removal means weak removal AND deleting all edges that contain that node. Currently, `Hypergraph.remove_node` implements strong removal only.

We should think whether or not this makes sense in the case of simp. comps.

leotrs:
See #76.

Closes #76.


nwlandry (approved):
Looks good!

PR #157, opened 2022-08-29
Implement categorical stats

### LINKED ISSUE #141: Differentiate between numeric and categorical statistics
Currently, the stats package treats all NodeStat and EdgeStat objects the same. But some of them, e.g. node attributes, tend to be string-valued and thus categorical. In particular, the `.mean(), .max(), .std()` and other similar methods shouldn't be defined for categorical statistics. For example:

```python3
H = xgi.Hypergraph([[1, 2, 3], [2, 3, 4, 5], [3, 4, 5]])
H.add_nodes_from(
        [
            (1, {"color": "red", "name": "horse"}),
            (2, {"color": "blue", "name": "pony"}),
            (3, {"color": "yellow", "name": "zebra"}),
            (4, {"color": "red", "name": "orangutan", "age": 20}),
            (5, {"color": "blue", "name": "fish", "age": 2}),
        ]
    )

# Attributes can be accessed via the stats package
H.nodes.attrs('color').asdict()
# -> {1: 'red', 2: 'blue', 3: 'yellow', 4: 'red', 5: 'blue'}

# .mean() is defined, just like with any other NodeStat...
H.nodes.attrs('color').mean
# -> <bound method IDStat.mean of NodeStat('attrs', args=('color',))>

# but of course it raises an exception.
H.nodes.attrs('color').mean()
# -> TypeError: cannot perform reduce with flexible type
```

leotrs:
See #141.

1. Implements categorical and numerical stats. Numerical stats are exactly the same as the stats we already had. Categorical stats are the same EXCEPT that some methods now raise the exception `StatKindError` when making some computations. For example, trying to compute the min, max, std, var, median, or mean of a categorical stat raises this error.

2. Impelements `IDStat.mode` which can be computed for either numerical or categorical stats.

3. The default is for a stat to be numerical.
 
4. **For developers**: if you want to add a new categorical stat, just add the corresponding function to either the `nodestats` or `edgestats` module, and then set the attribute `kind` of that function to `statkinds.CATEGORICAL`. There is an example [here](https://github.com/ComplexGroupInteractions/xgi/pull/157/files#diff-b3629fd2bee61fbc225b06c9940f4f6507a061fedcbe8e4561ddac90892b04c2R116), which sets node attributes to categorical. If you want to add a new numerical stat, there is nothing new for you to do.

5. **For users**: when adding a new stat using the decorators `nodestat_func` or `edgestat_func`, use the new keyword argument `categorical` to `True`.

6. Adds tests.

Closes #141 .

### DIFF
--- a/tests/stats/test_nodestats.py
+++ b/tests/stats/test_nodestats.py
@@ -3,7 +3,7 @@ import pandas as pd
 import pytest
 
 import xgi
-from xgi.exception import IDNotFound, XGIError
+from xgi.exception import IDNotFound, XGIError, StatKindError
 
 
 def test_filterby_wrong_stat():
--- a/tests/stats/test_nodestats.py
+++ b/tests/stats/test_nodestats.py
@@ -202,6 +202,25 @@ def test_attrs(hyperwithattrs, attr1, attr2, attr3, attr4, attr5):
     assert filtered.asdict() == {3: attr3}
 
 
+def test_attrs_categorical(hyperwithattrs):
+    H = hyperwithattrs
+    assert H.nodes.attrs("color").mode() == "blue"
+    assert H.nodes.attrs("color").mode("all") == ["blue", "red"]
+
+    with pytest.raises(StatKindError):
+        H.nodes.attrs("color").mean()
+    with pytest.raises(StatKindError):
+        H.nodes.attrs("name").min()
+    with pytest.raises(StatKindError):
+        H.nodes.attrs("name").max()
+    with pytest.raises(StatKindError):
+        H.nodes.attrs("name").std()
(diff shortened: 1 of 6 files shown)


leotrs:
One standing issue I see with this is that **every attribute** is now considered categorical. I think this is fine for now, since if someone has a numerical attribute, they'd be better off defining a separate nodestat instead of storing it as an attribute. However, in the future we may want to give the user the freedom to determine whether **each attribute** is numerical or categorical.

leotrs:
Closing this as it does not address the issues addressed in today's meeting. See [here](https://github.com/ComplexGroupInteractions/xgi/issues/141#issuecomment-1279073694).

PR #159, opened 2022-08-30
Add function for Kuramoto model synchronization

saad1282:
Added a dynamics module and created a Kuramoto model synchronization function.


tests/dynamics/test_synchronization.py
```diff
@@ -0,0 +1,12 @@
+import pytest
+
+import xgi
+from xgi.exception import XGIError
+import numpy as np
+
+@pytest.mark.slow
```
leotrs, line 7: Question: How slow is slow? Would this test be much too different if we used a hypergraph with say 50 nodes instead of 100?

tests/dynamics/test_synchronization.py
```diff
@@ -0,0 +1,12 @@
...
+def test_compute_kuramoto_order_parameter():
+    H1 = xgi.random_hypergraph(100, [0.05, 0.001], seed=0)
+    r = xgi.compute_kuramoto_order_parameter(H1, 1, 1, np.ones(100), 10, 0.002)
+   
+    assert len(r) == 10
```
leotrs, line 12: Question: this only tests the length of the output, which should always match the fifth argument in the call in the previous line, no? If so, I suggest hard-coding the actual numeric output and testing against that. There should be no problem since we are fixing the seed two lines above anyway.

xgi/dynamics/synchronization.py
```diff
@@ -0,0 +1,95 @@
...
+    "Synchronization of phase oscillators on complex hypergraphs"
+    by Sabina Adhikari, Juan G. Restrepo and Per Sebastian Skardal
+    https://doi.org/10.48550/arXiv.2208.00909
+
+    Examples
+    --------
+    >>> import numpy as np
+    >>> import xgi
+    >>> n = 100
+    >>> H = xgi.random_hypergraph(n, [0.05, 0.001], seed=None)
```
leotrs, line 45: Same as above, would this example be functionally the same if we used a hypergraph with 50 nodes instead of 100? This will make it considerably faster.

xgi/dynamics/synchronization.py
```diff
@@ -0,0 +1,95 @@
...
+    >>> w = 2*np.ones(n)
+    >>> theta = np.linspace(0, 2*np.pi, n)
+    >>> H.add_nodes_from(range(n))
```
leotrs, line 48: Why are more nodes being added?

xgi/dynamics/synchronization.py
```diff
@@ -0,0 +1,95 @@
...
+    https://doi.org/10.48550/arXiv.2208.00909
+
+    Examples
+    --------
+    >>> import numpy as np
+    >>> import xgi
+    >>> n = 100
+    >>> H = xgi.random_hypergraph(n, [0.05, 0.001], seed=None)
+    >>> w = 2*np.ones(n)
+    >>> theta = np.linspace(0, 2*np.pi, n)
```
leotrs, line 47: When is theta being used?

xgi/dynamics/synchronization.py
```diff
@@ -0,0 +1,95 @@
...
+    >>> H.add_nodes_from(range(n))
+    >>> k2 = 2
+    >>> k3 = 3
+    >>> R = compute_kuramoto_order_parameter(H, k2, k3, w)
```
leotrs, line 51: I would rather do the following as it's shorter and in this case exactly the same.
```suggestion
    >>> R = compute_kuramoto_order_parameter(H, k2=2, k3=3, w)
```

leotrs (changes requested):
Thank you for your contribution @saad1282 ! I left a few minor comments that hopefully won't be too time-consuming to address.

The algorithm it self looks good to me. It's clean and easy to read. However I'm a bit concerned about performance. I assume the operations inside the main loop are hard to vectorize since they iterate over the links/triangles, is that correct? Or are you aware of some way of writing down these operations using matrices? If we can get rid of those inner loops and write the entire computation in a single/few matrix computations then we'd gain a lot in performance.

nwlandry:
> Thank you for your contribution @saad1282 ! I left a few minor comments that hopefully won't be too time-consuming to address.
> 
> The algorithm it self looks good to me. It's clean and easy to read. However I'm a bit concerned about performance. I assume the operations inside the main loop are hard to vectorize since they iterate over the links/triangles, is that correct? Or are you aware of some way of writing down these operations using matrices? If we can get rid of those inner loops and write the entire computation in a single/few matrix computations then we'd gain a lot in performance.

Hi @leotrs. @saad1282 and I worked on implementing a numpy version and I haven't tested line-by-line but I suspect that forming the adjacency matrix and tensor is relatively expensive. Suggestions welcome!

```
def compute_kuramoto_order_parameter_fast(H, k2, k3, w, timesteps=10000, dt=0.002):
    """This function calculates the order parameter for the Kuramoto model on hypergraphs.

    This solves the Kuramoto model ODE on hypergraphs with edges of sizes 2 and 3
    using the Euler Method. It returns an order parameter which is a measure of synchrony.

    Parameters
    ----------
    H : Hypergraph object
        The hypergraph on which you run the Kuramoto model
    k2 : float
        The coupling strength for links
    k3 : float
        The coupling strength for triangles
    w : numpy array of real values
        The natural frequency of the nodes.
    timesteps : int greater than 1, default: 10000
        The number of timesteps for Euler Method.
    dt : float greater than 0, default: 0.002
        The size of timesteps for Euler Method.

    Returns
    -------
    r_time : numpy array of floats
        timeseries for Kuramoto model order parameter

    References
    ----------
    "Synchronization of phase oscillators on complex hypergraphs"
    by Sabina Adhikari, Juan G. Restrepo and Per Sebastian Skardal
    https://doi.org/10.48550/arXiv.2208.00909

    Examples
    --------
    >>> import numpy as np
    >>> import xgi
    >>> n = 100
    >>> H = xgi.random_hypergraph(n, [0.05, 0.001], seed=None)
    >>> w = 2*np.ones(n)
    >>> theta = np.linspace(0, 2*np.pi, n)
    >>> H.add_nodes_from(range(n))
    >>> k2 = 2
    >>> k3 = 3
    >>> R = compute_kuramoto_order_parameter(H, k2, k3, w)
    """
    H_int = xgi.convert_labels_to_integers(H, "label")
    n = H_int.num_nodes
    links = H_int.edges.filterby("size", 2).members()
    A = np.zeros([n, n])
    for i, j in links:
        A[i, j] += 1
        A[j, i] += 1

    triangles = H_int.edges.filterby("size", 3).members()

    B = np.zeros([n, n, n])

    for t in triangles:
        for i, j, k in product(t, repeat=3):
            B[i, j, k] += 1
    

    theta = np.linspace(0, 2 * np.pi, n)
    r_time = np.zeros(timesteps)

    for t in range(timesteps):

        r2 = np.zeros(n, dtype=complex)

        r1 = A.dot(np.exp(1j*theta))

        r2 = np.exp(-1j*theta).dot(B).dot(np.exp(2j*theta))

        d_theta = (
            w
            + k2 * np.multiply(r1, np.exp(-1j * theta)).imag
            + k3 * np.multiply(r2, np.exp(-1j * theta)).imag
        )
        theta_new = theta + d_theta * dt
        theta = theta_new
        z = np.mean(np.exp(1j * theta))
        r_time[t] = abs(z)

    return r_time
```

saad1282:
> Thank you for your contribution @saad1282 ! I left a few minor comments that hopefully won't be too time-consuming to address.
> 
> The algorithm it self looks good to me. It's clean and easy to read. However I'm a bit concerned about performance. I assume the operations inside the main loop are hard to vectorize since they iterate over the links/triangles, is that correct? Or are you aware of some way of writing down these operations using matrices? If we can get rid of those inner loops and write the entire computation in a single/few matrix computations then we'd gain a lot in performance.

@leotrs This is a good point. This is now issue #167.

tests/dynamics/test_synchronization.py
saad1282: It reduces it down from 8 seconds to 2 seconds for 10000 timesteps. We will either change the size of the hypergraph or reduce the timesteps to make it faster.

tests/dynamics/test_synchronization.py
saad1282: We totally agree that there should be more rigorous tests. We will have to figure out better tests and your suggestion is a good option.

xgi/dynamics/synchronization.py
saad1282: See the comment above.

xgi/dynamics/synchronization.py
saad1282: Good point. We just removed it.

xgi/dynamics/synchronization.py
saad1282: Fixed.

leotrs (approved):

PR #161, opened 2022-08-31
Cleanup method

nwlandry:
* Added a cleanup method that removes artifacts specified by the user: multi-edges, singletons, isolates. It also can convert all labels to consecutive integers.
* Modified the `duplicates()` method to not include the first instance of the node/edge in the list of duplicates, so that when you call `H.remove_edges_from(H.edges.duplicates())` it only removes the redundant edges, not all of the edges. In preparation of only supporting non-loopy edges, the `duplicates()` method has been simplified to not differentiate between duplicate nodes in an edge.

### DIFF
--- a/xgi/classes/hypergraph.py
+++ b/xgi/classes/hypergraph.py
@@ -282,7 +282,7 @@ class Hypergraph:
 
         """
         if node not in self._node:
-            self._node[node] = []
+            self._node[node] = set()
             self._node_attr[node] = self._node_attr_dict_factory()
         self._node_attr[node].update(attr)
 
@@ -316,7 +316,7 @@ class Hypergraph:
                 newdict = attr.copy()
                 newdict.update(ndict)
             if newnode:
-                self._node[n] = []
+                self._node[n] = set()
                 self._node_attr[n] = self._node_attr_dict_factory()
             self._node_attr[n].update(newdict)
 
@@ -415,13 +415,13 @@ class Hypergraph:
             raise XGIError("Cannot add an empty edge")
 
         uid = next(self._edge_uid) if not id else id
-        self._edge[uid] = []
+        self._edge[uid] = set()
         for node in members:
             if node not in self._node:
-                self._node[node] = []
+                self._node[node] = set()
                 self._node_attr[node] = self._node_attr_dict_factory()
-            self._node[node].append(uid)
-            self._edge[uid].append(node)
+            self._node[node].add(uid)
+            self._edge[uid].add(node)
 
         self._edge_attr[uid] = self._hyperedge_attr_dict_factory()
         self._edge_attr[uid].update(attr)
@@ -532,9 +532,9 @@ class Hypergraph:
                     raise XGIError("Invalid ebunch format") from e
                 for n in members:
                     if n not in self._node:
-                        self._node[n] = []
+                        self._node[n] = set()
                         self._node_attr[n] = self._node_attr_dict_factory()
-                    self._node[n].append(uid)
+                    self._node[n].add(uid)
                 self._edge_attr[uid] = self._hyperedge_attr_dict_factory()
             return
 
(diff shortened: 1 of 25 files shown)


xgi/classes/hypergraph.py
```diff
@@ -964,3 +964,47 @@ def dual(self):
         dual._hypergraph = deepcopy(self._hypergraph)
 
         return dual
+
+    def cleanup(
+        self,
+        remove_isolates=True,
+        remove_singletons=True,
+        remove_multiedges=True,
```
leotrs, line 972:
```suggestion
        isolates=True,
        singletons=True,
        multiedges=True,
```

xgi/classes/hypergraph.py
```diff
@@ -964,3 +964,47 @@ def dual(self):
...
+
+        Parameters
+        ----------
+        remove_isolates : bool, optional
+            Whether to remove isolated nodes, by default True
+        remove_singletons : bool, optional
+            Whether to remove singleton edges, by default True
+        remove_multiedges : bool, optional
+            Whether to remove duplicate edges, by default True
+        reindex : bool, optional
```
leotrs, line 985: the name of the function contains the word `label` so to keep it consistent, this parameter should be called `relabel` too

xgi/classes/reportviews.py
```diff
@@ -375,7 +375,8 @@ def duplicates(self):
         Notes
         -----
         The IDs returned are in an arbitrary order, that is duplicates are not
-        guaranteed to be consecutive.
+        guaranteed to be consecutive. For IDs with the same bipartite neighbors,
+        only the first ID added is not a duplicate.
```
leotrs, line 379:
```suggestion
        guaranteed to be consecutive.
        If two IDs have the same bipartite neighbors, only one of them is returned.
```

leotrs: I wonder if we should add an option to return all of them, not only those ones that are actually repeated (and ignoring the first one). In a way, `duplicates` is similar to a `groupby` operation (it's grouping IDs that have the same bipartite neighbors) and `groupby` operations always return all of the elements involved, they never ignore the first one.

leotrs: Or am I overthinking things again? :sweat_smile:

xgi/convert.py
```diff
@@ -88,7 +88,9 @@ def convert_to_hypergraph(data, create_using=None):
         from_incidence_matrix(data, create_using)
 
     else:
-        raise XGIError("Not able to convert to a hypergraph because the input data was not a supported type!")
+        raise XGIError(
+            "Not able to convert to a hypergraph because the input data was not a supported type!"
```
leotrs, line 92:
```suggestion
            "Input data has unsupported type."
```

xgi/classes/hypergraph.py
```diff
@@ -964,3 +964,47 @@ def dual(self):
...
+            ee = temp.edges
+
+            self.remove_nodes_from(list(self._node))
+            self.add_nodes_from((n, deepcopy(attr)) for n, attr in nn.items())
+            
+            self.remove_edges_from(list(self._edge))
+            self.add_edges_from(
+            (e, id, deepcopy(temp.edges[id]))
+            for id, e in ee.members(dtype=dict).items()
+        )
```
leotrs, line 1009: Make sure to run black on this.

leotrs (changes requested):
Thanks for this! BTW, does this depend on #162?

Also, should we add a parameter `in_place`? If `in_place` is False, then a new hypergraph is returned and the original one is left intact. This might simplify some of the code inside the `if reindex:` branch.

xgi/classes/hypergraph.py
nwlandry: Done.

xgi/classes/hypergraph.py
nwlandry: Done.

xgi/classes/hypergraph.py
nwlandry: Done.

xgi/classes/reportviews.py
nwlandry: Added as issue [#163](https://github.com/ComplexGroupInteractions/xgi/issues/163).

leotrs:
Pro tip: next time try rebasing your PR rather than merging main into it bc it makes it hard to review. The current diff shows both changes from this PR as well as the other one. If rebasing does not work/is scary, you can also just close it and start a new PR.

leotrs (approved):

Issue #158, opened 2022-08-30
Coerce all edges to sets


leotrs:
This will be done in order to forbid the appearance of repeated nodes in an edge (such an edge is called "loopy" hyperedge). @nwlandry and I discussed this and realized that XGI incurs in non-trivial performance costs due to supporting loopy hypergedges, however the number of applications where loopy hyperedges is necessary or useful is very few. So the tradeoff is just not there. We have decided to NOT support loopy hyperedges from now on. In essence, this means that the members of each edge (and also the memberships of each node) will be stored as sets. So even if the user passes in a list or some other collection, it will be coerced into a set.

PR #162, opened 2022-08-31
Convert to sets

nwlandry:
Converted all instances of edges to sets from lists in response to #158. Changed unit tests to reflect this.

### DIFF
--- a/tests/classes/test_hypergraph.py
+++ b/tests/classes/test_hypergraph.py
@@ -26,11 +26,11 @@ def test_constructor(edgelist5, dict5, incidence5, dataframe5):
         == list(H_hg.edges)
     )
     assert (
-        list(H_list.edges.members(0))
-        == list(H_dict.edges.members(0))
-        == list(H_mat.edges.members(0))
-        == list(H_df.edges.members(0))
-        == list(H_hg.edges.members(0))
+        H_list.edges.members(0)
+        == H_dict.edges.members(0)
+        == H_mat.edges.members(0)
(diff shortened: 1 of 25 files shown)


tests/test_convert.py
```diff
@@ -42,12 +42,12 @@ def test_convert_to_graph(edgelist2, edgelist5):
 
 def test_to_hyperedge_list(edgelist1):
```
leotrs, line 43: We should change the definitions of `edgelistX` in `conftest.py` to be sets. Then we won't need all those list comprehensions later on.

leotrs (changes requested):
Thanks so much for taking the lead on this one! Does this depend on #161 ?

tests/classes/test_hypergraph.py
```diff
@@ -187,66 +187,66 @@ def test_add_node_to_edge():
...
     assert H.edges.members(dtype=dict) == {
-        "fruits": ["apple", "banana", "pear"],
-        "veggies": ["lettuce"],
+        "fruits": {"apple", "banana", "pear"},
+        "veggies": {"lettuce"},
     }
 
 
 def test_add_edges_from_iterable_of_members():
     edges = [[0, 1], [1, 2], [2, 3, 4]]
```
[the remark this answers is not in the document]
nwlandry: Good call.

tests/test_convert.py
nwlandry: done.

leotrs (approved):

PR #165, opened 2022-09-07
Generalized node-edge centrality

### LINKED ISSUE #132: Make `node_edge_centrality` more general
In "Node and edge nonlinear eigenvector centrality for hypergraphs" by Tudisco and Higham, their formalism is more general than the implementation in xgi because
1. It allows general functions for both nodes and edges
2. Nodes and edges may be weighted
3. One can use norms besides the L2 norm.

nwlandry:
Added lambda function default arguments for $f$, $g$, $\varphi$, $\psi$ as defined by Tudisco and Higham. Default behavior is identical as before. Fixes #132.

### DIFF
--- a/xgi/algorithms/centrality.py
+++ b/xgi/algorithms/centrality.py
@@ -157,13 +157,33 @@ def apply(H, x, g=lambda v, e: np.sum(v[list(e)])):
     return new_x
 
 
-def node_edge_centrality(H, max_iter=100, tol=1e-6):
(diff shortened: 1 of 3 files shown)


leotrs (approved):
Documentation could be nicer, but I'm approving because we have bigger fish to fry.

Also, can you please open an issue recording the problem of having a single algorithm in three places (algorithms, nodestats, edgestats).

PR #169, opened 2022-09-13
Add asv for benchmarking

nwlandry:
Added a benchmarking suite for the core hypergraph data structure and a README explaining how to use asv.


nwlandry:
Made a mistake renaming the branch...

PR #173, opened 2022-09-21
Speedup drawing

nwlandry:
Attempted to make `xgi.draw()` a bit faster.

### DIFF
--- a/tutorials/Tutorial 5 - Plotting.ipynb
+++ b/tutorials/Tutorial 5 - Plotting.ipynb
@@ -272,8 +272,7 @@
     "    pos,\n",
     "    node_size=H.nodes.degree,\n",
     "    node_lw=H.nodes.degree,\n",
-    "    node_fc=H.nodes.cec_centrality,\n",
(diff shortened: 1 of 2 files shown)


leotrs (approved):
Out of curiosity, do you know how much faster this is?

nwlandry:
> Out of curiosity, do you know how much faster this is?

Tested on the email-enron dataset (after cleanup with default argument) for the edges and the size EdgeStat:
_color_arg_to_dict (old): 3.08 s ± 290 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
_color_arg_to_dict (new): 119 ms ± 9.86 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
_scalar_arg_to_dict (old): 2.74 s ± 294 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
_scalar_arg_to_dict (new): 3.19 ms ± 74.5 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

leotrs:
So a 1000x speedup???? Jeez.

Issue #183, opened 2022-10-03
add function to get number of edges of size d


maximelucas:
We don't an easy short way of doing this anymore. 

Right now we can do it as `len(H.edges.filterby("size", d))`.

PR #192, opened 2022-10-05
Minor feats

maximelucas:
Implements #183 and #184.


leotrs:
@nwlandry recently we discussed making `num_edges` a method instead of a property - do you remember what was the context of that? If we ever get to that, then we might as well just add the `order` parameter to `num_edges` instead of adding a second method like on this PR.

nwlandry:
I believe that this was in the context of simplicial complexes where there are different notions of number of simplices (i.e., is it all the subfaces as well, or just maximal simplices?)

PR #200, opened 2022-10-24
Partial clique complex

maximelucas:
Added an option to fill only some cliques with probabilities `ps` to `xgi.flag_complex()`.

Started this branch a long time ago so we'll need to merge main, but easier to check the new code before that.


leotrs:
> Started this branch a long time ago so we'll need to merge main, but easier to check the new code before that.

An alternative is to rebase your branch over main, which gives a much cleaner commit :) Don't do it now tho since it usually does not play well with open PRs..

xgi/generators/classic.py
```diff
@@ -191,16 +201,26 @@ def flag_complex(g, max_order=2, seed=None):
...
+    if not ps: # promote all cliques
+        S.add_simplices_from(max_cliques, max_order=max_order)
+    else: # promote cliques with a given probability
+        for i, p in enumerate(ps[:max_order - 1]):
+            d = i + 2 # simplex order
+            cliques_d = [x for x in max_cliques if len(x) == d + 1]
+            cliques_d_to_add = []
+            for el in cliques_d:
+                if seed.random() <= p:
+                    cliques_d_to_add.append(el)
```
leotrs, line 228:
```suggestion
            cliques_d_to_add = [el for el in cliques_d if if seed.random() <= p]
```

maximelucas:
We'll have to try one day hehe. I've just resolved the conflicts now actually. You're too fast, had not seen your message :)

xgi/generators/classic.py
```diff
@@ -196,17 +206,27 @@ def flag_complex(g, max_order=2, seed=None):
...
     S = SimplicialComplex()
     S.add_nodes_from(nodes)
-    S.add_simplices_from(max_cliques, max_order=max_order)
+    S.add_simplices_from(edges)
+    if not ps: # promote all cliques
+        S.add_simplices_from(max_cliques, max_order=max_order)
+    else: # promote cliques with a given probability
+        for i, p in enumerate(ps[:max_order - 1]):
+            d = i + 2 # simplex order
+            cliques_d = [x for x in max_cliques if len(x) == d + 1]
```
leotrs, line 224: Here, you are iterating over `max_cliques` once per iteration of the loop on line 222. A better way of doing this would be to define a dictionary before the loop starts:
```python3
cliques_d = defaultdict(list)
for x in max_cliques:
    cliques_d[len(x)].append(x)
```
and then just use `cliques_d[d+1]` in each of the iterations of the loop. This should in general be faster IF we expect `max_cliques` to be large...

leotrs (changes requested):

xgi/generators/classic.py
maximelucas: Nice thanks!

xgi/generators/classic.py
```diff
@@ -196,17 +207,26 @@ def flag_complex(g, max_order=2, seed=None):
...
+    else: # promote cliques with a given probability
+        cliques_d = defaultdict(list)
+        for x in max_cliques:
+            cliques_d[len(x)].append(x)
+        for i, p in enumerate(ps[:max_order - 1]):
+            d = i + 2 # simplex order
+            cliques_d_to_add = [el for el in cliques_d[d+1] if seed.random() <= p]
+            S.add_simplices_from(cliques_d_to_add, max_order=max_order)
 
     return S
```
leotrs, line 231: This is just a tiny nitpick, feel free to ignore (I'm already accepting). I prefer this because in this way, the code that actually does most of the work ends up on a lower level of indentation so it's easier to read:

```suggestion
    if not ps: # promote all cliques
        S.add_simplices_from(max_cliques, max_order=max_order)
        return S
        
    # promote cliques with a given probability
    cliques_d = defaultdict(list)
    for x in max_cliques:
        cliques_d[len(x)].append(x)
            
    for i, p in enumerate(ps[:max_order - 1]):
        d = i + 2 # simplex order
        cliques_d_to_add = [el for el in cliques_d[d+1] if seed.random() <= p]
        S.add_simplices_from(cliques_d_to_add, max_order=max_order)

    return S
```

leotrs (approved):

PR #202, opened 2022-10-28
Docs spiff up

### LINKED ISSUE #185: Doc for H.degree() not pointing to H.nodes.degree()
As a result, `H.degree(order=2)` works, but the argument `order` is not in the docs.

leotrs:
* Added a short blurb on the landing page. This was a feedback I received from someone who looked at the page and told me they have no idea what the library was even for!
* Added a reference that seemed relevant.
* Updated python version to 3.11. 
* Fixes #185.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,9 +74,9 @@ Leo Torres <leo@leotrs.com>
 
 Iacopo Iacopini <iacopiniiacopo@gmail.com>
 
-Maxime Lucas <maxime.lucas@isi.it>
+Maxime Lucas <maxime.lucas@centai.eu>
 
-Giovanni Petri <giovanni.petri@isi.it>
+Giovanni Petri <giovanni.petri@centai.eu>
 
 Alice Patania <apatania@uvm.edu>
 
--- a/docs/source/api/classes/xgi.classes.function.rst
+++ b/docs/source/api/classes/xgi.classes.function.rst
@@ -21,6 +21,8 @@
    .. autofunction:: is_possible_order
    .. autofunction:: is_uniform
    .. autofunction:: max_edge_order
+   .. autofunction:: maximal_simplices
+   .. autofunction:: num_edges_order
    .. autofunction:: set_edge_attributes
    .. autofunction:: set_node_attributes
    .. autofunction:: unique_edge_sizes
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -48,6 +48,21 @@ The Comple\ **X** **G**\ roup **I**\ nteractions `(XGI) <https://github.com/Comp
 library provides data structures and algorithms for modeling and analyzing complex systems
 with group (higher-order) interactions.
 
+Many datasets can be represented as graphs, where pairs of entities (or nodes) are
+related via links (or edges).  Examples are road networks, energy grids, social
+networks, neural networks, etc.  However, in many other datasets, more than two entities
+can be related at a time.  For example, many scientists (entities) can collaborate on a
+scientific article together (links), and multiple email accounts (entities) can all
+participate on the same email thread (links).  In this latter case, graphs no longer
+present a viable alternative to represent such datasets.  It is for this kind of
+datasets, where the interactions are given among groups of more than two entities (also
+called higher-order interactions), that XGI was designed for.
+
+XGI is implemented in pure Python and is designed to seamlessly interoperate with the
+rest of the Python scientific stack (numpy, scipy, pandas, matplotlib, etc).  XGI is
+designed and developed by network scientists with the needs of network scientists in
+mind.
+
 - Repository: https://github.com/ComplexGroupInteractions/xgi
 - PyPI: https://pypi.org/project/xgi/
 - Documentation: https://xgi.readthedocs.io/
@@ -74,7 +89,7 @@ If that command does not work, you may try the following instead
 
    pip install -e .\[all\]
 
-XGI was developed and tested for Python 3.7-3.10 on Mac OS, Windows, and Ubuntu.
+XGI was developed and tested for Python 3.7-3.11 on Mac OS, Windows, and Ubuntu.
 
 
 Academic References
@@ -91,6 +106,10 @@ Academic References
 
 * `What are higher-order networks? <https://arxiv.org/abs/2104.11329>`_, Christian Bick,
   Elizabeth Gross, Heather A. Harrington, Michael T. Schaub.
+
+* `From networks to optimal higher-order models of complex systems
+  <https://www.nature.com/articles/s41567-019-0459-y>`_, Renaud Lambiotte, Martin
+  Rosvall, and Ingo Scholtes.
 
 
 Contributing
--- a/xgi/classes/hypergraph.py
+++ b/xgi/classes/hypergraph.py
@@ -74,7 +74,7 @@ class Hypergraph:
     Unique IDs are assigned to each node and edge internally and are used to refer to
     them throughout.
 
-    The `attr` keyword arguments are added as hypergraph attributes. To add node or ede
+    The `attr` keyword arguments are added as hypergraph attributes. To add node or edge
     attributes see :meth:`add_node` and :meth:`add_edge`.
 
     In addition to the methods listed in this page, other methods defined in the `stats`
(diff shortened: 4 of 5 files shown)


leotrs:
@maximelucas This PR addresses #185. Could you please confirm that the proposed solution actually solves the issue to your satisfaction?

The old behavior was:

```python3
>>> H.degree?
Signature: H.degree(node=None, *args, **kwargs)
Docstring: <no docstring>
File:      ~/code/xgi/xgi/classes/hypergraph.py
Type:      function
```

The new behavior is:
```python3
>>> H.degree?
Signature: H.degree(node=None, *args, **kwargs)
Docstring:
Equivalent to H.nodes.degree.asdict(). For accepted *args and
**kwargs, see documentation of H.nodes.degree.
File:      ~/code/xgi/xgi/classes/hypergraph.py
Type:      function

>>> H.size?
Signature: H.size(node=None, *args, **kwargs)
Docstring:
Equivalent to H.edges.size.asdict(). For accepted *args and
**kwargs, see documentation of H.edges.size.
File:      ~/code/xgi/xgi/classes/hypergraph.py
Type:      function
```

nwlandry (approved):
Looks great, Leo!!

maximelucas:
I like it, thanks Leo. I'm thinking now we might need to go further. 
Because if a new user starts looking `H.nodes.degree` they won't find anything in the docs: 
- if they look for it in classes, `Hypergraph.py` doesn't say anything about stats and has no pointers: https://xgi.readthedocs.io/en/latest/api/classes/xgi.classes.hypergraph.Hypergraph.html#xgi.classes.hypergraph.Hypergraph.edges
- if they know to look in the stats packages (how would they?), they might end up here: https://xgi.readthedocs.io/en/latest/api/stats.html, which clearly contains more info. But the new user has to make a conceptual jump by understanding that information about `H.nodes.degree` is actually described in `xgi.stats.nodestats.degree(H)`, which looks pretty different. And there is no example to make that particular link explicit (there are somewhere else above on the page).

I'm not sure how yet, but I'm realising we might need to make the link between these things much more explicit for new users. So far, the best "explanation" of how stats work and what they are is probably the tutorial.

maximelucas:
Okay I realise now that the user can get some info from 

```python
>>> H.nodes.degree?
Signature:   H.nodes.degree(*args, **kwargs)
Type:        NodeStat
String form: NodeStat('degree')
File:        ~/Dropbox (ISI Foundation)/WORK/SCIENCE/xgi/xgi/stats/__init__.py
Docstring:  
An arbitrary node-quantity mapping.

`NodeStat` objects represent a mapping that assigns a value to each node in a
network.  For more details, see the `tutorial
<https://github.com/ComplexGroupInteractions/xgi/blob/main/tutorials/Tutorial%206%20-%20Statistics.ipynb>`_.
```

The docstring describes NodeStat (rather than the degree stat) and points to the tutorial. 
Maybe this is already enough for this PR. I do feel like we need to make this better in the docs too, even if with another PR.

leotrs:
Oh that's a good point. The docstring of `xgi.stats.nodestats.degree` is never really being exposed to the user. The user is not even meant to know that `degree` lives in `xgi.stats.nodestats`. 

We could try to merge the two docstrings somehow but I'm not sure what's the best/most efficient way of doing this. In particular, a big problem is that `H.nodes(node_subset).degree(*args, **kwargs).asdict()` is equivalent to `xgi.stats.nodestats.degree(H, node_subset, *args, **kwargs)`. So the docstring for the latter contains documentation for parameters `H` and `node_subset` which are not part of `H.nodes.degree`. So we can't just copy-paste the docstrings. We'll have to change something more fundamental I think (or write docstrings twice...).

maximelucas:
How is the equivalence between `H.nodes().degree().asdict()` and `xgi.stats.nodestats.degree(H)` implemented in the code?

I think: 
- the docs of `H.nodes` and `H.edges` should mention the stats and point to the relevant documentation (maybe the stats package?). 
- the stats package documentation page should explicitly mention the equivalence.
- the docs of `H.degree` (and other stats) should at least point to the docs of `H.nodes.degree` 
- but then the docs of `H.nodes.degree` should be explicit, mention the equivalence with the stats package syntax, and be findable also online.
If there is a way to solve the last point without copy-pasting, great. If not, I suggest we copy-paste (at least for now).

leotrs:
> How is the equivalence between `H.nodes().degree().asdict()` and `xgi.stats.nodestats.degree(H)` implemented in the code?

The short answer is that each NodeStat object is created with a reference to one of the functions in `xgi.stats.nodestats`. For example, the `H.nodes.degree` object contains a reference to `xgi.stats.nodestats.degree`.

The long answer is, well, long. When a user executes `H.nodes.degree` for the first time a number of things happen:
1. First, `H.nodes` tries to search for a `degree` method, but doesn't find one. So `H.nodes.__getattr__` is executed.
2. This just calls `H.nodes._dispatcher.__getattr__`, which actually searches for a function called `degree` inside the `xgi.stats.nodestats` module.
3. Once the `degree` function is found, a `NodeStat` object is built that wraps around it (this is in line 83 of `xgi.stats.__init__.py`. In particular, this `NodeStat` object stores the`degree` function as `self._func`.
4. The dispatcher object `H.nodes._dispatcher` will then **remember** this `NodeStat` object, so next time the user accesses `H.degree`, the same object is returned and the search doesn't happen again
5. So from now on, typing `H.nodes.degree` just returns a `NodeStat` object.
6. Doing `H.nodes.degree.asdict()` eventually accesses `H.nodes.degree._val`, which in turn calls `self.func(self.net, self.view.ids, *self.args, **self.kwargs)`, where `self.net` is the same as `H`.
7. The same happens with any other call such as `H.nodes.degree.aslist()` or `H.nodes.degree.asnumpy()`, etc. They all just access `self._val`, which calls `self._func`.

(Note: writing this all down makes me realize that we can simplify this entire pipeline a great deal in at least two different ways, so I'll work on changing that soon!)

> 1. the docs of `H.nodes` and `H.edges` should mention the stats and point to the relevant documentation (maybe the stats package?).

As you point this out, I realize that the docs of `H.nodes` are... terrible:

```python3
```


> 2. the stats package documentation page should explicitly mention the equivalence.
> 3. the docs of `H.degree` (and other stats) should at least point to the docs of `H.nodes.degree` (which it now does)
> 4. but then the docs of `H.nodes.degree` should be explicit, mention the equivalence with the stats package syntax, and be findable also online.
> 
> If there is a way to solve the last point without writing the docstrings twice, great. If not, I suggest we write it twice (at least for now).

maximelucas:
> The long answer is, well, long. When a user executes `H.nodes.degree` for the first time a number of things happen:

Note: your PR #209 proposes to simplify that. 

> As you can see, these latter docs actually DO mention and point to the tutorial! Perhaps we should add another link to the stats package, and make sure that H.nodes gives better docs.

Oh okay that's where it is. But yes, it should also be available from `H.nodes` which is the only one exposed to the user. And add a link to the stats package sounds good too. 

> How about something like this:

Nice, looks much better. Is this hard-coded copy-paste or you found another clever way? If just copy-pasted, why not just adapt the docs as follows: 

```python
H.nodes.degree?
Signature:   H.nodes.degree(*args, **kwargs)
Type:        NodeStat
String form: NodeStat('degree')
File:        ~/code/xgi/xgi/stats/__init__.py
Docstring:  
A node mapping with name "degree".

The degree of a node is the number of edges it belongs to.

Executing `H.nodes.degree(*args, **kwargs).asdict()` is equivalent to
`xgi.stats.nodestats.degree(H, *args, **kwargs)`. The available *args
and **kwargs are listed in the docstring of `xgi.stats.nodestats.degree`.

For more details, see the `tutorial
<https://github.com/ComplexGroupInteractions/xgi/blob/main/tutorials/Tutorial%206%20-%20Statistics.ipynb>`_.

Parameters
----------
bunch : Iterable
    Nodes in `net`.
order : int | None
    If not None (default), only count the edges of the given order.
weight : str | None
    If not None, specifies the name of the edge attribute that determines the weight
    of each edge.

Returns
-------
dict 
```

leotrs:
> Nice, looks much better. Is this hard-coded copy-paste or you found another clever way? If just copy-pasted, why not just adapt the docs as follows:

What I posted is possible by just doing something like `func.__doc__ = some_str_here`. So we can easily append the docstring of one function to that of some other function. Modifying each docstring is much much more difficult tho.

maximelucas:
I did not know that. Great then, I like your solution.

PR #203, opened 2022-10-28
new add method

nwlandry:
Allows users to add two hypergraphs together. This reindexes all edges and keeps them all.

### DIFF
--- a/tests/classes/test_hypergraph.py
+++ b/tests/classes/test_hypergraph.py
@@ -57,16 +57,16 @@ def test_contains(edgelist1):
     assert [1, 2, 3] not in H
 
 
-def test_add(edgelist1, edgelist2, hyperwithattrs):
+def test_lshift(edgelist1, edgelist2, hyperwithattrs):
     H1 = xgi.Hypergraph(edgelist1)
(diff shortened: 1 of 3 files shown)


leotrs:
I haven't read the entire code yet BUT the `+` operator should absolutely at all times represent an operation that is commutative. If it is not commutative, then we shouldn't use the `+` operator.

nwlandry:
😭

leotrs:
OK how about this?

```python3
# Keeps edge ids of H1 and relabels the edge ids of H2
# All edges (after relabeling) keep all their attributes
# The node attributes of H1 take precedence over the node attributes of H2
# The hypergraph attributes of H1 take precedence over the hypergraph attributes of H2
X = H1 << H2
```

Since this is not commutative, I propose we use `<<` since I don't think we'd be using it for any other purpose.

Note that if `H1` and `H2` have no attributes at all, then `H1 << H2` and `H2 << H1` are isomorphic (though the edge ids are different).

nwlandry:
Okay, I did your suggestion with the small difference that H2 attributes take preference, but very happy to change to H1 precedence.

leotrs (approved):
I'm happy.

Issue #180, opened 2022-10-03
add density function


maximelucas:
We could add that with the group size as argument. 
This would go in `xgi/classes/function.py`?

leotrs:
Yes for now I'd rather have it work as `np.density(H)`.

maximelucas:
I've never looked for a definition but that's what I was thinking to from size `d`. 

For the global density, I would've just summed over `d`. `2^n` is not the total number of potential hyperedges, is it?

In HyperNetX they do it as (https://pnnl.github.io/HyperNetX/build/_modules/reports/descriptive_stats.html#dist_stats):
* nrows = number of nodes (rows in the incidence matrix)
* ncols = number of edges (columns in the incidence matrix)
* ncells = number of filled cells in incidence matrix
* density = ncells/(nrows*ncols)

leotrs:
> For the global density, I would've just summed over d. 2^n is not the total number of potential hyperedges, is it?

Yes, it is the total number of possible subsets of a set of size `n`.

leotrs:
> density = ncells/(nrows*ncols)

This is the density of the incidence matrix, interesting. I guess this is more an indicator of how big the hyperedges are rather than an indicator of how many hyperedges there are. For example, this density would equal `1` if the hypergraph has a single hyperedge that contains all nodes.

maximelucas:
Ok yea my bad. Although `2^n` counts the empty set which I assume we don't want. So I'd say `2^n -1`?
(It also counts all singletons edges which I always feel a bit weird about but oh well)
For some reason Wolfram gave me a more complicated-looking answer when I asked to sum of n choose k's.

I agree about the oder density. Also, when having multiple edges, it tells you how much their they intersect in terms of nodes, I think. Maybe we can implement it with another name, `incidence_density` or something.

leotrs:
https://www.wolframalpha.com/input?i=sum+for+k+%3D+0+to+n+of+%28n+choose+k%29

maximelucas:
I did it wrong 👍🏻

leotrs:
> So I'd say 2^n -1?

That sounds like a good idea.

> Also, when having multiple edges, it tells you how much their they intersect in terms of nodes, I think.

I don't think this is true. If you have exactly two edges of size `a` and `b`, on `n` nodes, then this density is always equal to `(a+b)/2n`, regardless of whether the edges are overlapping or not.

> Maybe we can implement it with another name, incidence_density or something.

[...]

PR #204, opened 2022-10-29
Implement density function and add tests

leotrs:
Partially addresses #180.

This PR implements the density function discussed in the above ticket. This PR DOES NOT implement the density with respect to the incidence matrix.

I thought this PR was already getting a little big so I'd rather do a separate one for the incidence matrix.

### DIFF
--- a/tests/classes/test_function.py
+++ b/tests/classes/test_function.py
@@ -381,13 +381,32 @@ def test_maximal_simplices(edgelist5, edgelist8):
         xgi.maximal_simplices(H)
 
 
-def test_density_empty():
+def test_density_no_nodes():
+    H = xgi.Hypergraph()
+    with pytest.raises(XGIError):
+        xgi.density(H)
+    with pytest.raises(XGIError):
+        xgi.density(H, order=3)
+    with pytest.raises(XGIError):
+        xgi.density(H, max_order=7)
+    with pytest.raises(XGIError):
+        xgi.density(H, ignore_singletons=True)
+    with pytest.raises(XGIError):
+        xgi.density(H, order=2, ignore_singletons=True)
+    with pytest.raises(XGIError):
+        xgi.density(H, max_order=2, ignore_singletons=True)
+
+
+def test_density_no_edges():
     tol = 1e-10
     H = xgi.Hypergraph()
+    H.add_node(0)
     assert xgi.density(H) < tol
     assert xgi.density(H, order=3) < tol
     assert xgi.density(H, max_order=7) < tol
     assert xgi.density(H, ignore_singletons=True) < tol
+    assert xgi.density(H, order=2, ignore_singletons=True) < tol
+    assert xgi.density(H, max_order=2, ignore_singletons=True) < tol
 
 
 def test_density_one_node():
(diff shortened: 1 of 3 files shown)


xgi/classes/function.py
```diff
@@ -688,3 +690,82 @@ def convert_labels_to_integers(H, label_attribute="label"):
...
+def density(H, order=None, max_order=None, ignore_singletons=False):
+    r"""Hypergraph density.
+
+    The density of a hypergraph is the number of existing edges divided by the number of
+    possible edges.
+
+    Let `H` have :math:`n` nodes and :math:`m` hyperedges. Then,
+
+    * `density(H) =` :math:`\frac{m}{2^n - 1}`,
+    * `density(H, count_singletons=True) =` :math:`\frac{m}{2^n - 1 - n}`.
```
maximelucas, line 704:
```suggestion
    * `density(H, ignore_singletons=True) =` :math:`\frac{m}{2^n - 1 - n}`.
```

xgi/classes/function.py
```diff
@@ -688,3 +690,82 @@ def convert_labels_to_integers(H, label_attribute="label"):
...
+
+    Here, :math:`2^n` is the total possible number of hyperedges on `H`, to which we
+    subtract :math:`1` because the empty hyperedge is not considered, or :math:`n` when
```
maximelucas, line 707:
```suggestion
    subtract :math:`1` because the empty hyperedge is not considered, and :math:`n` when
```

tests/classes/test_function.py
```diff
@@ -379,3 +379,161 @@ def test_maximal_simplices(edgelist5, edgelist8):
...
+
+    H = xgi.Hypergraph()  # two nodes, one singleton edge
+    H.add_nodes_from([0, 1])
+    H.add_edge([0])
+    assert abs(xgi.density(H) - 1 / 3) < tol
+    assert abs(xgi.density(H, order=0) - 1 / 2) < tol
+    assert xgi.density(H, order=1) < tol
+    assert xgi.density(H, ignore_singletons=True) < tol
+
+    H = xgi.Hypergraph()  # two nodes, the other singleton edge
```
maximelucas, line 425: Is this case necessary? It seems symmetric with the previous one

xgi/classes/function.py
```diff
@@ -688,3 +690,82 @@ def convert_labels_to_integers(H, label_attribute="label"):
...
+        Whether to consider singleton edges.  Ignored if `order` is not None and
+        different from :math:`0`.
+
+    Notes
+    -----
+    If both `order` and `max_order` are not None, `max_order` is ignored.
+
+    """
+    n = H.num_nodes
+    if n < 1 or H.num_edges < 1:
```
maximelucas, line 738: Do we want to return 0 if there are 0 nodes? The denom would be 0 technically.
Could also return None or some error like "Cannot compute the density of empty Hypergraph".

xgi/classes/function.py
```diff
@@ -688,3 +690,82 @@ def convert_labels_to_integers(H, label_attribute="label"):
...
+        return H.edges.filterby("order", val, mode=mode)
+
+    if order is None and max_order is None:
+        numer = H.num_edges
+        denom = 2 ** n - 1
+        if ignore_singletons:
+            numer -= len(order_filter(0, mode="eq"))
+            denom -= n
+
+    if order is None and max_order is not None:
```
maximelucas, line 751: better as elif?

xgi/classes/function.py
```diff
@@ -688,3 +690,82 @@ def convert_labels_to_integers(H, label_attribute="label"):
...
+        if max_order >= n:
+            raise ValueError("max_order must be smaller than the number of nodes")
+        numer = len(order_filter(max_order, "leq"))
+        denom = sum(comb(n, _ord + 1, exact=True) for _ord in range(max_order + 1))
+        if ignore_singletons:
+            numer -= len(order_filter(0, mode="eq"))
+            denom -= n
+
+    if order is not None:  # ignore max_order
```
maximelucas, line 760: elif?

maximelucas:
Great Leo, thanks. Just a few details

tests/classes/test_function.py
leotrs: I guess not, deleted!

xgi/classes/function.py
leotrs: Good catch!

leotrs:
@maximelucas Thanks, just addressed all your concerns I think.

maximelucas (approved):

PR #208, opened 2022-11-01
Implement set operations for NodeView and EdgeView

leotrs:
Since NodeView and EdgeView ultimately inherit from `collections.abc.Set`, the user should be able to perform set operations on them. As reported in #206 this was not the case. This PR fixes that.

New functionality:
```python3
    H = hyperwithattrs
    nodes1 = H.nodes.filterby_attr("color", "blue")
    nodes2 = H.nodes.filterby("degree", 2, "geq")
    assert set(nodes2 - nodes1) == {3, 4}
    assert set(nodes1 - nodes2) == set()
    assert set(nodes1 & nodes2) == {2, 5}
    assert set(nodes1 | nodes2) == {2, 3, 4, 5}
    assert set(nodes1 ^ nodes2) == {3, 4}
```
Before this PR, any of the set operations would raise an exception.

The fix was easy. Just had to override `collections.abc.Set._from_iterable` to play well with the rest of the class.

Also added mention of this to the Tutorial.

### DIFF
--- a/tests/classes/test_reportviews.py
+++ b/tests/classes/test_reportviews.py
@@ -242,3 +242,23 @@ def test_bool(edgelist1):
     assert bool(H.edges) is False
     H = xgi.Hypergraph(edgelist1)
     assert bool(H.edges) is True
+
+
+def test_set_operations(hyperwithattrs):
+    H = hyperwithattrs
+
+    nodes1 = H.nodes.filterby_attr("color", "blue")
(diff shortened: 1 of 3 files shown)


maximelucas (approved):

maximelucas:
Wow nice. The result of `nodes2 - nodes1` is still a `NodeView` right?

leotrs:
> Wow nice. The result of `nodes2 - nodes1` is still a `NodeView` right?

Yep!

Issue #163, opened 2022-09-01
Compress edges


nwlandry:
Implement way of collapsing identical elements. Right now, `IDView.duplicates()` simply chooses the lowest ID as the non-duplicate edge, but there are several ways of
- weighting the collapsed nodes/edges:
  1. Leave the resulting single node/edge unweighted
  2. Weighting the resulting single node/edge by its original multiplicity
- labeling the node/edge:
  1. Give the new node/edge the smallest id of the duplicates
  2. Give the new node/edge a tuple ID where the tuple are the nodes/edges that were compressed.
  3. Give the new node/edge a completely new ID
This would affect the duplicates method and potentially require another method.

Something along these lines is the [`collapse_identical_elements`](https://github.com/pnnl/HyperNetX/blob/master/hypernetx/classes/entity.py) in HyperNetX.

nwlandry:
What do we think the syntax should be for calling this? Originally, I was thinking `H.edges.collapse(weighted=True, relabel="tuple")` and same for nodes, but unfortunately, this would involve modifying the hypergraph from the IDView, which seems to contradict the notion of what a view is. What about `H.collapse_edges()` and `H.collapse_nodes()`?

leotrs:
Agreed this method does not belong in the `EdgeView`. It belongs in `Hypergraph`. The main objective of `Hypergraph` is after all to modify the network's structure.

What about `H.collapse_edges(weighted=weight_mode, index=index_mode, return_collapse_dict)`. Here "weight_mode" could be None (no weight) or str (used as the name of the attribute that stores the weight), and "index_mode" could be one of "tuple", "min", or "auto". Auto means assign a completely new ID automatically.

I'm using "index" instead of "label" because elsewhere in the codebase we refer to nodes and edges using the word "ID", not the word "label". What do you think?

maximelucas:
Would "merge" rather "collapse" make sense?
Collapse sounds a bit dramatic 😄

leotrs:
As per #177, the simplicial complex class will (might?) have a method that "collapses" things so it might be wise to choose a different name for this one. I agree merge sounds nice.

nwlandry:
Some things I'm wondering:
* What should we do with overlapping attributes? Make them a list?
* Similarly, each edge could, in principle, have a weight. Should these be used in computing the merged edge weight? If that's the case, should we store it in a "weight" attribute?
[...]

PR #210, opened 2022-11-03
Merge edges

nwlandry:
Partially addressing #163.


xgi/classes/hypergraph.py
```diff
@@ -935,6 +936,77 @@ def clear_edges(self):
...
+        """Merges edges which have the same members.
+
+        Parameters
+        ----------
+        rename : str, optional
+            Either "first", "tuple", or "new", by default "first"
+            If "first", the new edge ID is the first of the sorted
+            duplicate edge IDs. If "tuple", the new edge ID is a 
+            tuple of the sorted duplicate edge IDs. If "new", the
+            _edge_uid counter is selected and incremented.
```
leotrs, line 951: We should assume the user has no knowledge of `_edge_uid` so we shouldn't mention it.
```suggestion
            tuple of the sorted duplicate edge IDs. If "new", a new ID will be
            selected automatically.
```

xgi/classes/hypergraph.py
```diff
@@ -935,6 +936,77 @@ def clear_edges(self):
...
+        merge_rule : str, optional
+            Either "first" or "union", by default "first"
+            If "first", takes the attributes of the first duplicate.
+            If "union", takes the set of attributes of all the duplicates.
+        multiplicity : bool, optional
+            Whether to return the multiplicity of the hyperedge,
+            by default True
```
leotrs, line 958: This method does not return anything.

xgi/classes/hypergraph.py
```diff
@@ -935,6 +936,77 @@ def clear_edges(self):
...
+                    new_id = sorted(dup_ids)[0]
+                elif rename == "tuple":
+                    new_id = tuple(sorted(dup_ids))
+                elif rename == "new":
+                    new_id = next(self._edge_uid)
+                else:
+                    raise XGIError("Invalid ID renaming scheme!")
+
+                if merge_rule == "first":
+                    id = sorted(dup_ids)[0]
```
leotrs, line 990:
```suggestion
                    id = min(dup_ids)
```

xgi/classes/hypergraph.py
leotrs, line 958: How about letting the user choose what is the name of the multiplicity attribute? For example, a user might want to keep this multiplicity stored in an attribute called "weight" instead of "multiplicity".

If you agree, I'd suggest making the `multiplicity` parameter be a string and use the value None to not store anything.

tests/classes/test_hypergraph.py
```diff
@@ -462,6 +462,37 @@ def test_clear_edges(edgelist1):
     assert len(H.edges) == 0
 
 
+def test_merge_duplicate_edges(hyperwithdupsandattrs):
+    H = hyperwithdupsandattrs.copy()
+    H.merge_duplicate_edges()
+    assert H.num_edges == 2
+    assert set(H.edges) == {0, 3}
+    assert H.edges[0] == {"color": "blue", "multiplicity": 3}
+    assert H.edges[3] == {"color": "purple", "multiplicity": 2}
```
leotrs, line 471: Please add a tiny integration test:
```python3
assert H.edges.attrs("multiplicity").asdict() == {0: 3, 3: 2}
```

leotrs (changes requested):
Great! Just some minor comments.

And a question: I think we had discussed at some point refactoring the output of `H.duplicate_edges` once this was implemented. Or am I remembering incorrectly?

xgi/classes/hypergraph.py
maximelucas: dup_ids (edge IDs) are always integers right? Just checking that those will always work

xgi/classes/hypergraph.py
```diff
@@ -935,6 +936,77 @@ def clear_edges(self):
...
+        rename : str, optional
+            Either "first", "tuple", or "new", by default "first"
+            If "first", the new edge ID is the first of the sorted
+            duplicate edge IDs. If "tuple", the new edge ID is a 
+            tuple of the sorted duplicate edge IDs. If "new", the
+            _edge_uid counter is selected and incremented.
+        merge_rule : str, optional
+            Either "first" or "union", by default "first"
+            If "first", takes the attributes of the first duplicate.
+            If "union", takes the set of attributes of all the duplicates.
```
maximelucas, line 955: Not fully clear to me what the result looks like explicitly. Something like `"color": {"red", "blue"}`?
Maybe examples would be useful.

maximelucas (commented):
Nice Nicho, thanks.

Adding an "intersection" merge_rule could also make sense to me. For each set of merging nodes, set the attribute "color" to "red" if all merging nodes are red, otherwise discard the color attribute.

tests/classes/test_hypergraph.py
nwlandry: Done.

xgi/classes/hypergraph.py
nwlandry: Sounds good.

xgi/classes/hypergraph.py
nwlandry: I checked and the min will work for both.

xgi/classes/hypergraph.py
nwlandry: Fixed.

nwlandry:
> Nice Nicho, thanks.
> 
> Adding an "intersection" merge_rule could also make sense to me. For each set of merging nodes, set the attribute "color" to "red" if all merging nodes are red, otherwise discard the color attribute.

I like this idea and am working on it now!

xgi/classes/hypergraph.py
nwlandry: Okay, I added a lot to the docstring...hopefully this addresses this?

maximelucas (approved):

xgi/stats/__init__.py
```diff
@@ -480,14 +480,6 @@ def nodestat_func(func):
...
 
     However, `my_degree` is not recognized as a node statistic.
-
-    >>> H.my_degree()
-    Traceback (most recent call last):
-    AttributeError: stat "my_degree" not among available node or edge stats
-    >>> H.nodes.my_degree
-    Traceback (most recent call last):
-    AttributeError: Stat 'my_degree' not defined
-
```
leotrs, line 490: Do we really just want to remove this part of the docstring instead of fixing it? If we just remove this chunk, now the surrounding test doesn't really make sense.

leotrs (approved):

xgi/stats/__init__.py
nwlandry: This should be fixed but just waiting for Github Actions to finish to double check.

leotrs: Suuuper weird this happened, I don't see why this would happen. In the interest of merging this quickly, why don't we just
```python3
AttributeError:...
```

nwlandry: Okay, this fixed it! Merging now.

Issue #181, opened 2022-10-03
old argument names in plotting tutorial


maximelucas:
For example, `edge_lc` should be changed to `dyad_lw`.

mcontisc:
I think it was solved in this commit d1975e6 by @iaciac.

PR #211, opened 2022-11-04
Partial fix #174 

### LINKED ISSUE #174: Speed up `draw_xgi_nodes` and `draw_xgi_hyperedges`
If we remove the for-loop in the function that draws the hypergraph nodes with `scatter` and instead pass iterables into the scatter function, it could be faster. Also, if we plot Collection objects instead of Patch objects, maybe it will be faster as well. (https://stackoverflow.com/questions/37573314/matplotlib-efficient-way-to-create-large-number-of-patch-objects)

iaciac:
Nodes are now plotted with iterables. 
The plotting of hyperedges has to be improved, but it is not as straightforward.

### DIFF
--- a/tutorials/Tutorial 5 - Plotting.ipynb
+++ b/tutorials/Tutorial 5 - Plotting.ipynb
@@ -108,13 +108,13 @@
     "cmap = plt.cm.Paired\n",
     "\n",
     "ax = plt.subplot(1, 2, 1)\n",
-    "xgi.draw(H, pos, ax=ax, edge_face_colormap=cmap)\n",
+    "xgi.draw(H, pos, ax=ax, edge_fc_cmap=cmap)\n",
     "\n",
     "# Qualitative colormap\n",
     "cmap = plt.cm.Purples\n",
     "\n",
     "ax = plt.subplot(1, 2, 2)\n",
-    "xgi.draw(H, pos, ax=ax, edge_face_colormap=cmap)"
+    "xgi.draw(H, pos, ax=ax, edge_fc_cmap=cmap)"
    ]
   },
   {
@@ -149,7 +149,7 @@
     "    node_ec=node_ec,\n",
     "    node_lw=node_lw,\n",
     "    node_size=node_size,\n",
-    "    edge_face_colormap=cmap,\n",
+    "    edge_fc_cmap=cmap,\n",
     ")"
    ]
   },
@@ -307,7 +307,7 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3.10.4 ('hypergraph')",
+   "display_name": "Python 3 (ipykernel)",
    "language": "python",
    "name": "python3"
   },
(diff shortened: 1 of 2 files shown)


xgi/drawing/xgi_pylab.py
```diff
@@ -206,17 +206,15 @@ def draw_xgi_nodes(
...
-        )
+    ax.scatter(
+        x=[pos[i][0] for i in H.nodes],
+        y=[pos[i][1] for i in H.nodes],
+        s=[node_size[i] ** 2 for i in H.nodes],
+        c=[node_fc[i] for i in H.nodes],
+        edgecolors=[node_ec[i] for i in H.nodes], 
+        linewidths=[node_lw[i] for i in H.nodes],
+        zorder=zorder,
+    )
```
leotrs, line 217: This is even faster as it only iterates over the nodes once:
```suggestion
    x, y, s, c, ec, lw, = zip(*[(pos[i][0], pos[i][1], node_size[i], node_fc[i], node_ec[i], node_lw[i]) for i in H.nodes])
    ax.scatter(x=x, y=y, s=s, c=c, edgecolors=ec, linewidths=lw, zorder=zorder)
```
(Or, uh, something like this, I haven't tested it!)

leotrs (commented):
Thanks for PRing! Just a quick suggestion.

xgi/drawing/xgi_pylab.py
```diff
@@ -206,17 +206,8 @@ def draw_xgi_nodes(
...
-        ax.scatter(
-            x,
-            y,
-            s=node_size[i] ** 2,
-            c=node_fc[i],
-            edgecolors=node_ec[i],
-            linewidths=node_lw[i],
-            zorder=zorder,
-        )
+    x, y, s, c, ec, lw, = zip(*[(pos[i][0], pos[i][1], node_size[i], node_fc[i], node_ec[i], node_lw[i]) for i in H.nodes])
```
iaciac, line 209:
```suggestion
    x, y, s, c, ec, lw, = zip(*[(pos[i][0], pos[i][1], node_size[i]**2, node_fc[i], node_ec[i], node_lw[i]) for i in H.nodes])
```

iaciac (commented):
Thanks @leotrs, just the node size adjustment was missing

leotrs (approved):

maximelucas:
Thank Iac. Just put a commented related to #181 but already approved.

maximelucas (approved):

nwlandry:
@iaciac can you give a comparison on the speed of drawing nodes before and after your PR?

iaciac:
If I remember correctly it's 3 times faster. I need to check

tutorials/Tutorial 5 - Plotting.ipynb
```diff
@@ -108,13 +108,13 @@
     "cmap = plt.cm.Paired\n",
```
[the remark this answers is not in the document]
nwlandry: Okay, I looked into this and fixed it. The reason why it still runs (the incorrect args simply do nothing) is because we have **kwargs as an argument to update the default drawing settings (it was too wordy to write each setting out) and so it can accept any kwarg. If it doesn't mean anything, however, nothing will happen.

PR #212, opened 2022-11-07
Simplicial Synchronization

Marconurisso:
- Added functions to build boundary matrices and Hodge Laplacians for oriented simplicial complexes. The orientation is implemented as a dictionary associating to each simplex ID a boolean value: 0 for the orientation coherent with the node ordering in the object and 1 for the opposite one.
- Added functions to simulate the simplicial Kuramoto model and compute its order parameter.
- Modified the _subfaces method to allow it to compute only the adjacent subfaces.
- Added some tests and an example notebook.
- Added the new synchronization functions to the documentation.


xgi/classes/simplicialcomplex.py
```diff
@@ -186,12 +186,16 @@ def add_simplex(self, simplex, **attr):
             faces = self._subfaces(simplex)
             self.add_simplices_from(faces)
 
-    def _subfaces(self, simplex):
+    def _subfaces(self, simplex, all=True):
```
maximelucas, line 189: We forgot to document the new argument. A docstring a bit more verbose would useful too.

xgi/dynamics/synchronization.py
```diff
@@ -92,3 +98,149 @@ def compute_kuramoto_order_parameter(H, k2, k3, w, theta, timesteps=10000, dt=0.
...
+def simulate_simplicial_kuramoto(
+    S,
+    orientations=None,
+    order=1,
+    omega=[],
+    sigma=1,
+    theta0=[],
+    T=10,
+    n_steps=10000,
+    index=False,
```
maximelucas, line 112: index  argument not document

xgi/linalg/matrix.py
```diff
@@ -361,3 +363,149 @@ def clique_motif_matrix(H, index=False):
...
+    nu = len(simplices_u_ids)
+
+    simplices_d_dict = dict(zip(simplices_d_ids, range(nd)))
+    simplices_u_dict = dict(zip(simplices_u_ids, range(nu)))
+
+    if index:
+        rowdict = {v: k for k, v in simplices_d_dict.items()}
+        coldict = {v: k for k, v in simplices_u_dict.items()}
+
+    if orientations == None:
```
leotrs, line 428:
```suggestion
    if orientations is None:
```

xgi/linalg/matrix.py
```diff
@@ -361,3 +363,149 @@ def clique_motif_matrix(H, index=False):
...
+        orientations = {
+            idd: 0 for idd in list(S.edges.filterby("order", 1, mode="geq"))
```
leotrs, line 430:
```suggestion
            idd: 0 for idd in S.edges.filterby("order", 1, mode="geq")
```

xgi/linalg/matrix.py
```diff
@@ -361,3 +363,149 @@ def clique_motif_matrix(H, index=False):
...
+        }
+
+    B = np.zeros((nd, nu))
+    if not (nu == 0 or nd == 0):
+        if order == 1:
+            for u_simplex_id in simplices_u_ids:
+                u_simplex = S.edges.members(u_simplex_id)
+                matrix_id = simplices_u_dict[u_simplex_id]
+                head_idx = list(u_simplex)[1]
+                tail_idx = list(u_simplex)[0]
```
leotrs, line 440: `u_simplex` is a set, no? Then I would very much recommend AGAINST converting a set to list and then using the resulting order. How/when is it guaranteed that `list(u_simplex)[1]` is the head and that `list(u_simplex)[0]` is the tail?

Marconurisso: In order to have orientation, one needs to extract some "reference orientation" for the simplices of the `SimplicialComplex` object. In this implementation, I chose the order in which the nodes appear in the set representing each simplex. If, for some reason, the order of the nodes in the set changes, then it is possible that the orientation of the simplex will be changed too. This can, however, be easily fixed by changing the orientation value in the input dictionary. I see that this is not optimal, but I don't see another simple way to do it without adding a whole new class explicitly designed to account for orientation.

leotrs: The problem is that sets are *not* ordered, there is no such a thing as "the order in which the nodes appear in the set". When you do `list(u_simplex)`, the returned list is not guaranteed to be always the same (at least not across machines/versions/environments). I am actually surprised that this PR passed the automated tests for this reason.

Perhaps we can do `head_idx, tail_idx = max(u_simplex), min(u_simplex)`, or something similar. Here you are taking the reference orientation to be that given by the sorted node IDs. Since most of the time the IDs are integers or strings, this is always well defined.

@maximelucas what do you think?

xgi/classes/simplicialcomplex.py
Marconurisso: Done

xgi/linalg/matrix.py
Marconurisso: Done

xgi/linalg/matrix.py
Marconurisso: Done

xgi/linalg/matrix.py
Marconurisso: Good point. Now the reference orientation is the one given by the sorted nodes. I think that this should solve the issues if we can assume that node labels are always strings or numbers.

xgi/linalg/matrix.py
```diff
@@ -425,19 +424,20 @@ def boundary_matrix(S, order=1, orientations=None, index=False):
...
 
     B = np.zeros((nd, nu))
     if not (nu == 0 or nd == 0):
         if order == 1:
             for u_simplex_id in simplices_u_ids:
-                u_simplex = S.edges.members(u_simplex_id)
+                u_simplex = list(S.edges.members(u_simplex_id))
+                u_simplex.sort(
+                    key=lambda e: (isinstance(e, str), e)
+                )  # Sort for reference orientation
```
leotrs, line 437: Given that this is a fairly surprising thing to do, please leave a comment in the code explaining why the tuple `(<bool>, e)` is necessary.

leotrs (approved):
Thanks @Marconurisso for your first contribution! :tada:  I left one more comment, but am approving now.

xgi/linalg/matrix.py
Marconurisso: Done

leotrs:
Perfect! feel free to merge when ready.

maximelucas (approved):

PR #213, opened 2022-11-07
Random models faster

maximelucas:
Speed up of `random_hypergraph()` and `random_simplicial_complex()`.
Running 
```python
%%timeit
N = 100 
ps = [0.00001, 0.000001, 0.00000000001]
H = xgi.random_hypergraph(N, ps)

```
we had 
```python
# -> 225 ms ± 2.04 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
and now
```python
# -> 154 ms ± 483 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

Because I switched to the numpy random number generator, the tests are not happy anymore. Not managing to make the seed work correctly so far.


xgi/generators/nonuniform.py
```diff
@@ -247,7 +248,7 @@ def dcsbm_hypergraph(k1, k2, g1, g2, omega, seed=None):
     return H
 
 
-@py_random_state("seed")
+@np_random_state(0)
```
leotrs, line 251: This should either be `np_random_state(2)` or `np_random_state("seed")`. Are those not working?

maximelucas: With `np_random_state(2)` the error seems slightly different: 

```python
______________________________ test_compute_kuramoto_order_parameter _______________________________

    @pytest.mark.slow
    def test_compute_kuramoto_order_parameter():
>       H1 = xgi.random_hypergraph(100, [0.05, 0.001], seed=0)

tests/dynamics/test_synchronization.py:11:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
xgi/utils/decorators.py:578: in func
    return argmap._lazy_compile(__wrapper)(*args, **kwargs)
<class 'xgi.utils.decorators.argmap'> compilation 4:4: in argmap_random_hypergraph_1
    ???
xgi/generators/nonuniform.py:285: in random_hypergraph
    rng = np.random.default_rng(seed=seed)
_generator.pyx:4849: in numpy.random._generator.default_rng
    ???
_pcg64.pyx:121: in numpy.random._pcg64.PCG64.__init__
    ???
bit_generator.pyx:517: in numpy.random.bit_generator.BitGenerator.__init__
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

>   ???
E   TypeError: SeedSequence expects int or sequence of ints for entropy not RandomState(MT19937)

bit_generator.pyx:299: TypeError
```

leotrs: Well damn. Yeah we seriously have to refactor the random state stuff. This was copy/pasted from NX, no?

maximelucas: Pretty much yea

nwlandry: I removed decorators in PR #227 so you can see what I mean. IDK if we decide to do that, but IMO it simplifies things a bit.

leotrs: The decorators are there so that the function can accept either a seed number or an object that generates a seed. The former is good for users and the latter is good for testing, and it also supports newer numpy syntax for random stuff.

Having said that, I never liked that we just depended on this black box module that was copy-pasted from NX. So I suggest we remove those decorators for now and make it a priority issue to figure out how to support modern numpy random stuff.*

* By "modern random stuff" I mean the attached, taken from [here](https://numpy.org/doc/stable/reference/random/generated/numpy.random.choice.html).
2022-11-17-074228_936x157_scrot

nwlandry:
@maximelucas I did a check comparing the number of edges using the old method versus the new method and verified that the number of edges is preserved, but as you anticipated, because we are switching from a random method to a numpy random method, the specific edges aren't the same.

maximelucas:
Great thanks @nwlandry! Had started locally but never got round to finish it :)
Updated the last one now, and fixed outputs and added a test.

PR #216, opened 2022-11-08
Hotfix 214

### LINKED ISSUE #214: bug in flag_complex
Without the `ps` argument, `xgi.flag_complex()` fills all 4 triangles of the complete graph, as expected:

```python
import networkx as nx
import xgi

G = nx.complete_graph(n=4)
S = xgi.flag_complex(G)
xgi.num_edges_order(S, 2)
# -> 4
```
Including it as `ps=[1]` should be equivalent  
```python
S = xgi.flag_complex(G, ps=[1])
xgi.num_edges_order(S, 2)
# -> 0 # huh?
```
We are only computing maximal cliques and feeding them to `add_simplices_from` which might not be adding subfaces as expected?

maximelucas:
- fixed #214 : all simplices are added correctly with max_order now
- added a `powerset` function to compute all subsets of a set
- added a `subfaces` function. 
- added tests and examples for all the above, and included them in docs

The `subfaces` extends the current private `_subfaces` present in `SimplicialComplex.py` with a few options, and is available publicly. I think it should ultimately replace `_subfaces`.

### DIFF
--- a/xgi/utils/utilities.py
+++ b/xgi/utils/utilities.py
@@ -57,7 +57,8 @@ def powerset(iterable, include_empty=False, include_full=False):
 
     Examples
     --------
-    >>> list(powerset([1,2,3,4])) # doctest: +NORMALIZE_WHITESPACE
+    >>> import xgi
+    >>> list(xgi.powerset([1,2,3,4])) # doctest: +NORMALIZE_WHITESPACE
     [(1,), (2,), (3,), (4,), (1, 2), (1, 3), (1, 4), (2, 3), (2, 4), (3, 4),
      (1, 2, 3), (1, 2, 4), (1, 3, 4), (2, 3, 4)]
 
--- a/xgi/utils/utilities.py
+++ b/xgi/utils/utilities.py
@@ -1,6 +1,8 @@
 """General utilities."""
 
 from collections import defaultdict
+from itertools import chain, combinations
+
 
 __all__ = ["dual_dict", "powerset"]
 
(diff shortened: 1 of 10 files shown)


xgi/utils/utilities.py
```diff
@@ -35,3 +36,42 @@ def dual_dict(edge_dict):
...
+
+    if include_empty:
+        start = 0
+    else:
+        start = 1
+
+    if include_full:
+        end = 1
+    else:
+        end = 0
```
leotrs, line 75: Nitpick:
```suggestion
    start = 0 if include_empty else 1
    end = 1 if include_fill else 0
```

xgi/utils/utilities.py
```diff
@@ -35,3 +36,42 @@ def dual_dict(edge_dict):
...
+
+    s = list(iterable)
```
leotrs, line 77: Question: doesn't `combinations` accept an iterable? Why convert to list?

leotrs (approved):
Approved, feel free to ignore my nitpick.

xgi/utils/utilities.py
maximelucas: Not sure, took it from itertools recipes (https://docs.python.org/3/library/itertools.html#itertools-recipes).
I'd keep it for safety, but shouldn't change much as the input should be a single edge.

xgi/utils/utilities.py
maximelucas: Done

PR #220, opened 2022-11-09
Updated add_simplices_from to match add_hyperedges_from

### LINKED ISSUE #218: max_order not taking attributes into account in add_simplices_from
Currently, the input of `add_simplices_from` needs to be a list of simplices, each simplex of the format `(1,2,3)` or with attributes, say `(1,2,3, {"color": "red}`. The `len` of the simplex is then 3 or 4 even though its size is 3. The initial part of the function, dealing with the `max_order` parameter, does not take the possibility of the attributes into account: 

https://github.com/ComplexGroupInteractions/xgi/blob/254557062b73fb5a93efeb74026534d6347207db/xgi/classes/simplicialcomplex.py#L232-L240

Adding the simplex `(1,2,3, {"color": "red"})` to a simplicial complex, we expect it to be added itself with its attribute, and all its edges and singleton nodes without attributes.

The actual result is the following:

```python

edges = [(1,2,3, {"color": "red"})]
max_order = 2

for edge in edges:
    if len(edge) >  max_order + 1:
        print(list(combinations(edge,  max_order + 1)))

## --> [(1, 2, 3), (1, 2, {'color': 'red'}), (1, 3, {'color': 'red'}), (2, 3, {'color': 'red'})]
```

The simplex `(1, 2, 3)` will still be added, but without attributes. Its edges will be added, ~~but with the attributes.~~ (EDIT: `S.add_simplices_from([(1,2,3, {"color": "red"})], max_order=2)` adds the pairwise edges without attributes, not sure why.) Note that we cannot see this error by looking only at `S.edges.members()` because the members are all correct.

More generally, are we not making our lives difficult by allowing some input edges to have attributes and others not? If that's a common let's keep it. If not, we could have a check at the beginning of the function to check if there are attributes, and deal with it more easily I guess.

maximelucas:
Updated `add_simplex` and `add_simplices_from`. Accepted input formats have changed to match the hypergraph methods. 

Also fixes #218, and the new format is to use  `S.add_simplices_from([((1,2,3), {"color": "red"})], max_order=2)`.

(I didn't try to speedup anything yet despite the name of the branch, because I found this discrepancy first)

### DIFF
--- a/xgi/classes/simplicialcomplex.py
+++ b/xgi/classes/simplicialcomplex.py
@@ -6,6 +6,7 @@ and can associate key/value attribute pairs with each undirected simplex and nod
 Multi-simplices are not allowed.
 """
 
+from collections.abc import Hashable, Iterable
 from itertools import combinations, count
 
 from ..exception import XGIError
@@ -167,7 +168,7 @@ class SimplicialComplex(Hypergraph):
         Examples
         --------
 
-        Add simplices with or without specifying an edge id.
+        Add simplices with or without specifying an simplex id.
 
         >>> import xgi
         >>> S = xgi.SimplicialComplex()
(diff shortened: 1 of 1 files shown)


leotrs (approved):
Thanks for this. No comments, feel free to merge when ready.

My only thought is that now `Hypergraph.add_edges_from` and `SimplicialComplex.add_simplices_from` are VERY long and VERY similar, and it will be a pain to maintain both. This is precisely why we need a common interface for both classes....

maximelucas:
> My only thought is that now `Hypergraph.add_edges_from` and `SimplicialComplex.add_simplices_from` are VERY long and VERY similar, and it will be a pain to maintain both. This is precisely why we need a common interface for both classes....

Agreed of course

Issue #221, opened 2022-11-09
flag_complex is slow because add_simplices_from is slow


maximelucas:
I have a pairwise dataset with 634 nodes and 18624 edges.
I want to fill its empty triangles with probability p. 
We have a function `flag_complex` that does exactly this, but it's slow: after running an hour, I gave up. 

It's slow because, to enforce simpliciality of a `SimplicialComplex`, the method `add_simplices_from` add its input hyperedges, but also all their subfaces down to pairwise edges. (Even more so, many of those subfaces might be shared between hyperedges, but we're not using that fact to our advantage.)

So, even pre-computing the triangles (with p=0.2), it's slow: 
```python
S = xgi.SimplicialComplex()
S.add_nodes_from(nodes) # instantaneous
S.add_simplices_from(edges) # takes a few minutes
S.add_simplices_from(triangles) # gave up after a few minutes
```

On the other hand, the `Hypergraph` class does not add all subfaces because it does not need to. So, this runs virtually instantaneously: 
```python
H = xgi.Hypergraph()
H.add_nodes_from(nodes)
H.add_edges_from(edges)
H.add_edges_from(triangles)
```

Of course this "solution" is less safe from the user point of view because it allows multiedges (not allowed in `SimplicialComplex`) and does not enforce simpliciality (`SimplicialComplex` does). So it is much more error prone. 

My current temporary solution is to do the above and pre-compute the triangles to emulate the `flag_complex` up to order 2 as follows, given a networkx Graph G: 

```python
import random
def find_triangles(G):
    triangles = set(frozenset((n, nbr, nbr2)) for n in G for nbr, nbr2 in combinations(G[n],2) if nbr in G[nbr2])
    return [set(tri) for tri in triangles]

triangles_empty = find_triangles(G)
p = 0.2
triangles = [el for el in triangles_empty if random.random() <= p]
```
This runs in a few seconds on my dataset.

leotrs:
Another problem is that `add_simplices_from` is recursive. As a rule of thumb, please don't implement recursive functions or methods in XGI. At least not when the recursion is expected to go more than a few times. Python is not optimized for deep recursive calls, and even without taking recursion into account, each new call to a function is more costly in python than it is in other languages.

In this case, instead of recursion I would keep a set of simplices that have yet to be added in memory and work through adding that set. Adding all subfaces within the same function call (i.e. no recursion) and keeping things in a set instead of a list (i.e. no repetitions) is surely going to improve things.

[...]

PR #223, opened 2022-11-10
speedup add_simplices_from

maximelucas:
`add_simplices_from` is slow, see #221. Profiling showed about 90% of the time is spent in the line checking existence of the simplex: 
```python
if not members or self.has_simplex(members):
```

`has_simplex()` now access members through `S._edge` rather than `S.edges.members(s)` which is significantly faster (see #221 for timings).

Here, adding 100 triangles seems about 10x faster:

```python
%timeit S.add_simplices_from(list(triangles)[:100]) 
# old -> 9.13 ms ± 28.4 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
# new -> 860 µs ± 5.18 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

The `add_simplices_from` can probably be improved even more, by rewriting it it without recursion for example. I have written a version but not pushed it because the seped improvement seems negligible in my case (it might not be with larger simplices). Also, it modifies the method much more, so has a larger chance of new bugs. And it changes the order in which simplices are added, so tests would need to be udpated. I would start with this minimal update at first, and maybe use the other version for a second step.

### DIFF
--- a/xgi/classes/simplicialcomplex.py
+++ b/xgi/classes/simplicialcomplex.py
@@ -612,4 +612,4 @@ class SimplicialComplex(Hypergraph):
         False
 
         """
-        return frozenset(simplex) in (s for s in self._edge.values())
+        return frozenset(simplex) in self._edge.values()
--- a/xgi/classes/simplicialcomplex.py
+++ b/xgi/classes/simplicialcomplex.py
@@ -612,4 +612,4 @@ class SimplicialComplex(Hypergraph):
         False
 
         """
-        return set(simplex) in (set(self.edges.members(s)) for s in self.edges)
+        return frozenset(simplex) in (s for s in self._edge.values())
--- a/xgi/utils/utilities.py
+++ b/xgi/utils/utilities.py
@@ -38,7 +38,7 @@ def dual_dict(edge_dict):
     return dict(node_dict)
 
 
-def powerset(iterable, include_empty=False, include_full=False):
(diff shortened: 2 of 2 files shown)


maximelucas:
Removed unnecessary generator expression in `has_simplex`, yielding further speed up (thanks profiling).

Timings, wrt to previous speedup. For
```python
S = xgi.SimplicialComplex()
simplices = [{0, 1, 2, 3}, {4}, {5, 6}, {6, 7, 8}]
```

```python
%timeit S.add_simplices_from(simplices, max_order=2)
# old -> 7.45 µs ± 10.9 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
# new -> 4.72 µs ± 16.9 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

```python 
%timeit xgi.random_simplicial_complex(250, [0.1, 0.001])
# old -> 2.92 s ± 65.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
# new -> 1.72 s ± 42.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

```python
%timeit S.add_simplices_from(triangles) # 61712 triangles
# old -> 2min 20s ± 331 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
# new -> 1min 17s ± 801 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

So we get an additional 2x speedup (especially when adding more simplices).

PR #225, opened 2022-11-15
Fix #224 uid counter initialisation

### LINKED ISSUE #224: bug: add_edge removes other edge
Create a Hypergraph:
```python
import xgi
H = xgi.Hypergraph({0: {1, 2}, 1: {1, 4}, 
                     2: {0, 1, 8}, 3: {0, 3, 5}})
H._edge
# -> {0: {1, 2}, 1: {1, 4}, 2: {0, 1, 8}, 3: {0, 3, 5}} # all fine
```
Then remove an edge
```python
H.remove_edge(2)
H._edge
# -> {0: {1, 2}, 1: {1, 4}, 3: {0, 3, 5}} # all fine
```

But then, when adding a new hyperedge, [1,9,2], it is assigned the ID 0 already in use, and the previous hyperedge with ID 0 has disappeared:
```python
H.add_edge([1,9,2])
H._edge
# {0: {1, 2, 9}, 1: {1, 4}, 3: {0, 3, 5}} # wrong
```
It should just be assigned a new ID, probably 4. Not sure what is causing this yet.

Instead, specifying the ID explicitly works, and gives the expected result, but the above is still wrong:
```python
H.add_edge([1,9,2], id=4)
H._edge
# -> {0: {1, 2}, 1: {1, 4}, 3: {0, 3, 5}, 4: {1, 2, 9}} # correct
```

maximelucas:
- fixed #224 counter initialisation in `add_edges_from` and `add_simplices_from`
- fixed `H.copy()` so that it does not throw an error for string IDs
- added tests for all


leotrs:
Thanks. I find this solution a bit convoluted. Since the edge uid is meant to be internal, it doesn't *really* matter what ids we associate to edges, right? Why not do something like this?
```python3
class Hypergraph:
    def __init__(self, incoming_data, ...):
        ...
        self._edge_uid = count(start=len(incoming_data))
        ...
```
Or something like this? In this way we know we will never generate uids that clash with whatever is in `incoming_data`.

maximelucas:
Good point, that's simpler. 
I actually stole the idea from what we had in `H.copy()`. But I guess there the convoluted solution is needed to have the count as the original?

leotrs:
I think we need to have the same count **while** we make the copy. After we make the copy, I wouldn't really care that whether they are the same.

maximelucas:
Yea but I guess I would hope that if I add an edge to the original and to the copy, they get the same id by default. No?

xgi/classes/hypergraph.py
```diff
@@ -668,6 +685,24 @@ def add_edges_from(self, ebunch_to_add, **attr):
...
                 e = next(new_edges)
             except StopIteration:
+
+                if format2 or format4:
+                    # If we don't set the start of self._edge_uid correctly, it will start at 0,
+                    # which will overwrite any existing edges when calling add_edge().  First, we
+                    # use the somewhat convoluted float(e).is_integer() instead of using
+                    # isinstance(e, int) because there exist integer-like numeric types (such as
+                    # np.int32) which fail the isinstance() check.
+                    edges_with_int_id = [
```
nwlandry, line 695: My initial thought on this is that `add_edges()` is not covered - is there a reason for this? Also, perhaps we can add an `update_uid` function or something that evaluates if the ID to be added (in either `add_edges_from` or `add_edge`) is an int and larger than the current uid and if so, updates the edge uid.

maximelucas: You mean `add_edge()` is not covered? I think it was already dealing correctly with new ids. 
The problem that I was fixing was that adding edges (without specifying their ids) was not updating the internal edge_uid. Because of this, any edge added after that was given uid 0 by default, replacing the existing edge with that uid. So the problem was not about the new uid, but about the current uid that was not correct. We could replace the current fix by a function sure.

nwlandry: I think that the same is true for `add_edge()`. If you run
```
H = xgi.Hypergraph()
H.add_edge([1, 2], id=0)
H.add_edge([3, 4])
H.edges.members(dtype=dict)
{0: {3, 4}}
```
I don't think it increments the counter, right?

maximelucas: It does increment it, the output is `{0: {1, 2}, 1: {3, 4}}` and the _edge_ui is `count(2)` :)

nwlandry: Mmmmm. Fair. What about this one, though?
```
import xgi
H = xgi.Hypergraph()
H.add_edge([1, 2], id=1)
H.add_edge([3, 4])
H.add_edge([5, 6])
H.edges.members(dtype=dict)
```

maximelucas: Ooh ok you got me, good one: output `{1: {5, 6}, 0: {3, 4}}` is wrong.

maximelucas:
Looking back into this. One aspect that is not covered at all: when adding edges and specifying their IDs, we don't check if these IDs exist, in any methods, potentially replacing existing edges. 

In those cases, what should we do: 1) throw an error/warning "this edge could not be added because this ID already exists", or 2) change the ID to one that does not exist and add the edge anyway (potentially with a warning)?

In case of numeric IDs, they might not be super meaningful (but also might be) and so adding anyway might be okay. But in case of more descriptive IDs for example strings, then updating might not be great. 

I think I'd go for throwing an error/warning and not adding, letting the user deal with this manually.

maximelucas:
I think it works now. 
The solution is not as concise as I would've liked but not too long either. 
Any ideas welcome, but if we agree it works maybe merge in the meantime? I had problems with the bug again because I had switched to the main branch where it isn't fixed.

xgi/classes/simplicialcomplex.py
```diff
@@ -478,6 +494,24 @@ def add_simplices_from(self, ebunch_to_add, max_order=None, **attr):
             try:
                 e = next(new_edges)
             except StopIteration:
+
+                if format2 or format4:
```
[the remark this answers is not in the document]
maximelucas: Done

maximelucas:
The tests for tutorials 3.11 pass but all others tutorial tests get cancelled, any idea why?

Also, after merging main I had to deal with potential tuple IDs. I know we all agreed but I'm wondering now if allowing tuple IDs might not be so great.

leotrs:
No idea what happened to the tests but I just re-ran them manually and all passed so :shrug:

tests/classes/test_hypergraph.py
```diff
@@ -1,5 +1,6 @@
 import pickle
 import tempfile
+from itertools import count
```
nwlandry, line 3: This import doesn't seem to be used?

xgi/utils/utilities.py
```diff
@@ -79,3 +79,35 @@ def powerset(
...
+    np.int32) which fail the isinstance() check.
+
+    Parameters
+    ----------
+    H : xgi.Hypergraph
+        Hypergraph of which to update the uid counter
+
+    """
+
+    edges_with_int_id = [
```
nwlandry, line 102: The only concern that I have with this is that it would be amazing if we could somehow leverage the `self._edge_uid` to make this an O(1) update instead of an O(|E|) update. As in, once we validate the edges that have been added to update `self._edge_uid`, we shouldn't need to iterate over these edges again when we update again.

tests/classes/test_hypergraph.py
maximelucas: Good catch

xgi/utils/utilities.py
maximelucas: I fully agree. I couldn't think of way now though. Should we merge and raise an issue to improve this?

leotrs: If you looked into it an it turned out to not be straightforward then yes let's just open a new issue.

maximelucas: I have an idea now but might not have time to implement it in the next days. I'll raise the issue.

nwlandry:
Yeah, let's merge. I have an idea of how to fix this as well!

PR #227, opened 2022-11-16
Removed random decorators

nwlandry:
Removed all decorators. One downside is that the random number generator cannot be user-defined, only the seed specified. Hopefully this helps PR #213.


leotrs (approved):
See my comment [here](https://github.com/ComplexGroupInteractions/xgi/pull/213#discussion_r1024806743) for my opinion. Approved.

PR #228, opened 2022-11-21
add cache arg

nwlandry:
Added the ability to cache the hypergraph once it's loaded with `load_xgi_data()`. This is now the default behavior.


nwlandry:
Closed because #261 will handle this.

PR #233, opened 2022-12-01
add the items() method so NodeStats are even more dict-like

leotrs:
This will allow NodeStats to be passed to some functions that expect dicts to iterate over them.

### DIFF
--- a/xgi/stats/__init__.py
+++ b/xgi/stats/__init__.py
@@ -117,6 +117,9 @@ class IDStat:
     def __iter__(self):
         return iter(self._val.items())
 
+    def items(self):
+        return iter(self._val.items())
+
     @property
     def _val(self):
         return self.func(self.net, self.view.ids, *self.args, **self.kwargs)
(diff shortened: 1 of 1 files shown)


maximelucas:
Nice. Would a test be useful?

maximelucas (approved):

xgi/stats/__init__.py
```diff
@@ -117,6 +117,9 @@ def name(self):
     def __iter__(self):
         return iter(self._val.items())
 
+    def items(self):
+        return iter(self._val.items())
```
nwlandry, line 121: Does this need to be wrapped in an iter?

leotrs: Good q - I think the answer is that it does not need to. But I chose to put it in there so it returns the same thing (specifically, the same type), as __iter__. 

...but now that I think back on it, it should probably better return the same type as `dict.items`, not the same type as `NodeStat.__iter__` :thinking:

nwlandry: done.

PR #234, opened 2022-12-02
Add node and hyperedge labels in the plotting functions

mcontisc:
- Update the old docstrings.
- Added the functions `draw_node_labels` and `draw_hyperedge_labels`.
- Updated `Tutorial 5 - Plotting.ipynb` to show the new functionalities. I also added an example to show how to plot a hypergraph with only hyperedges of a certain order.
- So far the labels can be only dict. Thus, we should implement the `.items()` function for NodeStat and EdgeStat.
- So far, the hyperedges labels are only drawn for Hypergraph and not for SimplicialComplex. Could you please give me some feedback whether it would make sense to add this functionality for SimplicialComplex as well?
- So far, I did not introduce any check for the `node_labels` and `hyperedge_labels` because I think the raised errors (from python) are already explicative. Let me know if I should improve them.


tutorials/Tutorial 5 - Plotting.ipynb
```diff
@@ -11,12 +11,13 @@
   },
```
leotrs, line 11: <p>passed* as well</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ComplexGroupInteractions/xgi/pull/234/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

leotrs, line 11: <p>@mcontisc, did you try using <code>xgi.subhypergraph(H, edges=id_edges_order)</code> ? If you tried that and it didn't work then we need to fix it so that it works simply by passing in a bunch of edges cc. @nwlandry </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ComplexGroupInteractions/xgi/pull/234/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

leotrs, line 11: <p>This should be <code>node_labels</code>,no?  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ComplexGroupInteractions/xgi/pull/234/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='22'/>

.idea/.gitignore
```diff
@@ -0,0 +1,8 @@
+# Default ignored files
```
leotrs, line 1: I think we should add `.idea/.gitignore` to the top level `.gitignore`.

leotrs (commented):
Awesome stuff! Just left a couple of comments/questions, nothing major.

tutorials/Tutorial 5 - Plotting.ipynb
mcontisc: <p>Right, I will correct it</p>

mcontisc: <p>ups, thanks!</p>

mcontisc: <p>It works but then it keeps the isolated nodes, and I didn't want them in the plot. Since it is a frozen object, I cannot use the function clean up and that's why I pass the set of nodes as well. </p>

mcontisc: <p>Where should I leave this note?</p>

.idea/.gitignore
mcontisc: For now I removed it, I will add it later on when it creates again

tutorials/Tutorial 5 - Plotting.ipynb
leotrs: Ah sorry for being vague. I just meant "let's write this down somewhere/somehow so that we don't forget".... by which I suppose I mean let's open an issue for it. 

I just opened it #235 .

leotrs (approved):

leotrs:
> So far the labels can be only dict. Thus, we should implement the .items() function for NodeStat and EdgeStat.

See #233.

> So far, the hyperedges labels are only drawn for Hypergraph and not for SimplicialComplex. Could you please give me some feedback whether it would make sense to add this functionality for SimplicialComplex as well?

@maximelucas 

> So far, I did not introduce any check for the node_labels and hyperedge_labels because I think the raised errors (from python) are already explicative. Let me know if I should improve them.

I think it's good for now!

tutorials/Tutorial 5 - Plotting.ipynb
maximelucas: <p>Two thoughts: </p><ul><li>to not keep isolated nodes, you can also do it like this, which is a bit shorter:</li></ul><pre class="ql-syntax" spellcheck="false">order = 2&nbsp;&nbsp;
edges_order = H.edges.filterby('order', order).members()
H_sub = xgi.Hypergraph(edges_order)
</pre><ul><li>when doing this in the past I actually wanted to keep isolated nodes so that I could use the same pos argument to plot different orders, like here https://github.com/maximelucas/plot_simplicial_complex/blob/main/draw_simplicial_complexes.ipynb</li></ul><p>		  Nothing needed for this one, just a thought if we end up providing a helper function to do this</p>

maximelucas: <blockquote>"an edge case" </blockquote><p>Good one Leo </p>

maximelucas:
Great idea, thanks Martina! 
Having it for SimplicialComplex too would make sense yes. But there's already an SC case where it works in the notebook is there not?
Just left a comment in the notebook too.

tutorials/Tutorial 5 - Plotting.ipynb
mcontisc: <p>Thank you @maximelucas : ) I wanted to use subhypergraph to show an additional feature of the library. But if you prefer, I can change !! </p>

mcontisc:
<p>In the notebook I only show how to use <code>node_labels</code> and not <code>hyperedge_labels</code>  because I don't know if make sense to have a label for the simplices </p>

tutorials/Tutorial 5 - Plotting.ipynb
maximelucas: Aah good idea. In this case I'd say it looks a bit too complicated, so I'd go with the shorter one and showcase subhypeprgraph somewhere else :)

leotrs: I believe subhypergraph is already used in another tutorial. On the other hand, it's not a terribly well documented feature so it'd be good to have it on this one too.

... but having said that, maybe adding it here randomly is not a good solution for the fact that it's poorly documented and we should just go with what's best for this particular case.

Ok @maximelucas you convinced me. @mcontisc please use Max's suggestion instead of subhypergraph (sorry I led you down that path!)

maximelucas:
Okay got it, because of potentially too many labels. I'd say in small SCs it might be useful to see them. In big ones with larger orders not really, but then the user can turn the labels off. 

In any case, we're plotting only the maximal simplices, right @iaciac ? And the option to show their labels might helpful to the user.

mcontisc:
<p>Yes, we are plotting only the maximal order.. okay I will implement it as well then and come back to u</p>

mcontisc:
> 

Added! However, it is less intuitive than for hypergraph because now if you pass a dictionary you don't know a priori which one will be the keys kept by the drawing function.. More specifically, with ```hyperedge_labels=True``` it works perfectly fine and shows the id labels. If you pass ```hyperedge_labels=SC.edges.size.asdict()``` does not work because the keys are different from the keys of ```H_``` used in the plot.

maximelucas:
Nice! And great point. Maybe add a note about this in the docstring? And then good to go I'd say.
(In the future we might only store maximal simplices or at least have a way to get maximal simplices which will make this better)

maximelucas (approved):

tutorials/Tutorial 5 - Plotting.ipynb
maximelucas: Yea that was my thinking too: let's showcase it when it's a natural best solution

mcontisc: <p>Done :)</p><p> </p>

leotrs:
Merged! 🎉 Grazie @mcontisc

Issue #244, opened 2022-12-07
other bug in add_simplices_from when specifying ids


maximelucas:
Running 

```python
edges = [({0, 1, 2}, 1), ({1, 2}, 0), ({2, 3, 4}, 2)]
S = xgi.SimplicialComplex()
S.add_simplices_from(edges)
```
we get 

```python
# -> UserWarning: uid 2 already exists, cannot add simplex {2, 3, 4}.
S._edge
# -> {1: frozenset({0, 1, 2}),
#     2: frozenset({0, 1}),
#     3: frozenset({0, 2}),
#     4: frozenset({1, 2})}
```
which is wrong for two reasons: 
1.  {2, 3, 4} should've been added but was not because its ID was taken by subfaces of  {0, 1, 2}
2. {1, 2} was assigned ID 4 even though the input specified 0

Both these issues can be solved by making `add_simplices_from` first add all input simplices **and then** add all their subfaces, rather than adding each input simplex and their subfaces immediately.

Issue #243, opened 2022-12-07
bug in add_simplices_from when specifying uid


maximelucas:
Running 

```python
S = xgi.SimplicialComplex()
S.add_simplices_from([({0, 1, 2}, 0, {})])
```
does not add `(0,1)`. 
That's because we update the uid after adding the subfaces... so it's trying to add `(0,1)` with id=0 but that id already exists...

maximelucas:
fixed in #247

PR #245, opened 2022-12-07
Added lost of tests for SimplicialComplex class

maximelucas:
- added lots of tests
- changed the behaviour of inherited Hypergraph methods such as `add_edge`: instead of raising an error and not adding the simplex, we now add the simplex and raise a warning. 
- a consequence of the previous point is that some HG methods that did not work for SCs now do (and raise a warning): e.g. `copy()`
- added tests to check the new warnings and other old warnings
- fixed an off-by-one in update_uid_counter #243 

The cleanup of the SimplicialComplex is not finished yet but I feel like this PR is getting big already. 
There is one test that does not pass, due to #244. I feel like fixing it would be a lot of code change and would be better in another PR. (The test is new and the fact that it does not pass just highlights a pre-existing bug)

### DIFF
--- a/tests/classes/test_hypergraph.py
+++ b/tests/classes/test_hypergraph.py
@@ -1,5 +1,6 @@
 import pickle
 import tempfile
+from warnings import warn
 
 import pytest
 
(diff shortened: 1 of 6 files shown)


maximelucas:
Yes exactly. Because fixing it will require to rewrite the add_simplices_function as non-recursive and change the orders in which simplices are added, so we will probably need to update existing tests too.

leotrs (approved):

leotrs:
Great! I'd suggest we merge this and then get on working the next PR asap.

PR #251, opened 2022-12-13
Update simplicial kuramoto module

arnaudon:
- remove non-necessary extra dimension for theta0 and omega
- allow use of scipy solve_ivp integrator for better precision
- use random default theta0 and omega, so it does not crash if user does not specify anything (which is allowed from function signature)

### DIFF
--- a/xgi/dynamics/synchronization.py
+++ b/xgi/dynamics/synchronization.py
@@ -1,6 +1,7 @@
 """Simulation of the Kuramoto model."""
 
 import numpy as np
+from scipy.integrate import solve_ivp
 
 import xgi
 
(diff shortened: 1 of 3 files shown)


arnaudon:
Sorry, I use black formatting, and I have strong reflex to use it with keyboard shortcut... I will undo it if you prefer a cleaner diff.

xgi/dynamics/synchronization.py
```diff
@@ -125,20 +122,22 @@ def simulate_simplicial_kuramoto(
     order: integer
         The order of the oscillating simplices
     omega: numpy.ndarray
-        The simplicial oscillators' natural frequencies, has dimension
-        (n_simplices of given order, 1)
+        The simplicial oscillators' natural frequencies, has dimension of n_simplices of given order
+        if None, random will be chosen
```
maximelucas, line 126:
```suggestion
        if None, randomly drawn from a normal distribution
```

xgi/dynamics/synchronization.py
```diff
@@ -125,20 +122,22 @@ def simulate_simplicial_kuramoto(
...
     sigma: positive real value
         The coupling strength
     theta0: numpy.ndarray
-        The initial phase distribution, has dimension
-        (n_simplices of given order, 1)
+        The initial phase distribution, has dimension n_simplices of given order
+        if None, random will be chosen
```
maximelucas, line 131:
```suggestion
        if None, randomly drawn from a normal distribution 
```

arnaudon:
I will close this, as we will just point to the other external package

PR #253, opened 2022-12-15
Remove data refs

nwlandry:
It was suggested to remove the `data` folder and this PR removes all references to the `disGene.txt` although the dataset remains until it is archived in xgi-data. This PR also adds the `to_bipartite_pandas_dataframe` method.


maximelucas (approved):

maximelucas:
In case you think of a few additional tests for the new function, can be safer, but looks good already.

PR #254, opened 2022-12-15
Update xgi_data.py

acuschwarze:
*added a `download_xgi_data` function that download data from gitlab to a local directory
*changed `load_xgi_data` so that it can now load data from online sources and local files. For loading data from online sources, it is now using the index.json and the data stored on gitlab.
*some refactoring so that `load_xgi_data` and `download_xgi_data` can use the same code snippets

### DIFF
--- a/xgi/readwrite/xgi_data.py
+++ b/xgi/readwrite/xgi_data.py
@@ -1,7 +1,7 @@
 import requests
 import json
 import os
-
+from warnings import warn
 from .. import convert
 from ..exception import XGIError
 
--- a/xgi/readwrite/xgi_data.py
+++ b/xgi/readwrite/xgi_data.py
@@ -5,8 +5,7 @@ import os
 from .. import convert
 from ..exception import XGIError
 
-__all__ = ["load_xgi_data", "download_xgi_data", "_clean_file_path", 
-    "_request_from_xgi_data"]
+__all__ = ["load_xgi_data", "download_xgi_data"]
 
 def load_xgi_data(dataset, path='', read=True, nodetype=None, edgetype=None, 
     max_order=None):
@@ -42,7 +41,7 @@ def load_xgi_data(dataset, path='', read=True, nodetype=None, edgetype=None,
     """
 
     if read:
-        cfp = _clean_file_path(path, dataset)
+        cfp = _make_unix_file_path(path, dataset)
         if os.path.exists(cfp):
             data = json.load(open(cfp, 'r'))
         else:
@@ -70,13 +69,13 @@ def download_xgi_data(dataset, path=''):
     """
 
     jsondata = _request_from_xgi_data(dataset)
-    jsonfile = open(_clean_file_path(path, dataset), 'w')
+    jsonfile = open(_make_unix_file_path(path, dataset), 'w')
     json.dump(jsondata, jsonfile)
     jsonfile.close()
 
 
-def _clean_file_path(path, dataset):
-    """Create a clean path to file to an xgi data set in json format.
+def _make_unix_file_path(path, dataset):
+    """Create a path to file in unix format to an xgi data set in a jason file.
     
     Parameters
     ----------
@@ -90,7 +89,7 @@ def _clean_file_path(path, dataset):
     Returns
     -------
     Filepath
-        The cleaned file path.
+        The unix-format file path.
     """
 
     if len(path)>0:
--- a/xgi/readwrite/xgi_data.py
+++ b/xgi/readwrite/xgi_data.py
@@ -41,7 +41,7 @@ def load_xgi_data(dataset, path='', read=True, nodetype=None, edgetype=None,
     """
 
     if read:
-        cfp = _make_unix_file_path(path, dataset)
+        cfp = os.path.join(path, dataset)
         if os.path.exists(cfp):
             data = json.load(open(cfp, 'r'))
         else:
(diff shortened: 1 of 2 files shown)


xgi/readwrite/xgi_data.py
```diff
@@ -1,20 +1,28 @@
 import requests
+import json
+import os
 
 from .. import convert
 from ..exception import XGIError
 
-__all__ = ["load_xgi_data"]
+__all__ = ["load_xgi_data", "download_xgi_data", "_clean_file_path",
```
maximelucas, line 8: Do we want/need to make the private functions _clean_file_path and _request_from_xgi_data available? If they might be useful somewhere else we could make them non-private or even put them in utilities.py

acuschwarze: I don't think that they are going to be useful anywhere else. I am used to making all functions available by default (because why not). But we can make them unavailable if you prefer it that way?

xgi/readwrite/xgi_data.py
```diff
@@ -32,15 +40,94 @@ def load_xgi_data(dataset, nodetype=None, edgetype=None, max_order=None):
...
+    """
+
+    jsondata = _request_from_xgi_data(dataset)
+    jsonfile = open(_clean_file_path(path, dataset), 'w')
+    json.dump(jsondata, jsonfile)
+    jsonfile.close()
+
+
+def _clean_file_path(path, dataset):
+    """Create a clean path to file to an xgi data set in json format.
```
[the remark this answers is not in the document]
acuschwarze: Yes

xgi/readwrite/xgi_data.py
```diff
@@ -1,20 +1,27 @@
...
     ----------
     dataset : str
         Dataset name. Valid options are the top-level tags of the
         index.json file in the xgi-data repository.
 
+    path : str, optional
+        Path to a local copy of the data set
+    read : bool, optional
+        If read==True, search for a local copy of the data set. Use the local
+        copy instead of the xgi-data repository.
```
maximelucas, line 24:
```suggestion
        copy if it exists, otherwise use the  xgi-data repository.
```

xgi/readwrite/xgi_data.py
```diff
@@ -32,15 +39,94 @@ def load_xgi_data(dataset, nodetype=None, edgetype=None, max_order=None):
     XGIError
        The specified dataset does not exist.
     """
-    index_url = "https://raw.githubusercontent.com/ComplexGroupInteractions/xgi-data/main/index.json"
+
+    if read:
+        cfp = _make_unix_file_path(path, dataset)
+        if os.path.exists(cfp):
+            data = json.load(open(cfp, 'r'))
+        else:
```
maximelucas, line 47: Maybe we could warn the user that the local copy the were looking for does not exist. What do you think?
Maybe a simple print or an actual warning, not sure

xgi/readwrite/xgi_data.py
```diff
@@ -70,13 +69,13 @@ def download_xgi_data(dataset, path=''):
...
-    jsonfile = open(_clean_file_path(path, dataset), 'w')
+    jsonfile = open(_make_unix_file_path(path, dataset), 'w')
     json.dump(jsondata, jsonfile)
     jsonfile.close()
 
 
-def _clean_file_path(path, dataset):
-    """Create a clean path to file to an xgi data set in json format.
+def _make_unix_file_path(path, dataset):
+    """Create a path to file in unix format to an xgi data set in a jason file.
```
maximelucas, line 78: json not jason, although I like it :)

maximelucas:
Nice thanks a lot Alice, looks good!
Couple of details I was wondering about, some of which you already fixed.

xgi/readwrite/xgi_data.py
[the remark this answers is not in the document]
maximelucas: Nice update!

xgi/readwrite/xgi_data.py
maximelucas: I saw you made them unavailable now, let's leave it like this and change later if we need it.

xgi/readwrite/xgi_data.py
acuschwarze: Sure! Do you want that simply printed or as a warning? Could you suggest a commit for this?

xgi/readwrite/xgi_data.py
```diff
@@ -1,20 +1,27 @@
 import requests
+import json
+import os
```
maximelucas, line 4:
```suggestion
from warnings import warn
```

xgi/readwrite/xgi_data.py
maximelucas:
```suggestion
        else:
            warn(f"No local copy was found at {cfp}. The data is requested from the xgi-data repository instead. To download a local copy, use `download_xgi_data`.")
```

xgi/readwrite/xgi_data.py
```diff
@@ -32,15 +39,94 @@ def load_xgi_data(dataset, nodetype=None, edgetype=None, max_order=None):
...
+    Parameters
+    ----------
+    dataset : str
+        Dataset name. Valid options are the top-level tags of the
+        index.json file in the xgi-data repository.
+
+    Returns
+    -------
+    Data
+        The requested data loaded from a json file.
```
maximelucas, line 118:
```suggestion
        The requested data loaded from a json file.
        
   See also
   ---------
   load_xgi_data
```

Could you also add a See also to load_xgi_data with download_xgi_data?

xgi/readwrite/xgi_data.py
```diff
@@ -32,15 +39,94 @@ def load_xgi_data(dataset, nodetype=None, edgetype=None, max_order=None):
...
+        The unix-format file path.
+    """
+
+    if len(path)>0:
+        if path[-1] not in ['/', '\\']:
+            path = path + '/'
+        elif path[-1]=='\\':
+            path = path.replace('\\', '/')
+
+    filepath = path + dataset + '.json'
```
nwlandry, line 101: Can you simplify this by doing something with `os.path`? Like `os.path.join()` or something like that?

nwlandry:
Pending changing "jason" to "json", I would say that this is good to go! I'll add an issue for potentially changing `_make_unix_file_path` to `os.path.join()` in the future.

nwlandry (approved):
Nice work! See comments in the main conversation! You should be able to squash and merge? Let me know if not.

xgi/readwrite/xgi_data.py
acuschwarze: good catch! just did

nwlandry: I think you're missing the ".json" now if I'm not mistaken.

maximelucas (approved):

maximelucas:
Nice, thanks!

PR #257, opened 2022-12-19
Refactor sync

maximelucas:
Split the simulation of the kuramoto and its order parameter into two functions

### DIFF
--- a/tests/dynamics/test_synchronization.py
+++ b/tests/dynamics/test_synchronization.py
@@ -5,32 +5,98 @@ from numpy.linalg import norm
 import xgi
 
 
-@pytest.mark.slow
-def test_compute_kuramoto_order_parameter():
-    H1 = xgi.random_hypergraph(100, [0.05, 0.001], seed=0)
-    r = xgi.compute_kuramoto_order_parameter(
-        H1, 1, 1, np.ones(100), np.linspace(0, 2 * np.pi, 100), 10, 0.002
-    )
+def test_simulate_kuramoto():
+    N = 5
+    n_steps = 10
+    dt = 0.002
+    theta0 = np.linspace(0, 2 * np.pi, N)
+    H1 = xgi.random_hypergraph(N, [0.05, 0.001], seed=0)
+    theta_time, times = xgi.simulate_kuramoto(H1, 1, 1, np.ones(N), theta0, n_steps, dt)
 
-    assert len(r) == 10
-    assert np.all(r >= 0)
+    assert theta_time.shape == (N, n_steps)
+    assert np.all(times == np.arange(n_steps) * dt)
 
     output = np.array(
         [
-            0.01050032,
-            0.01101978,
-            0.01155656,
-            0.01210898,
-            0.01267545,
-            0.01325451,
-            0.01384479,
-            0.01444504,
-            0.01505404,
-            0.01567069,
+            [
+                0.00000000e00,
+                2.00000000e-03,
+                4.00000000e-03,
+                6.00000000e-03,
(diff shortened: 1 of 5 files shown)


xgi/dynamics/synchronization.py
```diff
@@ -63,10 +66,19 @@ def compute_kuramoto_order_parameter(H, k2, k3, w, theta, timesteps=10000, dt=0.
...
 
-    r_time = np.zeros(timesteps)
+    theta_time = np.zeros((n, timesteps))
+    times = np.arange(timesteps) * dt
+
+    if omega is None:
+        omega = np.random.normal(0, 1, n)
+
+    if theta is None:
+        theta = np.random.normal(0, 1, n)
```
nwlandry, line 76: Why not a random uniform on [0, 2pi]?

xgi/generators/nonuniform.py
```diff
@@ -290,8 +290,8 @@ def random_hypergraph(N, ps, seed=None):
     if (np.any(np.array(ps) < 0)) or (np.any(np.array(ps) > 1)):
         raise ValueError("All elements of ps must be between 0 and 1 included.")
 
-    nodes = range(N)  # list(G.nodes())
-    hyperedges = []  # hyperedges = list(G.edges())
+    nodes = range(N)
```
nwlandry, line 293: Good catch.

xgi/dynamics/synchronization.py
```diff
@@ -26,19 +27,21 @@ def compute_kuramoto_order_parameter(H, k2, k3, w, theta, timesteps=10000, dt=0.
         The coupling strength for links
     k3 : float
         The coupling strength for triangles
-    w : numpy array of real values
-        The natural frequency of the nodes.
+    omega : numpy array of real values
+        The natural frequency of the nodes. If None (default), randomly drawn from a normal distribution
     theta : numpy array of real values
-        The initial phase distribution of nodes.
+        The initial phase distribution of nodes. If None (default), randomly drawn from a normal distribution
```
nwlandry, line 33: See comment below.

nwlandry (approved):
This is a great addition. Thanks for doing this!

tests/dynamics/test_synchronization.py
```diff
@@ -5,32 +5,98 @@
...
+    theta0 = np.linspace(0, 2 * np.pi, N)
+    H1 = xgi.random_hypergraph(N, [0.05, 0.001], seed=0)
+    theta_time, times = xgi.simulate_kuramoto(H1, 1, 1, np.ones(N), theta0, n_steps, dt)
 
-    assert len(r) == 10
-    assert np.all(r >= 0)
+    assert theta_time.shape == (N, n_steps)
+    assert np.all(times == np.arange(n_steps) * dt)
 
     output = np.array(
```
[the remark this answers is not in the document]
maximelucas: Good point, will do.

xgi/dynamics/synchronization.py
maximelucas: Good point. Alexis used this in his PR so I added here too for consistency. Also thought about the uniform one, but perfectly uniform might be "dangerous" because with identical frequencies it is sometimes a stable state, so you won't see any convergence. Could do a uniform on [0, 2pi] + some small noise. Or leave it like this and let the user pick. What do you think?

maximelucas: Updated to what you said, more natural imo

xgi/dynamics/synchronization.py
maximelucas: Done

PR #263, opened 2023-01-03
Flag2d

maximelucas:
- added `find_triangles()` function
- added `flag_complex_d2()` much faster than `flag_complex()`, while it is being made faster (#264)
- refactored `random_flag_complex_d2()` to use `flag_complex_d2()`
- added tests for all the above

### DIFF
--- a/tests/generators/test_classic.py
+++ b/tests/generators/test_classic.py
@@ -63,6 +63,15 @@ def test_flag_complex():
     assert S4.num_edges == S5.num_edges
     assert set(S4.edges.members()) == set(S5.edges.members())
 
+def test_flag_complex_d2():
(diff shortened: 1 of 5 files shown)


nwlandry:
Looks great! The only question I have is what the difference is between `nx.triangles` and `xgi.find_triangles`.

maximelucas:
Great question! `nx.triangles` counts, `xgi.find_triangles` lists.

`nx.triangles` returns a dict with number of triangles keyed by node label. No way to know what the triangles are.
`xgi.find_triangles` returns a list of all triangles as 3-node sets

nwlandry (approved):
Looks great!

PR #268, opened 2023-01-10
Fix networkx issue

nwlandry:
Made XGI compatible with NetworkX 3.0 and changed use of scipy sparse matrices to scipy sparse arrays.


maximelucas:
Nice! Thanks Nich.
I remember we used to have a warning because of the csr_matrix being obsolete, and I think we did something somewhere to not see it anymore. We could remove it now. Anyone remembers where?

PR #274, opened 2023-01-21
Fixed the centralities so that they are positive and 1-normalized

### LINKED ISSUE #273: cec_centrality output has random sign
We should enforce it to be positive. Maybe other centralities have a similar behaviour?

nwlandry:
Fixes #273.


leotrs (commented):
Looking good, thanks. Just one thought, we should probably document the fact that the centralities are being normalized with the 1-norm.

nwlandry:
Fixed!

leotrs (approved):

Issue #276, opened 2023-02-09
nodes named after different numbers in adjacency matrix and .draw


cperezln:
i have this code
`h = xgi.Hypergraph([[1, 2, 3, 7], [4], [5, 6, 7]])
xpl.draw(h, node_labels = True)
a, r = mx.adjacency_matrix(h, index = True)
df = pd.DataFrame(a.toarray())`
and the output of it, in images from xpl draw (being xpl from xgi.drawing import xgi_pylab as xpl) and the dataframe are
![imagen](https://user-images.githubusercontent.com/69525230/217884579-b336800c-936c-4ce0-ab95-9c0d5503f693.png)
![imagen](https://user-images.githubusercontent.com/69525230/217884635-525aac45-a433-44ae-8464-355b7a6fad5d.png)
as you can see, the most connected node in the plot is 7, but in the dataframe its not. also, they seem to be unordered. why is this?

maximelucas:
Hi @cperezln, thanks for you question. 
This is expected behaviour, two things are happening here.
1. nodes are added to the hypergraphs in the order in which they are in your input edges. This means that 7 is added before 4 in your case. 
2. the adjacency matrix being a numpy array, it contains no information about which columns/rows correspond to which node labels. So when you convert it to a pandas Dataframe, by default the index will from 0 to N-1, regardless of the node labels. That is why there is an option to return the mapping between matrix index and node label with `index=True`, which you used. If you can tell pandas to use the node labels with:
 `df = pd.DataFrame(a.toarray(), index=r.values(), columns=r.values())`
then you get:
    <img width="163" alt="Screenshot 2023-02-09 at 22 41 16" src="https://user-images.githubusercontent.com/7493360/217945691-317b4e8b-f4b9-412d-b866-5a779ef02e9b.png">
    from which you can see that node 7 is indeed the most connected, as expected.

To control the default order from point 1., you can always add nodes before the edges with 
```python
H = xgi.Hypergraph()
H.add_nodes_from(range(1,8)) 
H = xgi.add_edges_from([[1, 2, 3, 7], [4], [5, 6, 7]])
```
This way the nodes will be ordered in ascending order. Converting the adjacency matrix to a Dataframe without specifying the index will still go from 0-6 instead of 1-7 though.

One more thing: all the functions you used can be called directly from `xgi.`: `xgi.draw()` and `xgi.adjacency_matrix()`.

I hope this answers your questions.

cperezln:
[...]

PR #277, opened 2023-02-13
added sort_ids option

nwlandry:
This addresses #276 if we feel that it is necessary. The only thing is that it is only implemented for the adjacency matrix and the incidence matrix, but if we decide that this is important, we should probably apply it to all of the matrices.


nwlandry:
I agree with you. How about this: let's not fix this but describe ways to permute rows/columns in numpy/scipy/pandas in the header of matrix.py so that there's documentation on this? I will close this and move the discussion to the issue.

PR #281, opened 2023-02-17
Add tox

arnaudon:
This is a first attempt to use tox in this repo. 
We can fine-tune the settings for lint, etc... and improve the code as well so all checks pass.

Usage is simple, from tox.ini, and various sections, we can do locally:
`tox -e lint`
`tox -e format`
`tox -e py3`
for lint check, formating the code or running tests.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -18,7 +18,7 @@ jobs:
       - name: Install packages
         run: |
           python -m pip install --upgrade pip wheel setuptools
-          pip install -r requirements/default.txt -r requirements/test.txt
+          pip install -r requirements/default.txt -r requirements/test.txt -r requirements/tutorial.txt
           pip install .
           python -m pip list
       - name: Run tests and collect coverage
--- a/.github/workflows/run-tox.yml
+++ b/.github/workflows/run-tox.yml
@@ -4,7 +4,7 @@ on:
   pull_request:
   push:
     branches:
-      - master
+      - main
 
 jobs:
   build:
--- a/.github/workflows/run-tox.yml
+++ b/.github/workflows/run-tox.yml
@@ -9,9 +9,10 @@ on:
 jobs:
   build:
 
-    runs-on: ubuntu-latest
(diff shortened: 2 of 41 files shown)


nwlandry:
Thanks so much for this contribution!! I am currently reading up on tox to make sure I understand what it does, then I will review this! Looks very exciting! :)

arnaudon:
You're welcome! So we need to setup the lining rules, or update the code (or both), and transfer tutorial tests into tox. And play with parameters until it's as you want.

arnaudon:
@nwlandry we could use that for notebooks: https://testbook.readthedocs.io/en/latest/index.html

arnaudon:
So I added the notebook test in a test file, and noticed one crahsed (last dynamics on kuramoto cell of the quickstart), so I'm not sure how well this was tested via github action. Anyway, not with testbook, one can even test the results of each cells, etc...

arnaudon:
A couple of things:
- I fixed pylint with minimalistic changes, but some things could be improved in later PR
- the original test don't seem to like the test notebook I added, if you want to keep these tests, I can try to see why it fails and fix it

nwlandry:
Okay, I finally had a chance to look at this. Overall, it looks good, but I think we should attempt to reproduce the testing that is already there and add the notebook testing to integrate tox, and then get more ambitious once we have something working. Thanks for all your work!

arnaudon:
@nwlandry this is now good to go I think, it should have the same tests as before. The massive increase of coverage is because the notebooks are now part of the coverage report. We can remove the 'tutorial' manual tests here, or in a later PR, as you prefer (we are testing them 3 times now haha)

arnaudon:
So I didn't include coverage in tox, as it does not work on windows, and I didn't look how to run it only on ubuntu (this is not a big deal atm). 
Also, it crashes the notebook tests randomly ( and pass if we retrigger) I think due to the fact we run the several times. Probably we need to remove some notebook tests already on this PR if its ok with you.

nwlandry:
I think tox is a good idea, but I think that starting small and working up to a full integration is best. Closing this for now.

PR #284, opened 2023-03-03
Add sparse argument to matrix functions

### LINKED ISSUE #269: control sparsity of xgi.adjacency_matrix
> Side problem that we might tackle at the same time: there's an argument `sparse=True` in `xgi.incidence_matrix` that gives control over the type of the output. Some other functions, or at least `xgi.adjacency_matrix`, does not have that argument but uses `xgi.incidence_matrix` internally. First, I think we could add that argument. Second, adjacency matrix should then all be output as sparse, but for some reason I think I remember I had cases when they were not, and I had no control over it.

_Originally posted by @maximelucas in https://github.com/ComplexGroupInteractions/xgi/issues/267#issuecomment-1376035743_

nwlandry:
Added the `sparse` argument to all but the `hodge_laplacian` and `boundary_matrix` functions. Fixes #269.

### DIFF
--- a/xgi/linalg/matrix.py
+++ b/xgi/linalg/matrix.py
@@ -179,7 +179,7 @@ def intersection_profile(H, order=None, sparse=True, index=False):
     return (P, coldict) if index else P
 
 
-def degree_matrix(H, order=None, sparse=False, index=False):
+def degree_matrix(H, order=None, index=False):
     """Returns the degree of each node as an array
 
     Parameters
@@ -189,8 +189,6 @@ def degree_matrix(H, order=None, sparse=False, index=False):
     order: int, optional
         Order of interactions to use. If None (default), all orders are used. If int,
         must be >= 1.
-    sparse: bool, default: False
-        Specifies whether the output matrix is a scipy sparse matrix or a numpy matrix
     index: bool, default: False
         Specifies whether to output disctionaries mapping the node and edge IDs to indices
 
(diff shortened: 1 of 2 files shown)


xgi/linalg/matrix.py
```diff
@@ -314,13 +321,12 @@ def multiorder_laplacian(H, orders, weights, rescale_per_node=False, index=False
             L_multi += L * w / np.mean(K)
 
     if index:
-        _, rowdict, _ = incidence_matrix(H, index=True)
-        return L_multi, rowdict
+        return L_multi, dict(zip(range(H.num_nodes), H.nodes))
```
maximelucas, line 324: Why not use index=True anymore? This won't work with nodes that are not ordered?

maximelucas:
Great! A few tests to check the sparseness would be safer?

xgi/linalg/matrix.py
nwlandry: I essentially copied the lines in the `incidence_matrix` function generating the `nodedict` and `edgedict` dictionaries because most of the incidence matrix function was unnecessary here. But not sure if that was your actual question...

xgi/linalg/matrix.py
```diff
@@ -244,11 +249,11 @@ def laplacian(H, order=1, rescale_per_node=False, index=False):
...
     """
-    A, row_dict = adjacency_matrix(H, order=order, weighted=True, index=True)
+    A, row_dict = adjacency_matrix(H, order=order, sparse=sparse, weighted=True, index=True)
     if A.shape == (0,):
         return (np.array([]), {}) if index else np.array([])
 
-    K = degree_matrix(H, order=order, index=False)
+    K = degree_matrix(H, order=order, sparse=sparse, index=False)
 
     L = order * np.diag(K) - A  # ravel needed to convert sparse matrix
```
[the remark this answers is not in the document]
nwlandry: Good point. I'll add a test.

xgi/linalg/matrix.py
maximelucas: Oh I see good call then

nwlandry:
I think I addressed all your comments. I had to make `degree_matrix()` return a square matrix in order to add the sparse keyword.

nwlandry:
The `clique_motif_matrix()` function now directly calls the `adjacency_matrix()` function.

maximelucas:
Great work, thanks Nich. 

In future, are we happy with a diagonal degree matrix like it is now?  For now it's fine, but conceptually I liked it better before I think.

maximelucas (approved):

nwlandry:
I trust your judgement so I will change it back to how it was before and then commit. :)

maximelucas:
Don't do if it breaks what you did for the sparseness :)

nwlandry:
Okay, I changed it back to a vector! There is no longer a "sparse" argument for the `degree_matrix()` function but it doesn't matter anyway because it is a diagonal matrix.

PR #286, opened 2023-03-06
added uniform generators

nwlandry:
Added
* hypergraph SBM method
* hypergraph planted partition method
* hypergraph Erdős–Rényi model

### DIFF
--- a/xgi/generators/uniform.py
+++ b/xgi/generators/uniform.py
@@ -257,7 +257,7 @@ def uniform_hypergraph_planted_partition_model(n, m, rho, k, epsilon):
 
 
 def uniform_erdos_renyi_hypergraph(n, m, k):
-    """Generate an m-uniform Erdos-Renyi hypergraph
+    """Generate an m-uniform Erdős–Rényi hypergraph
 
     Parameters
     ----------
--- a/xgi/generators/uniform.py
+++ b/xgi/generators/uniform.py
@@ -12,8 +12,8 @@ from .classic import empty_hypergraph
 
 __all__ = [
     "uniform_hypergraph_configuration_model",
-    "uniform_SBM_hypergraph",
-    "uniform_hypergraph_planted_partition_model",
+    "uniform_HSBM",
+    "uniform_HPPM",
     "uniform_erdos_renyi_hypergraph",
 ]
 
@@ -102,8 +102,8 @@ def uniform_hypergraph_configuration_model(k, m, seed=None):
     return H
 
 
-def uniform_SBM_hypergraph(n, m, p, sizes, seed=None):
-    """Create a uniform SBM hypergraph.
+def uniform_HSBM(n, m, p, sizes, seed=None):
+    """Create a uniform hypergraph stochastic block model (HSBM).
 
     Parameters
     ----------
@@ -133,10 +133,10 @@ def uniform_SBM_hypergraph(n, m, p, sizes, seed=None):
         - If the sum of the vector of sizes does not equal the number of nodes.
     Exception
         If there is an integer overflow error
-    
+
     See Also
     --------
-    uniform_hypergraph_planted_partition_model
+    uniform_HPPM
 
     References
     ----------
@@ -159,7 +159,7 @@ def uniform_SBM_hypergraph(n, m, p, sizes, seed=None):
         raise XGIError("Sum of sizes does not match n")
 
     if seed is not None:
-        random.seed(seed)
+        np.random.seed(seed)
 
     node_labels = range(n)
     H = empty_hypergraph()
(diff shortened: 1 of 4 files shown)


xgi/generators/uniform.py
```diff
@@ -89,3 +100,253 @@ def uniform_hypergraph_configuration_model(k, m, seed=None):
...
+
+    p = p_out * np.ones([2] * m)
+    p[tuple([0] * m)] = p_in
+    p[tuple([1] * m)] = p_in
+
+    return uniform_SBM_hypergraph(n, m, p, sizes)
+
+
+def uniform_erdos_renyi_hypergraph(n, m, k):
+    """Generate an m-uniform Erdős–Rényi hypergraph
```
maximelucas, line 260: This is different from our non-uniform version of an ER hypergraph which takes `p` rather than `k` as an argument. Do we want to make it consistent? Or offer both options?

```suggestion
    """Generate an m-uniform Erdős–Rényi hypergraph.
    
    This creates an hypergraph with `n` nodes where hyperedges of size `m` created at random to obtain a mean degree of `k`. 
```

xgi/generators/uniform.py
```diff
@@ -89,3 +100,253 @@ def uniform_hypergraph_configuration_model(k, m, seed=None):
...
+    
+    See Also
+    --------
+    random_hypergraph
+    """
+    node_labels = range(n)
+    H = empty_hypergraph()
+    H.add_nodes_from(node_labels)
+
+    p = k / (m * n ** (m - 1))
```
maximelucas, line 285:
```suggestion
    p = k / (m * n ** (m - 1)) # wiring probability
```

xgi/generators/uniform.py
```diff
@@ -89,3 +100,253 @@ def uniform_hypergraph_configuration_model(k, m, seed=None):
...
+            while index < max_index:
+                indices = _index_to_edge_partition(index, partition_sizes, m)
+                e = [partition[block[i]][indices[i]] for i in range(m)]
+                if len(e) == m:
+                    H.add_edge(e)
+                index += np.random.geometric(p[block])
+    return H
+
+
+def uniform_hypergraph_planted_partition_model(n, m, rho, k, epsilon):
```
maximelucas, line 197: Naming not consistent with the SBM function above. Did you not use the acronym because less known than SBM?

xgi/generators/uniform.py
```diff
@@ -89,3 +100,253 @@ def uniform_hypergraph_configuration_model(k, m, seed=None):
             del stubs[index]
 
     return H
+
+
+def uniform_SBM_hypergraph(n, m, p, sizes, seed=None):
+    """Create a uniform SBM hypergraph.
```
maximelucas, line 106: spell SBM out

xgi/generators/uniform.py
```diff
@@ -89,3 +100,253 @@ def uniform_hypergraph_configuration_model(k, m, seed=None):
...
+    n : int > 0
+        Number of nodes
+    m : int > 0
+        Hyperedge size
+    rho : float between 0 and 1
+        The fraction of nodes in community 1
+    k : float > 0
+        Mean degree
+    epsilon : float > 0
+        Imbalance parameter
```
maximelucas, line 211: add a seed?

xgi/generators/uniform.py
```diff
@@ -89,3 +100,253 @@ def uniform_hypergraph_configuration_model(k, m, seed=None):
...
+    p[tuple([0] * m)] = p_in
+    p[tuple([1] * m)] = p_in
+
+    return uniform_SBM_hypergraph(n, m, p, sizes)
+
+
+def uniform_erdos_renyi_hypergraph(n, m, k):
+    """Generate an m-uniform Erdős–Rényi hypergraph
+
+    Parameters
```
maximelucas, line 262: missing a seed

maximelucas:
Looks nice, thanks Nich! I was thinking of adding a uniform ER function for a long time myself. 

Maybe add a basic tests and a black pass?
Would also be better if someone else could have another look at the algos themselves.

xgi/generators/uniform.py
nwlandry: good point. I will write as an acronym and explain in the docstring.

nwlandry:
All your comments should be addressed I believe.

maximelucas:
Nice!!

maximelucas (approved):

maximelucas:
(I think in the future we might want to make the naming of all our generators more consistent. Older generators did not use acronyms. Some use "hypergraph" in front, others at the end. Do we need to mention "hypergraph" in all of their names?)

nwlandry:
I agree! I will commit this and add your suggestion as an issue.

PR #288, opened 2023-03-09
Up-versioned requirements

### LINKED ISSUE #287: Cannot install xgi
Dear contributors, 

First of all, thank you so much for the hard work putting all these together. Amazing!
I tried to redo the tutorials, however, I found I cannot install the xgi package at all. How frustrating. 
After I pip installed the xgi package, if I go to the notebook and tried to import it, I got this error message that avoid me having the package imported: 
"ImportError: cannot import name 'csr_array' from 'scipy.sparse' (/opt/anaconda3/lib/python3.9/site-packages/scipy/sparse/__init__.py)"
It seems like something's going on with scipy when importing xgi. 
I really have no idea how to deal with this problem. 

I hope my question would be helpful for your debugging process and looking forward to hearing the solution!

Best wishes, 
Sally

nwlandry:
Now matches NetworkX requirements. Fixes #287.

### DIFF
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -1,6 +1,6 @@
-numpy>=1.19
-scipy>=1.5
-pandas>=1.0
+numpy>=1.20
+scipy>=1.8
+pandas>=1.3
 networkx>=2.7
 requests>=2.0
-matplotlib>=3.3
+matplotlib>=3.4


maximelucas:
Thanks Nich. Are those all needed for functionalities we use?

nwlandry:
* scipy is per the issue brought up here
* numpy is because of https://numpy.org/neps/nep-0029-deprecation_policy.html
* the pandas and matplotlib requirements are there because networkx >=2.7 (which we require) require these versions anyway

PR #289, opened 2023-03-09
Fix `add_edge()` bug

nwlandry:
This fixes an error where
```python
import xgi
H = xgi.Hypergraph()
H.add_edge({1, 2, 3}, id=0}
```
returns an empty hypergraph.


maximelucas:
Good catch!

maximelucas (approved):

PR #290, opened 2023-03-11
Line graph and vector centrality functions added

goznalo-git:
Hi there, 

I really like this library, and I've been using it for a while now. I would like to have these two additions included, and I'm sure others would benefit from them.
- The function `convert_to_line_graph()` takes a Hypergraph and constructs its associated line graph (a pairwise graph whose nodes correspond to the edges of the Hypergraph, and two of those nodes are linked together if their associated edges share at least one node).
- The function `vector_centrality()`, defined in https://doi.org/10.1016/j.chaos.2022.112397, which allows one to extract a vector-like eigenvector centrality of a Hypergraph by using the abovementioned connection to its line graph. 

I updated the changelog and the docs, as well as making sure the formatting was fine with isort and black.

best,
Gonzalo

### DIFF
--- a/docs/source/api/algorithms/xgi.algorithms.centrality.rst
+++ b/docs/source/api/algorithms/xgi.algorithms.centrality.rst
@@ -11,4 +11,4 @@ xgi.algorithms.centrality
    .. autofunction:: HEC_centrality
    .. autofunction:: ZEC_centrality
    .. autofunction:: node_edge_centrality
-   .. autofunction:: vector_centrality
+   .. autofunction:: line_vector_centrality
--- a/docs/source/api/algorithms/xgi.algorithms.centrality.rst
+++ b/docs/source/api/algorithms/xgi.algorithms.centrality.rst
@@ -11,3 +11,4 @@ xgi.algorithms.centrality
    .. autofunction:: HEC_centrality
    .. autofunction:: ZEC_centrality
    .. autofunction:: node_edge_centrality
+   .. autofunction:: vector_centrality
--- a/xgi/algorithms/centrality.py
+++ b/xgi/algorithms/centrality.py
@@ -273,7 +273,7 @@ def line_vector_centrality(H):
     """
 
     if not xgi.is_connected(H):
-        raise XGIError("This method is not defined non-connected hypergraphs.")
+        raise XGIError("This method is not defined for non-connected hypergraphs.")
 
     LG = convert_to_line_graph(H)
     LGcent = nx.eigenvector_centrality(LG)
--- a/xgi/algorithms/centrality.py
+++ b/xgi/algorithms/centrality.py
@@ -285,7 +285,7 @@ def line_vector_centrality(H):
     }
     hyperedge_dims = {tuple(edge): len(edge) for edge in H.edges.members()}
 
-    D = np.max(list(hyperedge_dims.values()))
+    D = H.edges.size.max()
 
     for k in range(2, D + 1):
         c_i = np.zeros(len(H.nodes))
--- a/xgi/algorithms/centrality.py
+++ b/xgi/algorithms/centrality.py
@@ -253,9 +253,7 @@ def node_edge_centrality(
 
 
 def vector_centrality(H):
-    """
-    Compute the vector centrality of nodes in hypergraphs
-
+    """The vector centrality of nodes in the line graph of the hypergraph.
     Parameters
     ----------
     H : Hypergraph
--- a/xgi/algorithms/centrality.py
+++ b/xgi/algorithms/centrality.py
@@ -252,7 +252,7 @@ def node_edge_centrality(
     }
 
 
-def vector_centrality(H):
+def line_vector_centrality(H):
     """The vector centrality of nodes in the line graph of the hypergraph.
     Parameters
     ----------
--- a/xgi/algorithms/centrality.py
+++ b/xgi/algorithms/centrality.py
@@ -6,6 +6,7 @@ from numpy.linalg import norm
 from scipy.sparse.linalg import eigsh
 
 from ..classes import convert_labels_to_integers, is_uniform
+from ..convert import convert_to_line_graph
 from ..exception import XGIError
 from ..linalg import clique_motif_matrix, incidence_matrix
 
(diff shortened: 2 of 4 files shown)


docs/source/api/algorithms/xgi.algorithms.centrality.rst
```diff
@@ -10,4 +10,5 @@ xgi.algorithms.centrality
    .. autofunction:: CEC_centrality
    .. autofunction:: HEC_centrality
    .. autofunction:: ZEC_centrality
-   .. autofunction:: node_edge_centrality
\ No newline at end of file
+   .. autofunction:: node_edge_centrality
+   .. autofunction:: vector_centrality
```
leotrs, line 14: I know this is probably the canonical name in the reference paper, but it might be too general for a general purpose library like XGI. Perhaps we should consider `line_vector_centrality` to make allusion to the fact that this uses the line graph.

xgi/convert.py
```diff
@@ -134,6 +134,36 @@ def convert_to_graph(H):
     return G
 
 
+def convert_to_line_graph(H):
+    """Line graph of the hypergraph H. This is the graph
+    whose nodes correspond to each hyperedge in H, linked
+    together if they share at least one vertex.
```
leotrs, line 140: Tiny suggestion:
```suggestion
    """Line graph of the hypergraph.
    
    The line graph of the hypergraph `H` is the graph whose
    nodes correspond to each hyperedge in `H`, linked together
    if they share at least one vertex.
```

xgi/algorithms/centrality.py
```diff
@@ -249,3 +250,61 @@ def node_edge_centrality(
     return {node_dict[n]: new_x[n] for n in node_dict}, {
         edge_dict[e]: new_y[e] for e in edge_dict
     }
+
+
+def vector_centrality(H):
+    """
+    Compute the vector centrality of nodes in hypergraphs
+
```
leotrs, line 258: Tiny suggestion, for uniformity with other docstrings elsewhere
```suggestion
    """The vector centrality of nodes in the line graph of the hypergraph.
```

xgi/algorithms/centrality.py
```diff
@@ -249,3 +250,61 @@ def node_edge_centrality(
...
+    LGcent = nx.eigenvector_centrality(LG)
+
+    vc = {node: [] for node in H.nodes}
+
+    edge_label_dict = {
+        tuple(edge): index for index, edge in enumerate(H.edges.members())
+    }
+    hyperedge_dims = {tuple(edge): len(edge) for edge in H.edges.members()}
+
+    D = np.max(list(hyperedge_dims.values()))
```
leotrs, line 290: Suggestion (might be slower?):
```suggestion
    D = H.edges.size.max()
```

xgi/algorithms/centrality.py
```diff
@@ -249,3 +250,61 @@ def node_edge_centrality(
...
+
+    for k in range(2, D + 1):
```
leotrs, line 292: This loop is first iterating over each possible edge order, then filtering to get all edges of said order, and the iterating over each node of each edge of those edges. So it's actually a four-ly nested loop. I wonder if there is a way to iterate over each edge (and each node in each edge) once?

xgi/algorithms/centrality.py
```diff
@@ -249,3 +250,61 @@ def node_edge_centrality(
...
+        c_i = np.zeros(len(H.nodes))
+
+        for edge, _ in list(filter(lambda x: x[1] == k, hyperedge_dims.items())):
+
+            for node in edge:
+                try:
+                    c_i[node] += LGcent[edge_label_dict[edge]]
+                except IndexError:
+                    raise Exception(
+                        "Nodes must be written with the Pythonic indexing (0,1,2...)"
```
leotrs, line 302: It's not super clear to me what is mean there by Pythonic indexing. Does this mean that the algorithm does not work for nodes that are labeled with strings? Does it work with nodes that are labeled with non-consecutive integers, or integers not starting at zero?

leotrs (commented):
Dear Gonzalo, thank you for your contribution! The entire purpose of XGI from its inception was to be a tool for the community by the community, so your involvement is more than welcome. I have reviewed the code and it looks like exactly the kind of addition we are always open to receiving. I have left some comments that can hopefully be solved in a few quick iterations and we'll be sure to merge this after that.

Perhaps another pair of eyes would be useful too @nwlandry @maximelucas.

After all the comments are resolved, it would also be nice to have one or two reference tests for the new features.

Thanks again!

maximelucas:
Thanks a lot @goznalo-git ! This looks like a quality PR.
I'll have a look this week.

xgi/convert.py
```diff
@@ -134,6 +134,38 @@ def convert_to_graph(H):
...
+    -------
+    LG : networkx.Graph
+         The line graph associated to the Hypergraph
+    """
+
+    LG = nx.Graph()
+
+    edge_label_dict = {
+        tuple(edge): index for index, edge in enumerate(H.edges.members())
+    }
```
maximelucas, line 159: Current code will give new labels to the hyperedges from 0 to M. This suggestion keeps the hyperedges labels, which might be unordered, and strings etc. Check it does not mess up the rest of your code.
```suggestion
    edge_label_dict = {tuple(edge): index for index, edge in H._edge}
```

xgi/convert.py
```diff
@@ -134,6 +134,38 @@ def convert_to_graph(H):
...
+    LG.add_nodes_from(edge_label_dict.values())
```
maximelucas, line 160: Current code is correct but will add nodes in their order of appearances in hyperedges. The following adds them in an ordered way:
```suggestion
    nodes = sorted(set(edge_label_dict.values()))
    LG.add_nodes_from(nodes)
```

xgi/convert.py
```diff
@@ -134,6 +134,38 @@ def convert_to_graph(H):
...
+
+    for edge1, edge2 in combinations(H.edges.members(), 2):
+        if edge1.intersection(edge2):
+            LG.add_edge(edge_label_dict[tuple(edge1)], edge_label_dict[tuple(edge2)])
```
maximelucas, line 164: Your code definitely works. I'm wondering if it might be faster using `H.edges.neighbors()`. I give it a try below, feel free to discard.
```suggestion
    edges = []
    for edge_id1 in H.edges():
        for edge_id2 in H.edges.neighbors(edge_id1):
            edges.append({edge_id1, edge_id2})
    LG.add_edges_from(edges)
```

xgi/algorithms/centrality.py
```diff
@@ -249,3 +250,59 @@ def node_edge_centrality(
...
+        raise XGIError("This method is not defined non-connected hypergraphs.")
+
+    LG = convert_to_line_graph(H)
+    LGcent = nx.eigenvector_centrality(LG)
+
+    vc = {node: [] for node in H.nodes}
+
+    edge_label_dict = {
+        tuple(edge): index for index, edge in enumerate(H.edges.members())
+    }
```
maximelucas, line 285: Do you need labeling from 0 to M? If not see below for suggestion to keep hyperedges labels

xgi/algorithms/centrality.py
```diff
@@ -249,3 +250,59 @@ def node_edge_centrality(
...
+    hyperedge_dims = {tuple(edge): len(edge) for edge in H.edges.members()}
+
+    D = H.edges.size.max()
+
+    for k in range(2, D + 1):
```
maximelucas, line 290: Shouldn't this go up to `D` since you're iterating over sizes?

xgi/algorithms/centrality.py
```diff
@@ -249,3 +250,59 @@ def node_edge_centrality(
...
+        c_i = np.zeros(len(H.nodes))
+
+        for edge, _ in list(filter(lambda x: x[1] == k, hyperedge_dims.items())):
```
maximelucas, line 293: Should also work with (check if not slower). Then you don't need to create `hyperedge_dims` anymore 
```suggestion
        for edge in H.edges.filterby("size", k).members():
```

maximelucas:
Thank you very much @goznalo-git. I added some suggestions to make use of some xgi features.
Thanks a lot!

goznalo-git:
Hi @leotrs @maximelucas 

Thanks for your feedback! There were some interesting suggestions and improvements, indeed. Let me know what you think of the actual status. 

best!

maximelucas:
Thank you @goznalo-git. I'll approve this for merging now, and raise some issues for our remaining comments so we can solve later together with adding tests.

maximelucas (approved):

leotrs:
@goznalo-git, thank you for your contribution. You are more than welcome to continue contributing anything you'd find useful, and to come chat with us whenever on our [zulip server](https://xgi.zulipchat.com/join/o3sjqnigvvpxma2fzv6g2m52/).

PR #293, opened 2023-03-14
tests: added basic for layouts

maximelucas:
- added initial tests for layout functions
- added seed argument for layout functions that did not have it


nwlandry (approved):
Looks great! Thanks so much! Having the seed arg will make life SO much easier.

PR #294, opened 2023-03-14
Fixed quickstart notebook

nwlandry:
Updated the quickstart notebook and updated the kuramoto function to output theta as a T x N matrix instead of an N x T matrix so that you don't have to transpose the output before plotting.


maximelucas (approved):

maximelucas:
Thanks Nich!

PR #297, opened 2023-03-17
DiHypergraphs

nwlandry:
This is a preliminary attempt at addressing #2. This requires edges to be of the form `{"head": {}, "tail": {}}` - happy to implement an alternative. Currently the following are implemented:
* Number of nodes and edges
* adding/removing a single node or iterable of nodes
* adding/removing a single edge or an iterable of edges

The following is not implemented (not exhaustive):
* NodeViews/EdgeView and by extension any stats
* Directly converting to a dihypergraph in the constructor
* Algorithms

I plan on adding tests, but not until the data structure/syntax is agreed upon.


requirements/developer.txt
```diff
@@ -1,4 +1,3 @@
 black[jupyter]==22.3.0
-pre-commit>=2.12
```
leotrs, line 2: What happened to this?

xgi/classes/dihypergraph.py
```diff
@@ -0,0 +1,610 @@
+"""Base class for undirected hypergraphs."""
```
leotrs, line 1:
```suggestion
"""Base class for directed hypergraphs."""
```

xgi/classes/dihypergraph.py
```diff
@@ -0,0 +1,610 @@
...
+
+from ..exception import XGIError
+from ..utils import IDDict, update_uid_counter
+from .reportviews import EdgeView, NodeView
+
+__all__ = ["DiHypergraph"]
+
+
+class DiHypergraph:
+    r"""A hypergraph is a collection of subsets of a set of *nodes* or *vertices*.
```
leotrs, line 16: Not sure you want comments on the docstring at this time, but this seems to be copy/pasted from the Hypergraph class, no?

xgi/classes/dihypergraph.py
```diff
@@ -0,0 +1,610 @@
...
+            self._node_attr[node] = self._node_attr_dict_factory()
+        self._node_attr[node].update(attr)
+
+    def add_nodes_from(self, nodes_for_adding, **attr):
+        """Add multiple nodes with optional attributes.
+
+        Parameters
+        ----------
+        nodes_for_adding : iterable
+            An iterable of nodes (list, dict, set, etc.).
```
leotrs, line 165: Personally, I don't like that we have to specify what an iterable is. The parenthesis also says that a dict is possible, but it's not clear what the keys and values o such a dict are. We can leave this for a different issue tho.

leotrs (commented):
Thanks for the progress. I haven't read everything in detail, but I'd suggest adding extensive tests before merging this.

Also, was there a discussion I missed about making DiHypergraph a completely different class from Hypergraph rather than using inheritance?

nwlandry:
Hi @leotrs - thanks for all the comments. Sounds good about cleaning up the documentation and adding tests. I picture the integration of directed hypergraphs into the library to be in 4-5 stages: 1) create a class which allows adding and removing of nodes and edges (this PR), 2) add directed node and edge stats, 3) edit the convert.py file so that DiHypergraphs can be initialized directly, 4) start adding compatibility for DiHypergraphs in some of the existing algorithms (connectedness, etc.), and 5) figure out inheritance.

For now, it seems easiest to keep the two classes separate, because DiHypergraphs are still very experimental, but once it's a bit more proven, ideally, we could leverage some of the same class structure.

So for this PR, I will only address step 1) and continue chipping away at this list.

leotrs:
>  5) figure out inheritance.

I guess I'm of two minds here. Are you saying that you are trying to merge something experimental and that is not ready for the end user? If so, there is nothing about that in this PR, and that should be mentioned up front. If this is the case then yeah I agree that inheritance is perhaps best considered after some skeleton is put in place. Whether that experimental skeleton should be merged to the main branch is another topic. I would much rather have a feature branch  where we experiment ad nauseam rather than have the main branch contain experimental stuff. For example, one side effect would be that either 1) we don't package/release anything until the experimental features are considered ready, or 2) we package/release experimental features with our next few releases. I'm against 2), and 1) seems like it wouldn't really mix well with our release cadence (though this depends on how much work is put on this and ensuing PRs).

On the other hand, if you are not really considering this as experimental and want the end-user to use DiHypergraphs as soon as this PR is merged, then I'd strongly suggest to consider inheritance much closer to the beginning of this whole process. In particular, I would not even touch stats before figuring out a more stable solution for inheritance.

> For now, it seems easiest to keep the two classes separate

FWIW, I agree that for the time being the two classes can remain separate until we have a better idea of how much code can be shared among them.

requirements/developer.txt
nwlandry: It's not used for anything currently, so I got rid of it.

xgi/classes/dihypergraph.py
nwlandry: Yes, that's correct. I still have yet to update the docs and add unit tests.

xgi/classes/dihypergraph.py
nwlandry: I agree. I will open as an issue.

nwlandry:
Major accomplishments with this PR:
* Added a DiHypergraph class
* Added DiEdgeView and DiNodeView classes
* Added directed node and edge stats which still inherit from IDStat
* Added many unit tests and corresponding documentation. Overall testing coverage goes down by ~1%, but I think that as we add additional functionality, that this will improve.

PR #300, opened 2023-03-21
Return axes

nwlandry:
Updated the plotting functions:
* Renamed `xgi_pylab.py` to `draw.py`.
* Removed `draw_xgi_nodes` became `draw_nodes`, `draw_xgi_hyperedges` became `draw_hyperedges`, and `draw_xgi_simplices` became `draw_simplices`.
* Added tests for the drawing helper functions.
* Added a `pca_transform` function which can be used to control the rotation of the drawing relative to the principal axes.


maximelucas:
Look great, thanks. The renaming is a good idea. 

Do you think the codecov/patch is failing because your new PCA function is not tested?

Could you add an example of how to use the PCA function in the notebook?

nwlandry:
All your comments should be addressed now! Added unit tests and an example in Tutorial 5.

xgi/drawing/layout.py
```diff
@@ -295,17 +295,17 @@ def weighted_barycenter_spring_layout(H, return_phantom_graph=False, seed=None):
...
     ----------
     pos : dict of numpy arrays
         The output from any layout function
-    theta : float
+    theta : float, optional
         The angle between the horizontal axis and the principal axis
-        measured counterclockwise.
+        measured counterclockwise, by default 0.
     degrees : bool, optional
         Whether the angle specified is in degrees, by default True
```
maximelucas, line 310: "If False, the angle is radians" I guess

maximelucas (approved):

maximelucas:
Left one more minor comment. Also the notebook for tutorial 5 has all outputs cleared. Could you rerun it to show the outputs?

xgi/drawing/layout.py
nwlandry: Fixed.

nwlandry:
I looked up how to fix the issue and hopefully the configuration file that I added will fix the error. The issue is that even though this PR increases the coverage, it edits code that doesn't have any coverage so it treats it like I just committed a bunch of code without tests.

maximelucas:
Yea that's what I understood too.
Looks great, thanks :)

PR #303, opened 2023-03-22
Speed and memory improvements in linalg

### LINKED ISSUE #301: multiorder laplacian not truly sparse
If `sparse=True` is passed to `multiorder_laplacian()`, the function computes Laplacians at each order with `sparse=False`, sums them, and then converts the results to sparse with:

https://github.com/ComplexGroupInteractions/xgi/blob/ebe04761fabe31a474664b476c5464c41ba651cf/xgi/linalg/matrix.py#L369-L370

This defeats the purpose, since all internal calculations are done using non-sparse matrices.

maximelucas:
- made `incidence_matrix()` cleaner and faster (about 4x)
- made output consistent for sparse/non-sparse for extreme cases (empty Hypergraph, no edges of order d)
- added tests
- made `multiorder_laplacian()` "truly sparse": all internal variable are sparse if sparse=True. This fixes #301.

As a result, `multiorder_laplacian()` is also faster.

Timings for the `email-eu` dataset: 
```python
%timeit LL = xgi.multiorder_laplacian(H, orders, weights, sparse=False)
# --> before: 9.5 s, after: 9.5s
%timeit LL = xgi.multiorder_laplacian(H, orders, weights, sparse=True)
# --> before: 9.5 s, after: 1.1s
```

That's a 10x speedup for the sparse case, and more memory efficient (all internals are sparse).

### DIFF
--- a/tests/drawing/test_layout.py
+++ b/tests/drawing/test_layout.py
@@ -1,6 +1,5 @@
-import pytest
-
 import numpy as np
+import pytest
 
 import xgi
 from xgi.exception import XGIError
@@ -81,6 +80,7 @@ def test_barycenter_spring_layout(hypergraph1):
     H = xgi.random_hypergraph(10, [0.2, 0.1])
     pos = xgi.barycenter_spring_layout(H)
     assert len(pos) == H.num_nodes
+
 
 def test_weighted_barycenter_spring_layout(hypergraph1):
 
--- a/tests/generators/test_nonuniform.py
+++ b/tests/generators/test_nonuniform.py
@@ -1,7 +1,7 @@
-import pytest
 import random
 
 import numpy as np
+import pytest
 
 import xgi
 
--- a/tests/linalg/test_matrix.py
+++ b/tests/linalg/test_matrix.py
@@ -596,10 +596,20 @@ def test_boundary_matrix(edgelist4):
 
 def test_empty_order(edgelist6):
     H = xgi.Hypergraph(edgelist6)
-    I, _, _ = xgi.incidence_matrix(H, order=1, index=True)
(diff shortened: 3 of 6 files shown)


nwlandry:
Thanks so much - this is awesome! One quick question which isn't essential for this PR is why the `A.setdiag(0)` statement in `adjacency_matrix()` throws a warning "Changing the sparsity structure of a csr_matrix is expensive. lil_matrix is more efficient." when the `order` keyword is specified. I tried figuring it out without any luck. Good to merge!

nwlandry:
One more thing - can you run isort and black?

maximelucas:
> why the `A.setdiag(0)` statement in `adjacency_matrix()` throws a warning "Changing the sparsity structure of a csr_matrix is expensive. lil_matrix is more efficient." when the `order` keyword is specified. 

From what I've understood, "changing the sparsity structure" here just means we are changing entries of A (the diagonal entries). And `csr_array` is not the ideal type of sparse array for this, it can be slow, so scipy is letting us know. `lil` an `coo` are better for modifying entries. In our case don't think it's a big problem, we're just modifying a few values once. May be worth checking in the future though.

maximelucas:
Btw, I have a 18k-node hypergraph, and the cluster used to takes hours and crash trying to compute the laplacian. It now takes 5s on my laptop 👀

nwlandry:
> Btw, I have a 18k-node hypergraph, and the cluster used to takes hours and crash trying to compute the laplacian. It now takes 5s on my laptop 👀

🤯

leotrs:
> Btw, I have a 18k-node hypergraph, and the cluster used to takes hours and crash trying to compute the laplacian. It now takes 5s on my laptop eyes

This would be nice to add as a regression test.

PR #308, opened 2023-03-27
keep_isolates argument in subhypergraph

### LINKED ISSUE #235: subhypergraph leaves unwanted isolated nodes
The workaround can be seen [here](https://github.com/ComplexGroupInteractions/xgi/pull/234#discussion_r1038056826), which is less than ideal IMO.

maximelucas:
Added keep_isolates argument in subhypergraph to fix #235.

### DIFF
--- a/xgi/classes/hypergraphviews.py
+++ b/xgi/classes/hypergraphviews.py
@@ -18,7 +18,7 @@ from .function import freeze
 __all__ = ["subhypergraph"]
 
 
-def subhypergraph(H, nodes=None, edges=None):
+def subhypergraph(H, nodes=None, edges=None, keep_isolates=True):
     """View of `H` applying a filter on nodes and edges.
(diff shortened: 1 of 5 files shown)


nwlandry (approved):
This looks awesome, thanks!

xgi/classes/hypergraphviews.py
```diff
@@ -18,7 +18,7 @@
 __all__ = ["subhypergraph"]
 
 
-def subhypergraph(H, nodes=None, edges=None):
+def subhypergraph(H, nodes=None, edges=None, keep_isolates=True):
```
leotrs, line 21: The default is True, but the docstring says the default is False!

maximelucas: Thanks, just fixed it.

Issue #309, opened 2023-03-28
clustering coefficient not 1 in all-to-all


maximelucas:
I'm not sure we're using the definition we want to be using for the clustering coefficient. 
If we are, we should make its behaviour more explicit in the docs. 

Running:
```python
H = xgi.random_hypergraph(3, [1, 1]) # all edges, all triangles
H.clustering()
# --> {0: 0.333, 1: 0.333, 2: 0.333} # those are not 1!
```

yields values that are not 1 (the value I expected). Changing the number of nodes also changes the result. 

Note that having only edges does work:
```python
H = xgi.random_hypergraph(3, [1]) # all edges, no triangles
H.clustering()
# --> {0: 1, 1: 1, 2: 1} # ok
```
because we are using a definition that directly generalises the pairwise clustering coefficient.

leotrs:
It sounds like we need a similar discussion as the one we had for density. (What is the right denominator?)

PR #312, opened 2023-03-28
Tests for draw functions

maximelucas:
- tests for draw functions
- added defaults param values to all draw functions
- tests for clustering

### DIFF
--- a/tests/drawing/test_draw.py
+++ b/tests/drawing/test_draw.py
@@ -95,9 +95,18 @@ def test_draw(edgelist8):
 
     ax = xgi.draw(H)
 
+    # number of elements
     assert len(ax.lines) == len(H.edges.filterby("size", 2))  # dyads
     assert len(ax.patches) == len(H.edges.filterby("size", 2, mode="gt"))  # hyperedges
     assert len(ax.collections[0].get_sizes()) == H.num_nodes  # nodes
+
+    # zorder
+    for line in ax.lines:  # dyads
+        assert line.get_zorder() == 3
+    for patch, z in zip(ax.patches, [2, 2, 0, 2, 2]):  # hyperedges
+        assert patch.get_zorder() == z
+    assert ax.collections[0].get_zorder() == 4  # nodes
+
     plt.close()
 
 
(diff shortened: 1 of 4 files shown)


tests/drawing/test_draw.py
```diff
@@ -85,3 +87,55 @@ def test_color_arg_to_dict(edgelist4):
...
+
+    ax = xgi.draw(H)
+
+    assert len(ax.lines) == len(H.edges.filterby("size", 2))  # dyads
+    assert len(ax.patches) == len(H.edges.filterby("size", 2, mode="gt"))  # hyperedges
+    assert len(ax.collections[0].get_sizes()) == H.num_nodes  # nodes
+    plt.close()
+
+
+def test_draw_nodes(edgelist8):
```
nwlandry, line 104: Can you add a test to make sure the lines, nodes, and hyperedges/simplices have the right zorder?

xgi/drawing/draw.py
```diff
@@ -888,3 +958,21 @@ def draw_hyperedge_labels(
         text_items[id] = t
 
     return text_items
+
+
+def _update_lims(pos, ax):
```
nwlandry, line 963: Genius!

nwlandry (commented):
This looks good, thanks! A few things:
* I would be in favor of keeping it as `net` instead of `H` in nodestats, but if you feel strongly, make sure that it's consistent.
* If you could add a test to check that the nodes, lines, edges are plotted on the right layer, that would be awesome, but frankly what you have is a huge improvement, so no sweat if not.
* I think that someone mentioned that the clustering coefficient is not between 0 and 1. Can you add this as an issue?
* I like adding the optional arguments to the draw nodes/edges functions.

Overall looks great and I think I'm good with merging this!

tests/drawing/test_draw.py
maximelucas: Done

maximelucas:
All done. I had already raised issue #309 about the clustering coefficient :) We'll need to test it once the right answer is decided.

PR #313, opened 2023-03-28
Centrality update

nwlandry:
* Removed the `ZEC_centrality` because it doesn't converge and the paper by Austin Benson talks a bit about this
* Added tests for all centrality measures.
* Added additional tests for the xgi_data module.
* Added the ability for centrality to handle empty/invalid hypergraphs. Because these are Node/EdgeStats, I opted to return NaN when the centralities are invalid.


maximelucas:
Genius. Thanks so much.

maximelucas (approved):

PR #316, opened 2023-04-01
Add clustering coefficient definitions

nwlandry:


nwlandry:
Before I un-draft this PR, I wanted to know what good names for the functions were so that I can also add them to the nodestats module. Any suggestions, @leotrs? I was thinking maybe we could remove "coefficient" from the names.

xgi/algorithms/clustering.py
```diff
@@ -0,0 +1,233 @@
...
+from ..linalg import adjacency_matrix
+
+__all__ = [
+    "local_clustering_coefficient",
+    "clustering_coefficient",
+    "two_node_clustering_coefficient",
+]
+
+
+# Properties of metabolic graphs: biological organization or representation artifacts?
```
maximelucas, line 13:
```suggestion
```

xgi/algorithms/clustering.py
```diff
@@ -0,0 +1,233 @@
...
+def local_clustering_coefficient(H):
+    """Compute the local clusterin coefficient
```
maximelucas, line 15: clusterin' sounds cool

xgi/algorithms/clustering.py
```diff
@@ -0,0 +1,233 @@
...
+
+            # include normalisation by degree k*(k-1)/2
+            result[n] = 2 * total_eo / (dv * (dv - 1))
+    return result
+
+
+def clustering_coefficient(H):
+    """Return the clustering coefficients for
+    each node in a Hypergraph.
+
```
maximelucas, line 94: Here the definition is `cc[i] = A^3[i,i] / (k[i] * (k[i] - 1))` correct?

xgi/algorithms/clustering.py
```diff
@@ -0,0 +1,233 @@
...
+                    # add it up
+                    total_eo = total_eo + eo
+
+            # include normalisation by degree k*(k-1)/2
+            result[n] = 2 * total_eo / (dv * (dv - 1))
+    return result
+
+
+def clustering_coefficient(H):
+    """Return the clustering coefficients for
```
maximelucas, line 92: This description could also fit the first function.

maximelucas:
Incredible work Nich, thanks so much. We needed these different clustering coefficients. 

Two comments:
- some of the variable names are not very explicit and make it harder to understand the code.
- the description is of all three functions is very similar. Could we briefly describe their specificity, what they actually compute? Maybe describe their behaviour in cases where they differ. Maybe also add their definition with a formula?

xgi/algorithms/clustering.py
```diff
@@ -0,0 +1,233 @@
...
+                        )  # add it up
+                    # add it up
+                    total_eo = total_eo + eo
+
+            # include normalisation by degree k*(k-1)/2
+            result[n] = 2 * total_eo / (dv * (dv - 1))
+    return result
+
+
+def clustering_coefficient(H):
```
maximelucas, line 91: is this one "global" if the other one is "local"?

leotrs:
@nwlandry Thanks for your work here. I agree I'd remove `clustering` from all the names. Other than that, we kinda do need descriptive names so I'm not sure there's a good way of shortening them further.

xgi/algorithms/clustering.py
nwlandry: Almost. Instead of k, it's the number of neighbors.

nwlandry:
> Incredible work Nich, thanks so much. We needed these different clustering coefficients.
> 
> Two comments:
> 
> * some of the variable names are not very explicit and make it harder to understand the code.
> * the description is of all three functions is very similar. Could we briefly describe their specificity, what they actually compute? Maybe describe their behaviour in cases where they differ. Maybe also add their definition with a formula?

Thanks :) I added a brief description in the docstrings. Some of the formulas are a bit long so I just described them.

nwlandry:
> @nwlandry Thanks for your work here. I agree I'd remove `clustering` from all the names. Other than that, we kinda do need descriptive names so I'm not sure there's a good way of shortening them further.

After chatting with Max, I decided to keep "coefficient" in the name for maximum clarity, but if we feel it's too clunky down the road, we can remove it.

nwlandry:
I think it should be ready for review!

xgi/algorithms/clustering.py
```diff
@@ -10,8 +10,64 @@
...
+    result = {}
+    for n in H.nodes:
+        k = len(H.nodes.neighbors(n))
+        denom = k * (k - 1) / 2
+        if denom == 0:
+            result[n] = np.NaN
+        else:
+            i = node_to_index[n]
+            result[n] = 0.5 * mat[i, i] / denom
+    return result
```
leotrs, line 62: Here `k` is first an array and later overwritten with a different integer every iteration of the loop. Also the loop is unnecessary. Here's my suggestion, haven't tested it tho, but this should be faster:
```suggestion
    adj, index = xgi.adjacency_matrix(H, index=True)
    node_to_index = {n: i for i, n in index.items()}

    k = adj.sum(axis=1)
    denom = k * (k - 1) / 2
    mat = adj.dot(adj).dot(adj)

    with np.errstate(divide='ignore', invalid='ignore'):
        result = np.nan_to_num(0.5 * mat.diagonal() / denom)

    return {n: result[i] for i, n in index.items()}
```

maximelucas:
No experience or strong opinion but if 0 seems used more let's go with that I guess

leotrs:
No strong opinion either so I'll let someone else decide this one :wink:

leotrs (approved):

maximelucas (approved):

nwlandry:
> No experience or strong opinion but if 0 seems used more let's go with that I guess

Sounds good! I'll add a note about this as well :)

nwlandry:
Just to leave good documentation on what I did:
* I added three definition of the clustering coefficient and added these definitions to the algorithms module as well as nodestats.
* I moved the utils and linalg modules to their own folder in the docs.
* I renamed the centrality measures to be more intelligible.
* I changed `np.NaN` to `np.nan` as per (https://numpy.org/doc/stable/reference/constants.html)
* I changed the evaluation of NaN values to `np.isnan(val)` as per (https://numpy.org/doc/stable/user/misc.html)
* I added unit tests for the clustering coefficients.

PR #318, opened 2023-04-02
Test line graph

### LINKED ISSUE #298: add tests for #290
Add test for the two new functions introduced in #290 :
- `convert_to_line_graph()`
- `vector_centrality()`

nwlandry:
* changed the name from `convert_to_line_graph` to `to_line_graph`.
* added `s-parameter` like in "Hypernetwork science via high-order hypergraph walks"
* added tests

Fixes #298.

### DIFF
--- a/xgi/convert.py
+++ b/xgi/convert.py
@@ -136,8 +136,8 @@ def convert_to_graph(H):
     return G
 
 
-def to_line_graph(H):
-    """Line graph of the hypergraph.
+def to_line_graph(H, s=1):
+    """The s-line graph of the hypergraph.
 
     The line graph of the hypergraph `H` is the graph whose
     nodes correspond to each hyperedge in `H`, linked together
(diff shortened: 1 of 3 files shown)


leotrs (approved):

xgi/convert.py
```diff
@@ -146,22 +147,30 @@ def convert_to_line_graph(H):
...
+    https://doi.org/10.1140/epjds/s13688-020-00231-0
     """
 
     LG = nx.Graph()
 
     edge_label_dict = {tuple(edge): index for index, edge in H._edge.items()}
 
-    nodes = sorted(set(edge_label_dict.values()))
-    LG.add_nodes_from(nodes)
+    LG.add_nodes_from(H.edges)
```
maximelucas, line 170: The nodes might be "unordered" now. Better to keep them ordered?

Also, just realised some of those nodes might not get edges and stay isolated. Not sure if we want that here or not.

xgi/convert.py
```diff
@@ -146,22 +147,30 @@ def convert_to_line_graph(H):
...
 
     for edge1, edge2 in combinations(H.edges.members(), 2):
-        if edge1.intersection(edge2):
+        if len(edge1.intersection(edge2)) >= s:
             LG.add_edge(edge_label_dict[tuple(edge1)], edge_label_dict[tuple(edge2)])
```
maximelucas, line 174: It would be faster to collect new edges in a list and then add them with `add_edges_from` I think

maximelucas:
Great work Nich!!

maximelucas (approved):

xgi/convert.py
nwlandry: I intentionally removed the sorting because there will be cases where sorting isn't possible. (Mixes of strings and ints for example)

xgi/convert.py
nwlandry: Good idea!

xgi/convert.py
nwlandry: Exactly. The `vector_centrality` measure is downstream of this method and relies on the line graph having the same nodes, so I will keep it as is for now.

xgi/convert.py
nwlandry: Okay, update is that modifying the for-loop to this:
```
edges_for_adding = []
    for edge1, edge2 in combinations(H.edges.members(), 2):
        if len(edge1.intersection(edge2)) >= s:
            edges_for_adding.append((edge_label_dict[tuple(edge1)], edge_label_dict[tuple(edge2)]))
    LG.add_edges_from(edges_for_adding)
```
is 3% slower than the current algorithm. Let me know if there's a way of writing this that might be faster.

maximelucas: Hm okay, unexpected. Then let's leave it as is for now I guess. `add_edges_from` may need some optimizing.

Issue #272, opened 2023-01-20
node size param accepts negative values without warning


maximelucas:
We should check that lengths, widths, sizes, etc. passed to the draw functions are non-negative and probably raise an error/warning. Otherwise matplotlib seems to go full chaos.

nwlandry:
This is a good point. It shouldn't be an issue when we use Stats to color/size nodes/edges, but for everything else, we should add a check.

PR #320, opened 2023-04-03
Drawing hypergraphs with hyperdedges as convex hulls 

thomasrobiglio:
I have defined a new function called `draw_hypergraph_hull` that allows for drawing hypergraphs with hyperdedges of order > 1 as convex hulls. This is done using the `ConvexHull` function from `scipy.spatial`. The drawing style is inspired from that of HyperNetX but implemented differently and coherently with the rest of xgi. I have also pushed a tutorial showing the different setting that can be passed to the function.


xgi/drawing/draw.py
```diff
@@ -985,3 +991,155 @@ def _update_lims(pos, ax):
...
+                alpha = 1,
+            )
+            ax.add_line(line)
+            
+        else:
+            coordinates = [[pos[n][0], pos[n][1]] for n in he]
+            draw_hull(node_positions=np.array(coordinates), radius=radius, ax=ax, contour_color='tab:gray', filling_color=palette[d-1], alpha_filling=alpha_filling)
+    
+    G=convert_to_graph(H)
+    nx.draw_networkx_nodes(G, pos=pos, node_size=node_size, node_color=node_color, node_shape=node_shape, ax=ax, linewidths=node_linewidths, edgecolors=node_edgecolors, alpha=1)
```
iaciac: Any particular reason not to use the `draw_nodes` from xgi instead of the networkx one? I'm just saying this for consistency with the other plotting functions

xgi/drawing/draw.py
```diff
@@ -985,3 +991,155 @@ def _update_lims(pos, ax):
...
+
+    ax.get_xaxis().set_ticks([])
+    ax.get_yaxis().set_ticks([])
+    ax.axis("off")
+    
+    if not max_order:
+        max_order = max_edge_order(H)
+    
+    if palette is None:
+        palette=sns.color_palette("deep", 6)
```
iaciac: I would probably pick a matplotlib palette to avoid having to add seaborn to the requirements just for this one

xgi/drawing/draw.py
thomasrobiglio: Hi @iaciac, actually there was no reason for using the nx, that's why it is corrected in a later push (https://github.com/ComplexGroupInteractions/xgi/pull/320/commits/ca1451f1a5c348412868eec9a709f33d22b75bd1)

xgi/drawing/draw.py
thomasrobiglio: Same as above, in my early attempts there were a lot of inconsistencies with the rest of xgi that I've corrected with the help of @maximelucas

xgi/drawing/draw.py
iaciac: Oooops sorry

nwlandry:
One quick ask: can you delete the `**2` from `node_size**2` in `draw_nodes`? I think it would partially fix #272.

maximelucas:
We could yea but I think the `**2` is how it works in most of matplotlib, like the scatter function?
In any case, if we do change that, we'd need to adapt the default size value.

leotrs:
It seems like accepting negative values is separate from whether or not we want to square them for scale. I'd leave that for a different PR.

xgi/drawing/draw.py
```diff
@@ -985,3 +987,246 @@ def _update_lims(pos, ax):
...
+
+    points = node_pos.copy()
+    for i in node_pos:
+        center = i
+        thetas = np.linspace(0, 2 * np.pi, num=100, endpoint=False)
+        circle = [
+            center + radius * np.array([np.cos(theta), np.sin(theta)])
+            for theta in thetas
+        ]
+        points = np.vstack([circle, points])
```
leotrs, line 1023: I believe the following two functions achieve the same result. The first is essentially a copy paste from the highlighted code.
```python3
def foo():
    node_pos = np.random.random(size=(100, 2))
    radius = 1.5
    points = node_pos.copy()

    for i in points:
        center = i
        thetas = np.linspace(0, 2*np.pi, num=100, endpoint=False)
        circle = [center + radius * np.array([np.cos(theta), np.sin(theta)]) for theta in thetas]
        points = np.vstack([circle, points])

    return points


def foo2():
    node_pos = np.random.random(size=(100, 2))
    radius = 1.5
    thetas = np.linspace(0, 2*np.pi, num=100, endpoint=False)
    points = np.vstack([
        p + radius * np.array([np.cos(thetas), np.sin(thetas)]).T
        for p in node_pos
    ])
    return np.vstack([node_pos, points])
```

If we all agree that these indeed produce the same output, then we should use the second one, as it's 40x faster:
```python3
In [37]: %timeit foo()
40 ms ± 620 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [38]: %timeit foo2()
698 µs ± 11.9 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

```

thomasrobiglio: Hi @leotrs, thank you for the input, it is now implemented in the second way.

leotrs: Thanks!

xgi/drawing/draw.py
```diff
@@ -1011,16 +1011,10 @@ def _draw_hull(node_pos, ax, edges_ec, facecolor, alpha, zorder, radius):
...
-        thetas = np.linspace(0, 2 * np.pi, num=100, endpoint=False)
-        circle = [
-            center + radius * np.array([np.cos(theta), np.sin(theta)])
-            for theta in thetas
-        ]
-        points = np.vstack([circle, points])
+    
+    thetas = np.linspace(0, 2*np.pi, num=100, endpoint=False)
+    points = np.vstack([p + radius * np.array([np.cos(thetas), np.sin(thetas)]).T for p in node_pos])
+    points = np.vstack([node_pos, points])
```
leotrs, line 1017: Even faster:
```suggestion
    offsets = radius * np.array([np.cos(thetas), np.sin(thetas)]).T
    thetas = np.linspace(0, 2*np.pi, num=100, endpoint=False)
    points = np.vstack([p + offsets for p in node_pos])
    points = np.vstack([node_pos, points])
```

leotrs (approved):

leotrs:
Thank you so much @thomasrobiglio ! I've approved but I'll let @maximelucas decide when to merge in case there's anything else to do here.

maximelucas:
Thanks Leo!

For completeness, the last thing would be to add a simple test for `draw_hypergraph_hull`, mimicking the other drawing tests here:
https://github.com/ComplexGroupInteractions/xgi/blob/main/tests/drawing/test_draw.py

Are you up for it @thomasrobiglio ?

thomasrobiglio:
@maximelucas I have added the test. I could not implement also the test for the number of elements and zorder of the dyads because the drawing of the hulls also generated 2D lines (and the number of those is very initialisation dependent). For the moment I couldn't think of a solution to this.

maximelucas:
Thanks Thomas, good to go for me!

nwlandry:
Thanks so much @thomasrobiglio!! Looks awesome!!

leotrs:
Congratulations @thomasrobiglio on your first merged PR. Hopefully first of many ;)

iaciac:
Amazing, @thomasrobiglio! This is both what XGI deserves and what it needs right now. Thanks a lot

Issue #319, opened 2023-04-03
improved folder structure for generators


maximelucas:
Right now, our `generators` are divided into 3 files: `classic`, `uniform`, `nonuniform`.

I see a few disadvantages. In `classic`, I feel like we've started putting everything that's not "purely random".  Also, our structure does not seem very scalable if we have more and more generators in the future. Finally, some generators are almost duplicates between `uniform` and `nonuniform` (`uniform_erdos_renyi_hypergraph` and `random_hypergraph`).

Networkx has the following structure: https://github.com/networkx/networkx/tree/main/networkx/generators
with more files (probably too many for us) with more explicit names.

We could have files like: 
- `classic`: `empty_hypergraph`, `complete_hypergraph` (new), `trivial_hypergraph`(new), etc.
- `simple`: `star_clique`, `sunflower`, etc.
- `lattice`: `ring_lattice`, + other new lattices
- `simplicial_complexes`: `flag_complex`, `flag_complex_d2`, etc. (this name might not be the best)
- `random`: `random_hypergraph`, `chung_lu_hypergraph `, `dcsbm_hypergraph `, `watts_strogatz_hypergraph`, etc.

For generators that are currently "duplicated" between `nonuniform` and `uniform`, we could could have common interface:
```python
xgi.random_hypergraph(N=10, ps=[0.1, 0.01])  # builds hyperedges for orders from 1 to len(ps)
xgi.random_hypergraph(N=10, ps=[0.1, 0.01], orders=[1, 2])  # equivalent to above
xgi.random_hypergraph(N=10, ps=[0.1])  # builds hyperedges for order 1
xgi.random_hypergraph(N=10, ps=[0.01], orders=[2])  # builds hyperedges for order 2 (uniform)
```
Other option: argument `order` that if `None` infers orders like now, and if an `int` say 2, goes uniform. I guess we would rarely need to use `orders=[2,3]` anyway.

I'm not sure if that would make sense for all generators currently in `uniform`, e.g. `configuration_model` or `planted_partition`. @nwlandry do you think we could fit those in this new structure in a way that makes sense?

This suggestion is also in line with making our file structure clearer and more discoverable.

nwlandry:
I think that we should address #292 when we work on this.

I think that most of these suggestions make sense. The only reason that I differentiated between uniform and non-uniform is because of the Chung-Lu/DCSBM where edge sizes are not guaranteed.

maximelucas:
[...]

PR #321, opened 2023-04-04
Restructure folder for generators and linalg

maximelucas:
Following #319


maximelucas:
It's already big enough to review I'd say. 

Done:
- divided `linalg` into: hypergraph_matrix, hodge_matrix, and laplacian_matrix
- divided the `generators` into: classic, simple, lattice, random, uniform, simplicial_complexes
- merge the uniform and nonuniform `random_hypergraph()` (this resulted in not have the option to pass a desired degree like the uniform version had, but we could add an external function to convert p to degree)
- divided the tests accordingly

nwlandry:
Overall, looks good! Some high-level things:

* Can you update the docs to match the change in structure? In addition, can you move utils and linalg into their own folders and update with the new module names? For reference, see PR #316.
* The one nice thing about the uniform Erdös-Rényi hypergraph is that it had O(|V| +|E|) complexity which I needed for some larger hypergraphs. Is there a way to incorporate this in the `random_hypergraph()` function?

xgi/linalg/hypergraph_matrix.py
nwlandry, line 1: I don't know if I like the "hypergraph_matrix.py" name. Maybe break this into "incidence" and "adjacency"? But of course the intersection profile and degree matrices don't clearly belong to either. My concern is that these matrices aren't exclusively for hypergraphs in the way that we set them up.

xgi/linalg/hypergraph_matrix.py
```diff
@@ -0,0 +1,289 @@
+"""Matrices associated to hypergraphs.
+
```
nwlandry, line 2: I think that for now, having the same header for every file might be okay, but we should at least change the first line because that shows up when navigating through readthedocs.

xgi/linalg/hypergraph_matrix.py
nwlandry, line 1: Because all these are matrices, do we need the `_matrix` at the end of the filename?

codecov.yml
```diff
@@ -1,14 +1,12 @@
 coverage:
   precision: 2
   round: down
-  range: 60..85
+  range: 70..100
```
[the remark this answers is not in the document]
maximelucas: I had it wrong, changed it back

xgi/linalg/hypergraph_matrix.py
maximelucas: I updated the first lines to be a bit more specific

xgi/linalg/hypergraph_matrix.py
maximelucas: Maybe not. Networkx has them because they have some files that are not matrices. Maybe let's keep for now and remove them later if we never need other files?

nwlandry:
Hi @maximelucas - I had the opportunity to look over this PR and looks great! I hope you don't mind that I did the following:
* I added the specific files relating to the modules that you added.
* I added the uniform_erdos_renyi_hypergraph back in, and I will add this as an issue so that eventually we can port it over to the `random` module as `fast_random_hypergraph` or something.

nwlandry (approved):

maximelucas:
Perfect, thanks Nich!

PR #323, opened 2023-04-05
Index-to-unit mappings from to_bipartite_graph. Fixes #322

### LINKED ISSUE #322: `to_bipartite_graph` documented return type is incorrect
The docstring for `xgi.to_bipartite_graph` states that the function returns a bipartite graph directly:
https://xgi.readthedocs.io/en/stable/api/convert.html?highlight=to_bipartite_graph#xgi.convert.to_bipartite_graph

But the function returns a triplet of a graph with two mappings, I think from integers to node / edges IDs?
https://xgi.readthedocs.io/en/stable/_modules/xgi/convert.html#to_bipartite_graph

leotrs:
The default has been set to False to mirror the behavior of linalg. Fixes #322.


maximelucas:
Look good thanks Leo!
Maybe just add one line to test the `index=False` case?

maximelucas (approved):

leotrs:
Thanks @maximelucas I added the test.

Issue #217, opened 2022-11-08
faster implementation of finding maximal simplices


maximelucas:
This question https://stackoverflow.com/questions/14106121/efficient-algorithm-for-finding-all-maximal-subsets contains an answer with an efficient python-coded solution. We might want to incorporate this to replace ours.

nwlandry:
This is a great idea! Especially since we've already built the forward and reverse maps.

nwlandry:
This is a first pass at a faster version - from initial testing, this method is anywhere from 4x-40x faster than the current method. The naïve approach is first and then the method based on the thread you linked is second:
```
def maximal_faces(H):
    """Gets the maximal faces given a hypergraph

    Parameters
    ----------
    H : Hypergraph

    Returns
    -------
    dict of iterables
        The dict of maximal faces.
    """
    edges = H.edges.members(dtype=dict)
    max_faces = {}

    for i, e in edges.items():
        flag = True
        old_max_faces = dict(max_faces)
        for j, f in old_max_faces.items():
            if set(e).issubset(f):
                flag = False
                break
            elif set(f).issubset(e):
                del max_faces[j]
        if flag:
            max_faces[i] = e
    return max_faces

def fast_maximal_faces(H):
    """Gets the maximal faces given a hypergraph

    Parameters
    ----------
    H : Hypergraph

    Returns
    -------
    dict of iterables
        The dict of maximal faces.
    """
    edges = H.edges.members(dtype=dict)
    nodes = H.nodes.memberships()
    max_faces = {}

    intersection = lambda x, y: x & y

    for i, e in edges.items():
        in_common = reduce(intersection, [nodes[n] for n in e])
        
        if len(in_common) == 1:
            [elem] = in_common
            if elem == i:
                max_faces[i] = e
    return max_faces
```

nwlandry:
Two more things:
1. The new method gets rid of multiples of maximal edges.
2. Finding maximal edges can be useful not just for simplicial complexes, but for hypergraphs too. I think we should extend the method to work for both. (The last I checked, it only worked for simplicial complexes)

leotrs:
Code looks good. Yes to also implementing it on hypergraphs.

PR #324, opened 2023-04-08
Added `maximal` method to EdgeView

nwlandry:
* Added the maximal method for both hypergraphs and simplicial complexes in a faster implementation described in #217
* Fixed an incidental bug in the `duplicates()` method in `IDView`
* Removed the old `maximal_simplices` method.


xgi/classes/reportviews.py
```diff
@@ -688,3 +692,48 @@ def singletons(self):
...
+        >>> H.edges.maximal().members()
+        [{1, 2, 3}, {3, 4}, {1, 2, 3}]
+        """
+        edges = self._id_dict
+        nodes = self._bi_id_dict
+        max_edges = set()
+
+        hashes = defaultdict(list)
+        for idx, members in edges.items():
+            hashes[frozenset(members)].append(idx)
```
leotrs, line 728: This is for repeated edges, no?

xgi/classes/reportviews.py
```diff
@@ -688,3 +692,48 @@ def singletons(self):
...
+
+        _intersection = lambda x, y: x & y
+
+        for i, e in edges.items():
+            if i not in max_edges:
+                in_common = reduce(_intersection, [nodes[n] for n in e])
```
leotrs, line 734: More memory efficient, and I don't think it would have much impact in running time:
```suggestion
                in_common = reduce(_intersection, (nodes[n] for n in e))
```

xgi/classes/reportviews.py
nwlandry: correct.

xgi/classes/reportviews.py
```diff
@@ -688,3 +692,55 @@ def singletons(self):
 
         """
         return self.filterby("size", 1)
+
+    def maximal(self):
+        """Returns the maximal edges as an EdgeView
```
maximelucas, line 697: I'd make it more explicit what a maximal edge, for example:
```suggestion
        """Returns the maximal edges as an EdgeView. Maximal edges are those that are not subsets of any other edges in the hypergraph.
```

maximelucas:
Looks great, thanks. 

You also changed this from being an XGI function to a View/method of Hypergraphs. This seems like a good idea. I'm just wondering if you had strong reasons for this, thinking about other XGI function in general?

maximelucas (approved):

nwlandry:
My main reason was that it seemed like a function related to edges and I'm in favor of migrating away from the function module to improve the structure of the docs like you've commented on before.

leotrs:
If it takes in an EdgeView and returns an EdgeView, then the most logical place to put it is EdgeView.

Just a nitpick for posterity. The current implementation has some code specifically to get rid of the duplicate edges. This adds a whole iteration over the edge set. If the user knows beforehand that there are no (or few) repeated edges, then this iteration is useless and would only be slowing down the computation. Perhaps in the future we can make this iteration optional and allow the user to determine whether or not it happens.

leotrs:
In a similar vein as before. We could also add another optional optimization where instead of doing reduce(intersection, ...), we could add a custom loop that quits early if any of the intermediate intersections reaches a desired property (that of having one item). If that's the case then you already know the edge is maximal and you don't need to look at the rest.

Like my prev comment, this would result in an optimization only in some cases so whether it's effective would depend on the structure of the network.

nwlandry:
> If it takes in an EdgeView and returns an EdgeView, then the most logical place to put it is EdgeView.
> 
> Just a nitpick for posterity. The current implementation has some code specifically to get rid of the duplicate edges. This adds a whole iteration over the edge set. If the user knows beforehand that there are no (or few) repeated edges, then this iteration is useless and would only be slowing down the computation. Perhaps in the future we can make this iteration optional and allow the user to determine whether or not it happens.

This is a good idea. I will do that!

PR #332, opened 2023-04-11
Added multiedge keyword to `maximal()` method

nwlandry:
Also removed an inconsistent unit test.

### DIFF
--- a/xgi/classes/reportviews.py
+++ b/xgi/classes/reportviews.py
@@ -693,9 +693,15 @@ class EdgeView(IDView):
         """
         return self.filterby("size", 1)
 
-    def maximal(self):
-        """Returns the maximal edges as an EdgeView. Maximal edges are those that are not subsets of any other edges in the hypergraph.
+    def maximal(self, multiedges=True):
+        """Returns the maximal edges as an EdgeView.
 
+        Maximal edges are those that are not strict subsets of any other edges in the hypergraph.
+
+        Parameters
+        ----------
+        multiedges : bool
+            Whether to allow multi-edges to be maximal edges.
         Returns
         -------
         EdgeView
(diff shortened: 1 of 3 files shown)


xgi/classes/reportviews.py
```diff
@@ -693,9 +693,15 @@ def singletons(self):
         """
         return self.filterby("size", 1)
 
-    def maximal(self):
-        """Returns the maximal edges as an EdgeView. Maximal edges are those that are not subsets of any other edges in the hypergraph.
+    def maximal(self, multiedges=True):
```
leotrs, line 696: I think the current parameter name and documentation are not terribly descriptive. I suggest  something like `with_multiplicity`, with documentation along the liens of "if a maximal edge occurs multiple times, return it according to its multiplicity". Also add a "Notes" section to the docstring that explains that setting `with_multiplicity` to True or False may have different performances depending on the number of repeated maximal edges.

leotrs (approved):

xgi/classes/reportviews.py
nwlandry: I just want to clarify that this keyword specifies whether multiedges can be maximal, not whether to return an edge with its multiplicity (though I might have misunderstood your point). If False, we are enforcing strict inclusion for maximal hyperedges and therefore, multiedges can't be included. If, however, multiedges=True, then we add all the maximal multiedges with non-strict inclusion.

I agree that it is not clear - any ideas for a new keyword name? I'm not sure `with_multiplicity` exactly captures what it is doing. I'll add a more detailed note like you suggested.

leotrs: Thanks for the explanation. I agree `with_multiplicity` is also not ideal. How about simply `strict`?

nwlandry: Perfect. I'll work on that right now.

PR #333, opened 2023-04-12
changed the org name to xgi

nwlandry:
Once this is approved, I will change the org name and then merge this PR.


nwlandry:
xgi is not available as a name

PR #334, opened 2023-04-13
change name to xgi-org

nwlandry:
To Do:

- [x] Change org name
- [x] Merge pull request
- [x] Release new version on PyPI so that the links work.
- [x] Update the link at JOSS


nwlandry:
codecov is no longer a python package, so I removed it.

PR #335, opened 2023-04-13
feat: added trivial hypergraph

maximelucas:


tests/generators/test_classic.py
```diff
@@ -9,3 +9,7 @@ def test_empty_hypergraph():
 def test_empty_hypergraph():
     SC = xgi.empty_simplicial_complex()
     assert (SC.num_nodes, SC.num_edges) == (0, 0)
+
+def test_trivial_hypergraph():
+    SC = xgi.trivial_hypergraph()
```
nwlandry, line 14: How about `H` instead of `SC`. Picky, I know ;)

xgi/generators/classic.py
```diff
@@ -98,3 +99,40 @@ def empty_simplicial_complex(create_using=None, default=None):
     if default is None:
         default = xgi.SimplicialComplex
     return _empty_network(create_using, default)
+
+
+def trivial_hypergraph(create_using=None, default=None):
+    """Returns a hypergraph with one node and zero edges.
```
nwlandry, line 105: Could you make this a more flexible function so that it creates a hypergraph with n nodes and zero edges, where 1 node is the default or something like that?

xgi/generators/classic.py
```diff
@@ -98,3 +99,43 @@ def empty_simplicial_complex(create_using=None, default=None):
     if default is None:
         default = xgi.SimplicialComplex
     return _empty_network(create_using, default)
+
+
+def trivial_hypergraph(n=1, create_using=None, default=None):
+    """Returns a hypergraph with one node and zero edges.
```
leotrs, line 105:
```suggestion
    """Returns a hypergraph with some nodes and zero edges.
```

xgi/generators/classic.py
```diff
@@ -98,3 +99,43 @@ def empty_simplicial_complex(create_using=None, default=None):
...
+        If a constructor, call it to create an empty hypergraph.
+    default : Hypergraph constructor (default None)
+        The constructor to use if create_using is None.
+        If None, then xgi.Hypergraph is used.
+
+    Returns
+    -------
+    Hypergraph object
+        A trivial hypergraph with `n` nodes
+
```
leotrs, line 122: See Also: empty_hypergraph

leotrs (approved):

xgi/generators/classic.py
maximelucas: Done

xgi/generators/classic.py
maximelucas: good catch, done

PR #337, opened 2023-04-17
feat: added complete_hypergraph

maximelucas:


tests/generators/test_classic.py
```diff
@@ -18,4 +23,52 @@ def test_trivial_hypergraph():
...
+    assert xgi.unique_edge_sizes(H1) == [2]
+    assert xgi.unique_edge_sizes(H2) == [2, 3]
+    assert xgi.unique_edge_sizes(H3) == [2, 3, 4]
+    assert xgi.unique_edge_sizes(H03) == [1, 2, 3, 4]
+
+    assert H1.num_edges == comb(N, 2)
+    assert H2.num_edges == comb(N, 2) + comb(N, 3)
+    assert H3.num_edges == comb(N, 2) + comb(N, 3) + comb(N, 4)
+    assert H03.num_edges == comb(N, 2) + comb(N, 3) + comb(N, 4) + N
+
```
nwlandry, line 68: I wonder if you could concisely test the edges with
```

assert H3.num_edges == [set(e) for r in [2, 3, 4] for e in combinations([0, 1, 2, 3], r)]
```

xgi/generators/classic.py
```diff
@@ -154,3 +157,59 @@ def trivial_hypergraph(n=1, create_using=None, default=None):
...
+        if max_order is None.
+
+    Return
+    ------
+    Hypergraph object
+        A complete hypergraph with `N` nodes
+
+    Note
+    ----
+    Only one of `order` and `max_order` can be specified by and int (not None).
```
nwlandry, line 187: I wonder if you could comment on the maximum edge order that is computationally feasible. Should there be a more conservative default so that people don't try to compute the complete hypergraph for $N=100$?

xgi/generators/classic.py
```diff
@@ -154,3 +157,59 @@ def trivial_hypergraph(n=1, create_using=None, default=None):
...
+    Additionally, at least one of either must be specified.
+    """
+    # this import needs to happen when the function runs, not when the module is first
+    # imported, to avoid circular imports
+    import xgi
+
+    if bool(order) == bool(max_order):
```
leotrs, line 194: The current version does not catch if both are specified as zero.
```suggestion
    if order is not None and max_oeder is not None:
```

maximelucas: Hahah I knew you were gonna make that comment. 
What I am doing trying to do is catching both those cases: (None, None), and (int, int).
Both cases are invalid and should raise the error.

I noticed it wouldn't catch 0 but thought using 0 for order or max_order is not useful anyway?

I can do a more verbose version with your line and another so that everything is explicit if you think that's better.

xgi/generators/classic.py
maximelucas: You mean mention that max_order < N -1?

The default (None) does not compute anything. So it's not computing all possible orders by default.

tests/generators/test_classic.py
nwlandry: mmmm. that's a good point. all good then.

xgi/generators/classic.py
nwlandry: I see - I should have looked more closely. Nonetheless, maybe you could add a rough estimate in the Notes section on max order/max $N$ for which computation is feasible. Something along the lines of "When the size of the hypergraph is smaller than 20, there are no restrictions on the max order. For all hypergraphs larger than this, it is advisable to set max_order to be less than 5 to make the computation feasible." It probably is more complicated than that, but just to warn the user that, for example, a hypergraph with 1,000 nodes and max_order=1000 will not work.

xgi/generators/classic.py
leotrs: I see. Yeah in general I think it's better to be explicit rather than implicit. Zen of python and what not,

PR #346, opened 2023-05-02
Try ruff

leotrs:
For a while now I've been meaning to run a [linter](https://en.wikipedia.org/wiki/Lint_(software)) over our codebase. These days, [ruff](https://beta.ruff.rs/docs/) seems to be what the cool kids are doing so that's what I did.

On the main branch, `ruff check .` shows 294 errors, while `ruff check . --ignore=F401,F403,E741,E501` shows 29 errors.

On this PR's branch, `ruff check .` shows 99 errors, while `ruff check . --ignore=F401,F403,E741,E501` shows 1 error.

The meaning of `--ignore=F401,F403,E741,E501` is that with this flag, ruff will not check for unused imports (rule E401), most of which happen within `__init__.py` modules anyway, will ignore star-imports (rule E403), will ignore ambiguous variable names such as `I` and `l` (rule E741), and will ignore line length restrictions (rule E501). Right now we have 99 exceptions to these rules. In the future it would be great to get this number as close to zero as possible.

It's a big PR; it might be easier to review commit-by-commit as I tried to group changes according to the same rules. For example, I think the first commit is the longest one, and it literally just rearranges code (mostly docstrings) according to line length restrictions.


xgi/algorithms/centrality.py
```diff
@@ -112,8 +112,8 @@ def h_eigenvector_centrality(H, max_iter=100, tol=1e-6):
 
     new_H = convert_labels_to_integers(H, "old-label")
 
-    f = lambda v, m: np.power(v, 1.0 / m)
-    g = lambda v, x: np.prod(v[list(x)])
+    f = lambda v, m: np.power(v, 1.0 / m)  # noqa: E731
```
maximelucas, line 115: the comment tells ruff to ignore a potential error?

xgi/readwrite/json.py
```diff
@@ -22,24 +22,24 @@ def write_json(H, path):
     # initialize empty data
     data = {}
 
-    # get overall hypergraph attributes, name always gets written (default is an empty string)
-    data["hypergraph-data"] = dict()
+    # name always gets written (default is an empty string)
+    data["hypergraph-data"] = {}
```
maximelucas, line 26: did we not change these to dict() in the past to avoid confusion with sets?

maximelucas:
Nice, thanks Leo. 

So this PR fixes issues shown by Ruff but does not implement Ruff as part of our workflow.
Do you mean to integrate it in the future?

Re. line length restrictions in doctstrings, I noticed that before, and I was wondering: why is `black` not taking care of that?

maximelucas (approved):

xgi/algorithms/centrality.py
leotrs: Yeah. For reference, you can see the explanation of each rule using `ruff rule <rule_number>`:

```bash
$ ruff rule E731
```
# lambda-assignment (E731)

Derived from the **pycodestyle** linter.

Autofix is sometimes available.

## What it does
Checks for lambda expressions which are assigned to a variable.

## Why is this bad?
Per PEP 8, you should "Always use a def statement instead of an assignment
statement that binds a lambda expression directly to an identifier."

Using a `def` statement leads to better tracebacks, and the assignment
itself negates the primary benefit of using a `lambda` expression (i.e.,
that it can be embedded inside another expression).

## Example
```python
f = lambda x: 2*x
```

Use instead:
```python
def f(x):
   return 2 * x
```

## References
- [PEP 8](https://peps.python.org/pep-0008/#programming-recommendations)

leotrs: I usually try to abide by this rule but in this case it makes sense to assign to a lambda, I think.

leotrs:
> So this PR fixes issues shown by Ruff but does not implement Ruff as part of our workflow. Do you mean to integrate it in the future?

If we collectively see it as a net positive then yes. If not, I'm happy to just keep a mental note and run it manually every once in a while. If we wanted to add it to our workflow, it would potentially create friction when merging PRs and onboarding new contributors ("please run ruff", etc).

> Re. line length restrictions in doctstrings, I noticed that before, and I was wondering: why is `black` not taking care of that?

Black does not do this. See [here](https://stackoverflow.com/a/71041192).

PR #347, opened 2023-05-03
Remove hardcoded alpha from hyperedge drawing

leotrs:
The transparency of hyperedges is currently hardcoded to 0.4.

The following code:
```python3
import xgi
H = xgi.Hypergraph([[1,2,3], [2,3,4]])
xgi.draw(H, edge_fc=[[0,0,1,1], [0,0,1,0.1]])
plt.show()
```

Before this PR:

After this PR:

Cc. @iaciac

### DIFF
--- a/xgi/drawing/draw.py
+++ b/xgi/drawing/draw.py
@@ -466,7 +466,6 @@ def draw_hyperedges(
             obj = plt.Polygon(
                 sorted_coordinates,
                 facecolor=edge_fc[id],
-                alpha=0.4,
                 zorder=max_order - d,
             )
             ax.add_patch(obj)


nwlandry:
I like the concept of this, but I picture alpha being a separate parameter with a default value of 0.4, instead of lumped with the color. As it is implemented in this PR, drawing with the default parameters, i.e., `xgi.draw(H)`, the edges will not be transparent at all, right?

leotrs:
Yeah you're right. 

I will say however that specifying the alpha of each color is actually very common.

Right now, the default value of 0.4 overrides any alpha value in the color. The current PR fixes this. I can also try to make it so that we accept alpha values as a separate array, separate from color. I honestly do not like that idea at all since then there would be two different ways of specifying the same thing.

nwlandry:
I see. So is there a way to have a default alpha entry for facecolor which gets overridden when the user specifies it?

leotrs:
Yeah you're right. In light of this being what MPL does, we should follow suit too.

leotrs:
Thanks both for your comments. I think having a default alpha value is nice too.

After playing with `plt.scatter`, it seems like:
1. specifying alpha as part of the color is supported,
2. specifying alpha as a standalone optional argument `alpha=0.5` is supported,
3. if both are specified, the latter overrides the former.

This is counterintuitive to me, but whatever. Do we want to mirror the same interface?

nwlandry:
> Thanks both for your comments. I think having a default alpha value is nice too.
> 
> After playing with `plt.scatter`, it seems like:
> 
> 1. specifying alpha as part of the color is supported,
> 2. specifying alpha as a standalone optional argument `alpha=0.5` is supported,
> 3. if both are specified, the latter overrides the former (i.e. all elements are plotted with the same alpha value, even if they each have different alpha values as part their color).
> 4. the argument `alpha=0.5` can only receive one float, i.e. you cannot pass in an array specifying different alpha values, one for each element.
> 
> This is counterintuitive to me for many reasons, but whatever. Do we want to mirror the same interface?

I think this plan makes sense for now. If we handle the logic flow that you've described in `_color_arg_to_dict()` we can (potentially?) always bundle the alpha value with the color when we convert to a dict.

maximelucas:
> Do we want to mirror the same interface?

I'd say let's mirror this for now yes

maximelucas:
I think this was solved with #456. If you're happy with this @leotrs we can close.

nwlandry:
I'm closing this since this was solved with #456. @leotrs can re-open this if necessary.

PR #348, opened 2023-05-03
Update sphinx theme

nwlandry:
Based on @leotrs suggestion.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -64,8 +64,8 @@ designed and developed by network scientists with the needs of network scientist
 mind.
 
 - Repository: https://github.com/xgi-org/xgi
-- PyPI: https://pypi.org/project/xgi/
-- Documentation: https://xgi.readthedocs.io/
+- PyPI: `latest release <https://pypi.org/project/xgi/>`_
+- Twitter: `@xginets <https://twitter.com/xginets>`_
 
 Sign up for our `mailing list <http://eepurl.com/igE6ez>`_ and follow XGI on `Twitter <https://twitter.com/xginets>`_ or `Mastodon <https://mathstodon.xyz/@xginets>`_!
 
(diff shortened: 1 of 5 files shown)


docs/source/index.rst
```diff
@@ -64,8 +64,8 @@ designed and developed by network scientists with the needs of network scientist
 mind.
 
 - Repository: https://github.com/xgi-org/xgi
-- PyPI: https://pypi.org/project/xgi/
-- Documentation: https://xgi.readthedocs.io/
```
leotrs, line 68: Why remove the link to the docs?

nwlandry: This might be a best practices thing, but I removed it because if the user is seeing the link (rendered as html), they will already be at the documentation homepage. Happy to add it back in.

nwlandry:
I added a more organized and concrete contribution guide and eliminated some redundancy in the GH, PyPI, and RTD landing pages.

maximelucas:
This theme is visually more pleasing. I just find two downsides to it:
1. The full menu is not shown anywhere. It used to be shown as a left sidebar. Now it's horizontal at the top of the page, but after "Core Classes" it just says "More" with a dropdown menu.
2. it feels much less compact than our current one, so I feel like I need to scroll a lot don't have a good overview of a page (I feel zoomed in). For example compare:
    - https://xgi--348.org.readthedocs.build/en/348/api/generators.html
    - https://xgi.readthedocs.io/en/stable/api/generators.html

Is there a way to change that?

docs/source/index.rst
leotrs: Okay gotcha, I guess that's fine.

leotrs:
Agree with Max, we should work on the top nav bar.

nwlandry:
Okay, I think it would make sense to split this into 2 PRs: this one will update the contributing guide and up-version sphinx, and a future PR will change the theme. How does this sound?

leotrs:
Yeah that makes more sense. Thanks Nich.

PR #352, opened 2023-05-05
Move tutorials

nwlandry:
Moving the tutorials to a separate repo so that people don't have to clone the entire XGI repo when all they want is the tutorials.


leotrs:
Have you received complaints about having to fork the entire thing?

Another solution (cleaner, I think) is to provide a binder for (or some other way of downloading) each tutorial.

nwlandry:
Good point. Let's discuss more at our next meeting.

leotrs:
Now that I think of this, I am against it. The main benefit of having the tutorials be part of the main repo is that they are part of the main test suite. This means that whenever there's any change to the library, the tutorials are guaranteed to be updated too. If they were in a separate repo, we would have to manually check every time.

PR #353, opened 2023-05-05
Clean up classes

nwlandry:
* Moved IDDict to utilities to prepare for the DiHypergraph class
* Removed unnecessary arguments in the constructor of IDView


leotrs (approved):
Glad the awful View constructor was cleaned up!

PR #354, opened 2023-05-05
added empty_edges keyword

nwlandry:
A user reached out and asked if we could allow empty edges. I added the `empty_edges` as an optional keyword to opt-in to this behavior.

### DIFF
--- a/tests/classes/test_hypergraph.py
+++ b/tests/classes/test_hypergraph.py
@@ -466,7 +466,7 @@ def test_double_edge_swap(edgelist1):
 
     with pytest.raises(IDNotFound):
         H.double_edge_swap(8, 3, 0, 1)
-    
+
     H = xgi.Hypergraph(edgelist1)
     with pytest.raises(XGIError):
         H.double_edge_swap(6, 7, 2, 3)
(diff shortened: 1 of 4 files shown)


leotrs:
I believe at some point we allowed for empty edges and then decided not to support them. Is this the case? If so, does anyone remember why we decided not tu support them?

maximelucas:
> Is this the case?
 I don't remember honestly. Do we know what the use-case would be?

leotrs:
@nwlandry :)

nwlandry:
I originally wanted to include them to represent physical rooms without people in them. One of my colleagues wanted to represent groups that xould have 0 members. Lastly, there are several existing inconsistencies in XagI mentioned in #403.

leotrs:
Okay yeah that sounds like a legit use case. We could try to implement this and the wait and see what breaks.

One caveat is that we'd need a whole battery of tests (several dozens) to test for integration of different corners of the library with empty edges.

nwlandry:
This PR is stale and so I'm closing this. I will work on re-opening this in the future.

PR #355, opened 2023-05-08
Faster flag complex

maximelucas:
Adresses #264 

`flag_complex` and `random_flag_complex` were computing all maximal cliques and then letting `add_simplices_from` dealing with the `max_order` parameter. 
This means that a small network with large density was already very long to compute, e.g. 
```python
H = xgi.random_flag_complex(50, 0.7)
```

I changed this: only cliques up to `max_order` are computed and then passed to `add_simplices_from`.

```python
%timeit H = xgi.random_flag_complex(50, 0.6, seed=1)
# --> 22.6 s ± 198 ms # before
# --> 269 ms ± 690 µs # now
```

Roughly a 100x speedup, expected to be more for larger and denser graphs.

I also added some tests because the ones we had were testing consistency but not correctness.


maximelucas:
(The new commits are the last two, not sure why the others appear here, I think they're from a past PR)

xgi/generators/classic.py
```diff
@@ -154,3 +157,70 @@ def trivial_hypergraph(n=1, create_using=None, default=None):
...
+    H = xgi.Hypergraph()
+
+    nodes = range(N)
+    H.add_nodes_from(nodes)
+
+    if order is not None:
+        edges = combinations(nodes, order + 1)
+    elif max_order is not None:
+        start = 1 if include_singletons else 2
+        end = max_order + 1
```
leotrs, line 219: Maybe add `assert end > start`

xgi/generators/simplicial_complexes.py
```diff
@@ -130,26 +130,31 @@ def flag_complex(G, max_order=2, ps=None, seed=None):
...
+    if max_order is None:
+        cliques_to_add = list(nx.find_cliques(G)) # max cliques
+    else: # avoid adding many unnecessary redundant cliques
+        cliques_to_add = []
+        for clique in nx.enumerate_all_cliques(G): # sorted by size
+            if len(clique)<=max_order+1: 
+                cliques_to_add.append(clique)
+            else:
+                break # dont go over whole list if not necessary
+        cliques_to_add = cliques_to_add[N:] # remove singletons
```
leotrs, line 146:
```suggestion
            if 1 < len(clique) <= max_order+1: 
                cliques_to_add.append(clique)
            else:
                break # dont go over whole list if not necessary
```

xgi/generators/simplicial_complexes.py
```diff
@@ -277,13 +282,21 @@ def random_flag_complex(N, p, max_order=2, seed=None):
...
+    if max_order is None:
+        cliques = list(nx.find_cliques(G)) # max cliques
+    else: # avoid adding many unnecessary redundant cliques
+        cliques = []
+        for clique in nx.enumerate_all_cliques(G): # sorted by size
+            if len(clique)<=max_order+1: 
+                cliques.append(clique)
+            else:
+                break # dont go over whole list if not necessary
+        cliques = cliques[N:] # remove singletons
```
leotrs, line 296: All of this code is repeated from above - maybe extract into a helper function.

leotrs (commented):
:)

xgi/generators/simplicial_complexes.py
maximelucas: I thought of that too but it does not work because of the `break`.. first cliques are singletons and the results list if empty

xgi/generators/classic.py
maximelucas: Done. Can also be = because adding +1 to end in the range two lines down.

xgi/generators/simplicial_complexes.py
maximelucas: Replaced with an if at the top

xgi/generators/simplicial_complexes.py
maximelucas: Done

maximelucas:
Thanks Leo! Should all be addressed.

leotrs (approved):

PR #364, opened 2023-05-17
Update references.bib

danielskatz:


nwlandry:
Thanks! - one thing I noticed was that the XGI-DATA url needs to be updated to https://gitlab.com/complexgroupinteractions/xgi-data. I will update that once I merge this in.

nwlandry:
@danielskatz I updated that link so should be good to go. Thanks so much!

Issue #365, opened 2023-05-17
Add a shortest path algorithm


acombretrenouard:
We could add an algorithm that computes shortest paths between nodes.
The notion of path is not unique, but a starting point could be : "_An ordered list of nodes where two consecutive nodes belong to at least one hyperedge. All nodes are distincts._"

I wrote such a feature, based on method [`all_shortest_paths`](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.shortest_paths.generic.all_shortest_paths.html#all-shortest-paths) from `networkx`.

I can work on adding it to the basecode !

nwlandry:
Great idea, Antoine - this would be awesome to add to XGI!

nwlandry:
Fixed by #368.

PR #368, opened 2023-05-19
Add shortest path algorithm

acombretrenouard:
Related to issue #365.

I added a new file ./xgi/algorithms/shortest_path.py containing two functions (`single_source_shortest_path` and `shortest_path_length`).
There is also a new function in ./xgi/utils/utilities.py.

The functions are structured as in `networkx` and implement the Disjkstra algorithm.
Naming could be improved.

Corresponding test (for algorithms and utilities) and documentation are written. The changelog is up to date as well.
I successfully run `pytest`, `isort .`, `pylint xgi/ --disable all --enable W0611` and `black .`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,3 +104,4 @@ This library may not meet your needs and if this is this case, consider checking
 * [HyperGraphs.jl](https://github.com/lpmdiaz/HyperGraphs.jl): A package in Julia for representing, analyzing, and generating hypergraphs which may be oriented and weighted.
 * [hyperG](https://cran.r-project.org/package=HyperG): A package in R for storing and analyzing hypergraphs
 * [NetworkX](https://networkx.org/): A package in Python for representing, analyzing, and visualizing networks.
+![](![https://github.com/acombretrenouard/xgi.git](![https://github.com/acombretrenouard/xgi.git](https://github.com/acombretrenouard/xgi.git)))
--- a/README.md
+++ b/README.md
@@ -104,4 +104,3 @@ This library may not meet your needs and if this is this case, consider checking
 * [HyperGraphs.jl](https://github.com/lpmdiaz/HyperGraphs.jl): A package in Julia for representing, analyzing, and generating hypergraphs which may be oriented and weighted.
 * [hyperG](https://cran.r-project.org/package=HyperG): A package in R for storing and analyzing hypergraphs
 * [NetworkX](https://networkx.org/): A package in Python for representing, analyzing, and visualizing networks.
-![](![https://github.com/acombretrenouard/xgi.git](![https://github.com/acombretrenouard/xgi.git](https://github.com/acombretrenouard/xgi.git)))
--- a/xgi/algorithms/shortest_path.py
+++ b/xgi/algorithms/shortest_path.py
@@ -1,0 +1,103 @@
+"""Algorithms for computing shortest paths in a hypergraph."""
+
+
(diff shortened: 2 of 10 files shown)


README.md
```diff
@@ -104,3 +104,4 @@ This library may not meet your needs and if this is this case, consider checking
 * [HyperGraphs.jl](https://github.com/lpmdiaz/HyperGraphs.jl): A package in Julia for representing, analyzing, and generating hypergraphs which may be oriented and weighted.
 * [hyperG](https://cran.r-project.org/package=HyperG): A package in R for storing and analyzing hypergraphs
 * [NetworkX](https://networkx.org/): A package in Python for representing, analyzing, and visualizing networks.
+![](![https://github.com/acombretrenouard/xgi.git](![https://github.com/acombretrenouard/xgi.git](https://github.com/acombretrenouard/xgi.git)))
```
nwlandry, line 107: What's the reason for this line?

xgi/algorithms/shortest_path.py
nwlandry, line 1: Looks great. My biggest question would be whether it makes more sense to store the path lengths as a numpy array with a dict mapping node IDs to matrix row/columns.

nwlandry (changes requested):
Overall, looks great @acombretrenouard! Clean code, nice comments, and good testing! My biggest question is on the data structure of the path lengths. Thanks for the contribution!

README.md
acombretrenouard: Actually I don't know. I saw it appearing after the pull request.
Just delete it !

xgi/algorithms/shortest_path.py
acombretrenouard: I tried to reproduce the signature of [shortest_path_length](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.shortest_paths.generic.shortest_path_length.html#networkx.algorithms.shortest_paths.generic.shortest_path_length) in `networkx`.
In particular the returns are structured as explained here :
> If the source and target are both specified, return the length of the shortest path from the source to the target.
> If only the source is specified, return a dict keyed by target to the shortest path length from the source to that target.
> If only the target is specified, return a dict keyed by source to the shortest path length from that source to the target.
> If neither the source nor target are specified, return an iterator over (source, dictionary) where dictionary is keyed by target to shortest path length from source to that target.

I think returning a generator instead of an array is great for computing shortest paths on-the-fly and saving memory. I guess with time people figured out this was the best way (?).


Thanks for the feedback :)

acombretrenouard:
I restored `README.md` as requested by @nwlandry.

xgi/algorithms/shortest_path.py
nwlandry: Gotcha. Thanks for the explanation! I think I'm good with merging this. And I guess that it only makes sense to take advantage of the $d(n_1, n_2) = d(n_2, n_1)$ symmetry if a matrix is outputted. Nice work!

nwlandry:
Congrats on your first XGI contribution, @acombretrenouard ! Excited to start computing some shortest paths! 😁

maximelucas:
Thanks @acombretrenouard, and congrats !!

What are the details of this generalization to hypergraphs? Are you looking at paths between nodes through hyperedges of any size? When you compute the shortest path, a pairwise edge and a triangle would both count as "length 1" in the path?

Issue #358, opened 2023-05-11
New drawing style


nwlandry:
It would be cool to have a method for drawing hypergraphs/simplicial complexes in the style of Fig. 1 in [this paper](https://arxiv.org/abs/2104.05457).

maximelucas:
Agreed. 

The first step would be to have an `order` param in `draw` to only draw a given order?
Like I did in Fig 1c here https://journals.aps.org/prresearch/pdf/10.1103/PhysRevResearch.2.033410
At that (pre-XGI time), I did it this way with the function `plot_simplicial_pure()` available here 
https://github.com/maximelucas/plot_simplicial_complex/blob/main/draw_simplicial_complexes.ipynb

Then we can choose to have it look like a multilayer or not I guess.

thomasrobiglio:
That would be nice! @nwlandry this could be done using matplotlib 3D plotting right? Or you had something else in mind?

nwlandry:
Awesome! That's a great starting point.

Yeah, I was picturing using 3d plotting and Max's code as well as [this script](https://github.com/jkbren/matplotlib-multilayer-network) for plotting multilayer networks is a good place to start, I think.

thomasrobiglio:
If nobody is already working on it I might try to build it.

nwlandry:
Sounds great!!

PR #369, opened 2023-05-19
New drawing style

thomasrobiglio:
Related to https://github.com/xgi-org/xgi/issues/358, allows for new drawing style where hyperedges/simplices of different orders are draw on different layers of a 3-dimensional plot.

drawing_example

### DIFF
--- a/tests/drawing/test_draw.py
+++ b/tests/drawing/test_draw.py
@@ -183,3 +183,40 @@ def test_draw_hypergraph_hull(edgelist8):
     assert ax.collections[0].get_zorder() == 4  # nodes
 
     plt.close()
+
+
+def test_draw_multilayer(edgelist8):
+    # hypergraph
+    H = xgi.Hypergraph(edgelist8)
+    ax1 = xgi.draw_multilayer(H)
+    assert xgi.max_edge_order(H) * 4 - 1 == len(ax1.collections)
+    plt.close()
+
+    # max_order parameter
+    max_order = 2
+    ax2 = xgi.draw_multilayer(H, max_order=max_order)
+    assert max_order * 4 - 1 == len(ax2.collections)
+    plt.close()
+
+    # conn_lines parameter
+    ax3 = xgi.draw_multilayer(H, conn_lines=False)
+    assert xgi.max_edge_order(H) * 3 == len(ax3.collections)
+    plt.close()
+
+    # custom parameters
+    ax4 = xgi.draw_multilayer(
+        H,
+        pos=pos,
+        node_fc="red",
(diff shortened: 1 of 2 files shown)


nwlandry:
Thanks for this contribution, @thomasrobiglio! Looks great 🤩. My only question is, Is there a reason for using PolyCollection instead of scatter for the nodes? Also maybe you can open an issue on allowing 3d axes in `draw_nodes()`, `draw_hyperedges()`, etc? Overall, great work!

thomasrobiglio:
Thank you for the comment @nwlandry!

Actually I'm not sure I understand you question - I use `ax.scatter` to draw the nodes at line [1417](https://github.com/thomasrobiglio/xgi/blob/5563578f39b0e87b7d772e603e1af521e82745a8/xgi/drawing/draw.py#L1417) and use `Poly3DCollection` at line [1409](https://github.com/thomasrobiglio/xgi/blob/5563578f39b0e87b7d772e603e1af521e82745a8/xgi/drawing/draw.py#L1409) to create the hyperedges that are then added as a 3d collection.

About opening the issues we can for sure do that. I'm a bit afraid that using `draw_nodes()` or `draw_hyperedges()` in this function would require to modify a lot the original functions while it might be easier to open issues (or do that on this PR 😄) about expanding this function to allow the same personalisation possibilities we have in the other plotting functions (e.g. edges/nodes labels or nodes' size/color from stats ...). Let me know what you think about that!

nwlandry (approved):
Looks great to me! The comments that I was suggesting don't seem feasible.

tests/drawing/test_draw.py
```diff
@@ -183,3 +183,41 @@ def test_draw_hypergraph_hull(edgelist8):
     assert ax.collections[0].get_zorder() == 4  # nodes
 
     plt.close()
+
+
+def test_draw_multilayer(edgelist8):
```
leotrs, line 188: Seeing that this is mostly testing the number of things drawn (right?). If so, I'd suggest a more descriptive test name. Test function names can be super long, it doesn't matter.

thomasrobiglio, line 188:
```suggestion
def test_correct_number_of_collections_draw_multilayer(edgelist8):
```

@leotrs thanks for the comment, you mean something like this?

maximelucas:
It looks very pretty @thomasrobiglio thanks :) I even showed it to my class on Friday hehe.

tests/drawing/test_draw.py
leotrs: Perfect!

PR #371, opened 2023-05-21
Views should be read-only

nwlandry:
I found a bug where `H.edges.members()` creates a list with entries that are references to `H._edge` and so modifying `H.edges.members()` modifies the internal data structure. Same with `H.nodes.memberships()`. Also modified a few of the docstrings.


maximelucas:
Nice! Could you add a quick test for that?

nwlandry:
> Nice! Could you add a quick test for that?

Good point! Just did.

maximelucas (approved):

tests/classes/test_reportviews.py
```diff
@@ -167,6 +167,30 @@ def test_edge_members(edgelist3):
...
+    H.edges.members(0).add("a")
+    assert "a" not in H.edges.members(0)
+
+    e = H.edges.members()
+    e[0].add("a")
+    assert "a" not in H.edges.members(0)
+
+    e = H.edges.members(dtype=dict)
+    e[0].add("a")
+    assert "a" not in H.edges.members(0)
```
leotrs, line 180: I'd rather this be a separate test :)

PR #380, opened 2023-06-05
Complement of a hypergraph

acombretrenouard:
I wrote a function to complement a hypergraph.
It is computationally expensive, and to core function is a bit sketchy yet : I used strings and sets to enumerate all possible hyperedges, because I think it is quicker.

Let me know what you think !


acombretrenouard:
By the way : is it "a hypergraph" or "an hypergraph" ?
: )

xgi/classes/function.py
```diff
@@ -903,3 +905,71 @@ def subfaces(edges, order=None):
 
         faces += faces_to_add
     return faces
+
+
+def complement(H):
+    """Returns the complement of hypergraph H.
+
+    The complement of an hypergraph will have the same nodes (same indexes) and
```
maximelucas, line 913: it's "a" because the "h" is pronounced :) 
```suggestion
    The complement of a hypergraph will have the same nodes (same indexes) and
```

xgi/utils/utilities.py
```diff
@@ -210,3 +211,42 @@ def min_where(dicty, where):
...
+    if k < 0 or N < 0:
+        raise ValueError("binomial_sequence must be given positive integers.")
+
+    res = set()
+
+    if k == 0:
+        seq = str()
+        for i in range(N):
+            seq += "0"
+        res.add(seq)
```
leotrs, line 244:
```suggestion
        res = {"0" * i for i in range(N)}
```

xgi/classes/function.py
```diff
@@ -903,3 +905,71 @@ def subfaces(edges, order=None):
...
+    will contain every possible hyperedge linking these nodes, provided it is
+    not already present in the hypergraph and it is not larger than its largest
+    hyperedge.
+    Hyperedges of size one are taken into account.
```
maximelucas, line 917:
```suggestion
    The complement Hc of a hypergraph H has the same nodes (same indexes) as H and 
    contains every possible hyperedge linking these nodes, except those present in H. 
    The maximum hyperedge size in Hc is the same as in H.
    
    Singleton edges are taken into account.
```
```suggestion
    The complement of an hypergraph will have the same nodes (same indexes) and
    will contain every possible hyperedge linking these nodes, provided it is
    not already present in the hypergraph and it is not larger than its largest
    hyperedge.
    Hyperedges of size one are taken into account.
```

xgi/classes/function.py
```diff
@@ -903,3 +905,71 @@ def subfaces(edges, order=None):
...
+        # add to Hc the edges that are not already in H \
+        # note : as edges and possible_edge are of type set (i.e. hashtables) \
+        # doing set operations (set.difference) is fast
+        for edge in possible_edges.difference(edges):
+            node_set = set()
+            for i in range(N):
+                if edge[i] == "1":
+                    node_set.add(i)
+                else:  # edge[i] == '0'
+                    pass
```
leotrs, line 973:
```suggestion
            node_set = {i for i in range(N) if edge[i] == "1"}
```

xgi/classes/function.py
```diff
@@ -903,3 +905,71 @@ def subfaces(edges, order=None):
...
+            edge = str()
+            for _01 in row:
+                edge += str(_01)
+            edges.add(edge)
+
+        # list all possible edges of size <= max_edge_size
+        if M == 0:
+            max_edge_size = 0
+        else:  # there is at least one non-empty edge
+            max_edge_size = H.edges.order.max() + 1
```
maximelucas, line 953: I think this should be equivalent
```suggestion
        max_edge_size = xgi.max_edge_order(H) + 1
```

leotrs:
Hi @acombretrenouard, thanks for your work on this.

I don't think we need the binary strings at all. You can use [combinations](https://docs.python.org/3/library/itertools.html#itertools.combinations) from the stadard library, which does exactly the same thing (i.e. treat elements in a collection unique based on their position and not on their value). This is guaranteed to be faster.

xgi/classes/function.py
```diff
@@ -903,3 +905,71 @@ def subfaces(edges, order=None):
...
+
+        possible_edges = set()
+        for size in range(1, max_edge_size + 1):
+            possible_edges = possible_edges.union(binomial_sequence(size, N))
+
+        # instanciate an hypergraph an add every node (in case a node is in no hyperedge)
+        Hc = Hypergraph()
+        for i in range(N):
+            Hc.add_node(i)
```
maximelucas, line 962: This only works if H has nodes with names from 0 to N-1, but not for strings or other orders (docstring specify Hc has same nodes and indices).
What about
```suggestion
        Hc.add_nodes_from(H.nodes)
```

maximelucas:
Thanks Antoine! A few duplicate comments with Leo, we reviewed at the same time without knowing it.
I also agree with Leo that finding all possible hyperedges can be done in a simpler way:
- we already have a [`powerset`](https://xgi.readthedocs.io/en/stable/api/utils/xgi.utils.utilities.html#xgi.utils.utilities.powerset) function that does that
- or the [`complete_hypergraph(max_order=max_edge_size+1)`](https://xgi.readthedocs.io/en/stable/api/generators/xgi.generators.classic.html#xgi.generators.classic.complete_hypergraph) function that creates a hypergraph with all possibles edges.  Right now that functions creates nodes as range(N). If you want to use that, to preserve the original nodes, we should add an arguments `nodes` that specifies the nodes and overrides range(N) if not None.

maximelucas:
Btw @leotrs @nwlandry did we not have a few basic hypergraph operations?
I tried `H1 - H2` but it says "-" does not exist. Did we call it difference or something? In any case I couldn't find it in the docs, we should maybe make it more visible.

leotrs:
H1.nodes - H2.node

xgi/classes/function.py
acombretrenouard: I used your description, it is clearer !

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 
 ## latest
 
+* Added a function to complement an hypergraph (`complement`) [#367](https://github.com/xgi-org/xgi/issues/367) (@acombretrenouard).
```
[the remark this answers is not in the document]
acombretrenouard: Fixed.

xgi/utils/utilities.py
acombretrenouard: We just have to rite `res = {"0" * N}`.

xgi/classes/function.py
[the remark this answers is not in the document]
acombretrenouard: This works out !

xgi/classes/function.py
```diff
@@ -903,3 +905,71 @@ def subfaces(edges, order=None):
...
+        if M == 0:
+            max_edge_size = 0
+        else:  # there is at least one non-empty edge
+            max_edge_size = H.edges.order.max() + 1
+
+        possible_edges = set()
+        for size in range(1, max_edge_size + 1):
+            possible_edges = possible_edges.union(binomial_sequence(size, N))
+
+        # instanciate an hypergraph an add every node (in case a node is in no hyperedge)
```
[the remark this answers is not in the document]
acombretrenouard: Yup.

xgi/classes/function.py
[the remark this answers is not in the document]
acombretrenouard: I used the `index` parameter in `incidence_matrix` to retrieve the node dictionary.
Thanks !

acombretrenouard: This is fixed.

xgi/classes/function.py
acombretrenouard: On this : the function `max_edge_order` is defined above in the same python file (`xgi/classes/functions.py`). I could use it directly and write `max_edge_size = max_edge_order(H) + 1` (with the `xgi.` prefix).

Is there a problem with this ?
The function `complement` might not work if it was placed **above** `max_edge_order` in the file, right ?

acombretrenouard:
@nwlandry I think that what we are looking for is `powerset` instead of `combinations` (we want subsets of all sizes...).

The problem with `powerset` is that it has no cutoff for the maximal size of hyperedges/subsets. I could add it though, and it would be close to what `complete_hypergraph` does.

However, once we have the full list L of all _possible_ hyperedges, I don't see how we can pop those that are already present in the hypergraph we want to complement.
To me, we need to access the hyperedges in L knowing only their members (nodes), and not knowing their IDs.

acombretrenouard:
This thing with removing hyperedges from the list of all possible hyperedges to build Hc makes the use of the incidence matrix and/or a node ordering somehow necessary.
What do you think ?

xgi/classes/function.py
maximelucas: You're right, it's in the same file so you don't need the `xgi.`. The order of the functions in the file should not matter.

maximelucas:
> How precisely do you think I can check this ?
> Is there a way to access an hyperedge knowing only its nodes ?

You can do `H.edges.members()` to get the nodes inside each hyperedge.
Or `H.edges.members(dtype=dict)` (or equvalient but faster: `H._edge`) to have the same but keyed by hyperedge index. 

Is that what you were looking for?

acombretrenouard:
Hello @maximelucas @leotrs @nwlandry, I am back at it !
I made some changes to the function using `powerset`.

I tried to optimize the step of set difference (choosing which hyperedge to include, which to not include...).
This means that I have to _translate_ every hyperedge into a hashable value (I chose strings), which is a bit cumbersome.

Moreover, the function will output a hypergraph with the same node indexes. These indexes need not be rankable.


Tell me what you think !
In particular, dropping the condition on rankability will make the code simpler...

nwlandry:
Hi @acombretrenouard - thanks for all your work on this and sorry for the delayed reply! It seems like everything has been addressed - the only thing to do now is to fix the merge conflicts.

We have removed `function.py` for better discoverability in the docs, so can you move the complement function (and its corresponding unit tests) to be in generators.classic (the test_classic.py file for the tests) and modify the docs to match?

nwlandry:
I did a rebase (sorry for the mess 😄) and I fixed the outstanding issues.

PR #381, opened 2023-06-05
fixed math that was not represented properly.

nwlandry:
I converted math expression to Sphinx so that it rendered properly.


thomasrobiglio:
also `line_vector_centrality` and `clustering_coefficient`

nwlandry:
@thomasrobiglio - I couldn't locate the math error in `line_vector_centrality`. Could you send me a line number or snippet so I can locate it? Thanks!

PR #382, opened 2023-06-05
docs: fix math part 2

nwlandry:
take 2.


thomasrobiglio (commented):
thank you, I will go over the whole sphinx later today or tomorrow to check if I can find something else to be fixed.

Issue #376, opened 2023-05-30
Add support for importing directed hypergraphs from BIGG


nwlandry:
The BIGG database has lots of chemical reaction networks which can be expressed as a directed hypergraph: http://bigg.ucsd.edu/

It would be nice to have something similar to `load_xgi_data()` for these datasets.

PR #383, opened 2023-06-05
Import bigg data

nwlandry:
Addresses #376.


nwlandry:
I am creating a new PR as a branch of main, because the diff was ridiculously big and didn't reflect the small changes that I actually made. Originally I based this branch off of the DiHypergraphs PR because it needed the DiHypergraph functionality, but now that it's merged, I'll just do a branch off of `main`.

PR #384, opened 2023-06-05
Load bigg data

nwlandry:
This addresses #376. This PR

* Moves the json from url functions to the utilities and slims down the xgi_data module
* Adds a bigg_data module.

### DIFF
--- a/xgi/readwrite/bigg_data.py
+++ b/xgi/readwrite/bigg_data.py
@@ -90,8 +90,10 @@ def _bigg_to_dihypergraph(d_index, d_model):
     id = d_model["id"]
 
     DH["name"] = id
+    
     info = next((item for item in d_index["results"] if item["bigg_id"] == id), None)
     DH["organism"] = info["organism"]
+    
     for m in d_model["metabolites"]:
         DH.add_node(m["id"], name=m["name"])
 
--- a/xgi/readwrite/bigg_data.py
+++ b/xgi/readwrite/bigg_data.py
@@ -51,9 +51,10 @@ def load_bigg_data(
     index_url = "http://bigg.ucsd.edu/api/v2/models"
     base_url = "http://bigg.ucsd.edu/static/models/"
 
+    index_data = request_json_from_url(index_url)
+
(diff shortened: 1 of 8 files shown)


xgi/readwrite/bigg_data.py
```diff
@@ -0,0 +1,97 @@
...
+from ..utils import request_json_from_url, request_json_from_url_cached
+
+__all__ = ["load_bigg_data"]
+
+
+def load_bigg_data(
+    dataset=None,
+    cache=True,
+):
+    """Load a data set from the xgi-data repository or a local file.
```
leotrs, line 17:
```suggestion
    """Load a data set from the ???
```

xgi/readwrite/bigg_data.py
```diff
@@ -0,0 +1,97 @@
...
+
+    Parameters
+    ----------
+    dataset : str, default: None
+        Dataset name. Valid options are the top-level tags of the
+        index.json file in the xgi-data repository. If None, prints
+        the list of available datasets.
```
leotrs, line 24:
```suggestion
    dataset : str, default: None
        Dataset name. Valid options are the top-level tags of the
        index.json file in the ??? repository. If None, prints
        the list of available datasets.
```

xgi/readwrite/bigg_data.py
nwlandry: fixed.

leotrs (approved):

xgi/readwrite/bigg_data.py
```diff
@@ -0,0 +1,113 @@
...
+    DH = DiHypergraph()
+
+    id = d_model["id"]
+
+    DH["name"] = id
+
+    for d in d_index["results"]:
+        if d["bigg_id"] == id:
+            DH["organism"] = d["organism"]
+            break
```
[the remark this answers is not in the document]
nwlandry: It's guaranteed by the data repo format, but I wrote it in a more compact way, so that the "organism" attribute will always exist but can be none. Good catch, thanks!

Issue #375, opened 2023-05-30
Visualize dihypergraphs


nwlandry:
With the release of PR #372, it would be nice to visualize dihypergraphs as bipartite graphs, with squares representing edges and circles representing nodes. We already have layout algorithms for nodes and phantom nodes in the layout module.

PR #387, opened 2023-06-06
Drawing directed hypergraph

thomasrobiglio:
As discussed in #375. See below for the drawing. Added a simple test, maybe I can add more. Modified the DiHypergraph tutorial showing a simple example of the drawing function. 
example_draw_dihypergraph


maximelucas:
It looks pretty 😍 Is that what you had in mind @nwlandry ? Should we also add an option to draw the hyperedge (like a triangle for example) below it like in the other drawing functions?

thomasrobiglio:
This should be easily done with the draw_hyperedges function but I'm afraid then it gets a bit ambiguous to have both the triangles and the squared markers representing the hyperedges

nwlandry:
This looks great, @thomasrobiglio!! My only suggestion would be to add a check that the input is indeed a directed hypergraph.

maximelucas:
Also why is one much lighter? Because it's a smaller hyperedge?

thomasrobiglio:
Yes, if not specified the coloring is like the standard draw in which the color depends on the edge size

thomasrobiglio:
Just noticed that edges of size 2 are treated as the other sizes so the function will still draw a square representing the edge... so we want to keep that or put simple arrows for edges of size 2?

leotrs:
Looking great! One thing - what do y'all think of an option to remove the squares? If all the lines are drawn up to the center of the square, then removing the squares will allow us to see the lines joining and forking again, which may be cool.

Alternatively, an option to change the size of the squares independently of the size of the circles).

maximelucas:
Agreed with Leo. Also I don't see an option to add node/hyperedge labels, right? Could we add it like in the other viz functions?

Also the triangle blue looks so pale by default, I'll raise an issue to see if we can improve the default coloring.

leotrs (approved):

leotrs (approved):

nwlandry:
Great work, @thomasrobiglio!

PR #392, opened 2023-06-08
update index.rst to have dihypergraph in quick references

thomasrobiglio:
I noticed that the DH does not appear in the quick references, I think this might fix it (but I'm not familiar at all with how Sphinx works, etc. so pls check it).

### DIFF
--- a/docs/source/api/classes/xgi.classes.dihypergraph.DiHypergraph.rst
+++ b/docs/source/api/classes/xgi.classes.dihypergraph.DiHypergraph.rst
@@ -4,8 +4,7 @@
 .. currentmodule:: xgi.classes.dihypergraph
 
 .. warning::
-    This is an experimental module. It has been thoroughly tested in development
-    but has not been used in research. It is not yet stable and may change in the future.
+    This is an experimental module. It is not yet stable and may change in the future.
 .. autoclass:: DiHypergraph
    :show-inheritance:
    :members:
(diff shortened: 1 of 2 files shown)


maximelucas:
I saw it too but not sure we want it there since it's still experimental?

leotrs:
Good point Max

thomasrobiglio:
Ah right, I did not think about that

leotrs:
@thomasrobiglio you were right - you can check the documentation built from this PR (instead of from the main branch) by clicking "docs/readthedocs.orx:xgi" under the "all checks have passed" button here in this PR.

nwlandry (approved):
I think this is a good idea. While this is still an experimental feature, perhaps you can just add (*Experimental feature!*) or something like that after the text, so that users can still easily find it if they want to use it.

thomasrobiglio:
Now it will appear like this... but maybe this is not what you had in mind @nwlandry (since also there was already the warning in the class docs). Then of course we can change the text, this one was co-authored by copilot :)

Screenshot 2023-06-09 alle 10 22 44

leotrs:
I like the warning admonition. I would delete the middle sentence and leave the first and last.

nwlandry:
Looks great! I agree with Leo on deleting the middle sentence and then I think you're good to merge, @thomasrobiglio. Thanks so much!

PR #397, opened 2023-06-12
Convert from simplex dict to SC

### LINKED ISSUE #330: Add the ability for the SimplicialComplex class constructor to handle hypergraphs

thomasrobiglio:
This should resolve #330 (the ability for the SC constructor to handle a HG was already implemented, this PR is about https://github.com/xgi-org/xgi/issues/330#issuecomment-1503337363).

I added a new function `from_simplex_dict` that takes a dict in the form: `{edge_id: members, ...}` and returns a SC. If the subfaces of existing simplices are not given in the dict then the function adds them with integer IDs.

Screenshot 2023-06-12 alle 19 48 15


leotrs (approved):

nwlandry (commented):
Great work, @thomasrobiglio! I just tried it out and it works well! Good to squash and merge!

Issue #398, opened 2023-06-13
xgi.Hypergraph(SC) not working anymore?


maximelucas:
Running 

```python
S = xgi.random_simplicial_complex(10, [0.05, 0.01], seed=1) # non-empty SC
H = xgi.Hypergraph(S) # convert SC to HG
```

yields an empty Hypergraph: no nodes and no edges. 

I'm pretty sure this used to work?

thomasrobiglio:
I think I've found the bug (coming from one of my first PR, the one about the convert function), I will open the PR

PR #345, opened 2023-04-29
Attempt at fixing convert

thomasrobiglio:
Hi all, 
this is my attempt at addressing issue https://github.com/xgi-org/xgi/issues/327 and some other troubles I've found using the functions in the `convert.py` file.

Now it should be possible to:
1. Convert SC to Hyperg. and viceversa (using `from_max_simplices` and an analogous criterion when going from the other way around i.e. hyperdges -> maximal faces then i had all subfaces)
2. Convert from list, dict, and pd.DataFrame to Hypergraph as was possible before but without the issue of not returning anything. In this case if a pass an already existing Hypergraph instance then I have the in-place modification of the instance, else the function returns a new hypergraph()
3. Same as (2) for list, pd.Dataframe -> SimplicialComplex()

Hopefully this is what you had in mind when discussing https://github.com/xgi-org/xgi/issues/327 and this can be a useful contribution.

PR #399, opened 2023-06-13
fix: HG using SC as data

thomasrobiglio:
Related to #398. The bug was coming from the condition I had set in the `convert_to_hypergraph()` in PR #345 (sry 😢). Now it should be fixed. 

There are two issues related to this:

1. The constructors `xgi.Hypergraph(SC)` and `xgi.SimplicialComplex(H)` are *not* tested! 
2. How it is implemented now if I call the cover `convert_to_hypergraph()` as a standalone with a SC as an object I get back the HG constructed from the maximal simplices of the SC, if instead, I call it via the constructor I get a HG in which every simplex in the SC becomes an edge. Is this ok or we want to change it?

For me it is intuitive to have the standalone convert give only the maximal simplices but I guess this is debatable.
Screenshot 2023-06-13 alle 12 24 22

### DIFF
--- a/tests/classes/test_simplicialcomplex.py
+++ b/tests/classes/test_simplicialcomplex.py
@@ -27,6 +27,13 @@ def test_constructor(edgelist5, dict5, incidence5, dataframe5):
     with pytest.raises(XGIError):
         xgi.SimplicialComplex(1)
 
+    H = xgi.Hypergraph(edgelist5)
+    S_h = xgi.SimplicialComplex(H)
+
+    assert set(S_h.nodes) == set(H.nodes) == set(S_list.nodes)
+    assert set(S_h.edges) == set(S_list.edges)
+    
+
 
 def test_string():
     S1 = xgi.SimplicialComplex()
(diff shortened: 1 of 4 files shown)


tests/classes/test_simplicialcomplex.py
```diff
@@ -27,6 +27,12 @@ def test_constructor(edgelist5, dict5, incidence5, dataframe5):
     with pytest.raises(XGIError):
         xgi.SimplicialComplex(1)
 
+    H = xgi.Hypergraph(edgelist5)
+    S_h = xgi.SimplicialComplex(H)
+
+    assert set(S_h.nodes) == set(H.nodes) == set(S_list.nodes)
+    assert set(S_h.edges) == set(S_list.edges)
```
maximelucas, line 34: This is only checking edge IDs, so it's almost like checking the number of edges. 
Can you check their members by comparing `H.edges.members()` or `H._edge`?

tests/classes/test_simplicialcomplex.py
```diff
@@ -27,6 +27,12 @@ def test_constructor(edgelist5, dict5, incidence5, dataframe5):
...
+
```
maximelucas, line 35:
```suggestion
assert set(H.edges.members()) <= set(S_h.edges.members()) # check it's a subset
```

maximelucas (approved):

PR #400, opened 2023-06-13
included functions from publication

maximelucas:
I added two functions I made here: https://github.com/maximelucas/HOI_shape_sync_differently/blob/main/utils.py
- `shuffle_hyperedges(H, order, p)`: shuffle hyperedges with a given probability
- `node_swap(H, nid1, nid2, order)`: swap node membership in edges of given order
- added tests for both

I put them in a new file `generators/randomizing.py`.

### DIFF
--- a/docs/source/api/generators.rst
+++ b/docs/source/api/generators.rst
@@ -13,4 +13,5 @@ generators package
    ~xgi.generators.random
    ~xgi.generators.uniform
    ~xgi.generators.simplicial_complexes
+   ~xgi.generators.randomizing
    
--- a/docs/source/api/generators/xgi.generators.random.rst
+++ b/docs/source/api/generators/xgi.generators.random.rst
@@ -1,13 +1,10 @@
-xgi.generators.random
-=====================
+xgi.generators.randomizing
+==========================
 
-.. currentmodule:: xgi.generators.random
+.. currentmodule:: xgi.generators.randomizing
 
-.. automodule:: xgi.generators.random
+.. automodule:: xgi.generators.randomizing
    
     .. rubric:: Functions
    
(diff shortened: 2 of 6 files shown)


maximelucas:
Finally had a better look.

- **functionality**: they are similar but different. `double_edge_swap()` picks two nodes and two edges, and swaps the membership of those two nodes in those two edges. `node_swap()` also picks two nodes, but then swaps their membership in all edges that they are part of, not just two. 

- **naming:** It's your function, but looping over all the edges that contain them. Because of that, it felt more natural for me to call it "node swap" rather than edge swap. Your function is more "symmetric" wrt to nodes and edges, I guess it could also be called "double_node_swap"? 

- **location:** to me in might make more sense conceptually to put both of them in `randomizing.py`? I understand that you see the swap as a fundamental operation. `hypergraph.py` now contains almost only methods to remove/add nodes/edges/attributes, "creating structure". In my mind, the two swap functions take an existing structure, and modify it by switching things around. In my case, I wrote `node_swap()` with randomization in mind, to degree sequences intact but changing degree correlations between orders. Maybe yours is less random as you pick the edges to use.

- **implementation:** `node_swap()` is clearly based on your implementation of  `double_edge_swap()` (variable names, structure, etc.). I added a few initial checks to make sure that the queried nodes exist etc. Conceptually it could be implemented as a loop where each iteration uses your function in practice I remember I thought that would be more of a hassle than anything. 

Happy to change anything, let me know what you think.

nwlandry:
@maximelucas: I'm good with you merging it, thanks for the additional clarification. We can probably revisit whether `double_edge_swap` should be in the `Hypergraph` class.

PR #401, opened 2023-06-13
Update docs

nwlandry:
* Moved the list of contributors from the readthedocs, license, setup, etc. into a dedicated file so that when contributors join or leave, we only update a single file.
* Renamed CONTRIBUTING.md
* Eventually, it would be good to recognize people even if they don't submit code according to the all-contributors standard (in a more visually pleasing format).


maximelucas:
Looks good, thanks Nich! One single page makes sense, networkx have the same:
https://networkx.org/documentation/stable/developer/about_us.html

Agreed on later finding a good way to recognise all contributions types.

PR #402, opened 2023-06-14
Fixed the _color_arg_to_dict function

### LINKED ISSUE #389: viz: node_fc_cmap does not work if node_fc is not a NodeStat
Let's say we plot a simple hypergraph

```python
edges = [[1, 2], [1, 2, 3], [2, 3], [3, 4, 5, 6]]
H = xgi.Hypergraph(edges)

# plot node size ~ degree, with shades of green
xgi.draw(H, node_fc=H.nodes.degree, node_fc_cmap=mpl.colormaps["Greens"])
```

Here `H.nodes.degree` is a `NodeStat`, it works:
Screenshot 2023-06-07 at 12 01 21

But if we replace it by a list with the same values  `H.nodes.degree.aslist()`, the cmap is discarded and back to its default:
Screenshot 2023-06-07 at 12 04 05

I guess we may have a similar problem with related variables:  `node_ec_cmap`, `dyad_color_cmap`,  `edge_fc_cmap`.

nwlandry:
This fixes #389.

### DIFF
--- a/xgi/drawing/draw.py
+++ b/xgi/drawing/draw.py
@@ -743,11 +743,10 @@ def _color_arg_to_dict(arg, ids, cmap):
     arg : There are several different valid arguments. They are:
 
         Single color values
-        
+
         * str
         * 3- or 4-tuple
-        * numpy array
-        
+
         Iterable of colors
 
         * numpy array of color values as described above
@@ -755,9 +754,10 @@ def _color_arg_to_dict(arg, ids, cmap):
         * dict of color values as described above
 
         Iterable of values
-        
+
         * list of floats
         * dict of floats
+        * numpy array of floats
 
         Stats
 
@@ -787,9 +787,7 @@ def _color_arg_to_dict(arg, ids, cmap):
     """
 
     # single argument. Must be a string or a tuple of floats
-    if isinstance(arg, str) or (
-        isinstance(arg, (tuple, ndarray)) and isinstance(arg[0], float)
-    ):
+    if isinstance(arg, str) or (isinstance(arg, tuple) and isinstance(arg[0], float)):
         return {id: arg for id in ids}
 
     # Iterables of colors. The values of these iterables must strings or tuples. As of now,
@@ -834,7 +832,7 @@ def _color_arg_to_dict(arg, ids, cmap):
                     if id in ids
                 }
 
-            if isinstance(arg, list) and isinstance(arg[0], (int, float)):
+            if isinstance(arg, (list, ndarray)) and isinstance(arg[0], (int, float)):
                 vals = np.interp(arg, [np.min(arg), np.max(arg)], [minval, maxval])
             else:
                 raise TypeError("Argument must be an iterable of floats.")
--- a/xgi/drawing/draw.py
+++ b/xgi/drawing/draw.py
@@ -16,7 +16,7 @@ from scipy.spatial import ConvexHull
 from .. import convert
 from ..classes import DiHypergraph, Hypergraph, SimplicialComplex, max_edge_order
 from ..exception import XGIError
-from ..stats import EdgeStat, NodeStat
+from ..stats import IDStat
 from .layout import _augmented_projection, barycenter_spring_layout
 
 __all__ = [
@@ -644,6 +644,7 @@ def draw_simplices(
                 color=dyad_color[id],
                 lw=dyad_lw[id],
                 zorder=max_order - 1,
+                cmap=settings["dyad"],
             )
             ax.add_line(line)
         else:
@@ -722,7 +723,7 @@ def _scalar_arg_to_dict(arg, ids, min_val, max_val):
         return {id: arg[id] for id in arg if id in ids}
     elif type(arg) in [int, float]:
         return {id: arg for id in ids}
-    elif isinstance(arg, NodeStat) or isinstance(arg, EdgeStat):
+    elif isinstance(arg, IDStat):
         vals = np.interp(arg.asnumpy(), [arg.min(), arg.max()], [min_val, max_val])
         return dict(zip(ids, vals))
     elif isinstance(arg, Iterable):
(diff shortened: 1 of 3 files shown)


thomasrobiglio:
The only error in the test that I was able to reproduce is the one at [this line](https://github.com/xgi-org/xgi/actions/runs/5269655064/jobs/9528091131#step:5:220) (which is present in the [branch](https://github.com/xgi-org/xgi/blob/b1d891a45311878cbed6fca9a86ebb115de63069/xgi/drawing/draw.py#L647) but not present in the [main](https://github.com/xgi-org/xgi/blob/8d2960438a238ee504de78858f752ab73df2a3f5/xgi/drawing/draw.py#L646) 🤯) I guess than the others are then  'consequences' (e.g. the test function does not reach `plt.close()` maybe). Is this possible?

nwlandry:
MVP!!!!! Thanks so much, @thomasrobiglio. That was the issue. I'll add a couple of tests and then undraft this PR.

thomasrobiglio (approved):
looks good :)

xgi/drawing/draw.py
```diff
@@ -739,7 +740,30 @@ def _color_arg_to_dict(arg, ids, cmap):
...
+        Single color values
+
+        * str
+        * 3- or 4-tuple
+
+        Iterable of colors
+
+        * numpy array of color values as described above
+        * list of color values as described above
+        * dict of color values as described above
```
maximelucas, line 754: for brevity (I would even maybe put numpy array and list on the same line): 
```suggestion
        Iterable of colors (each specified as above)

        * numpy array
        * list
        * dict
```
one question: in the `dict` case, I guess values are used? Keys should refer to node/edge IDs?

xgi/drawing/draw.py
```diff
@@ -739,7 +740,30 @@ def _color_arg_to_dict(arg, ids, cmap):
...
+
+        Iterable of values
+
+        * list of floats
+        * dict of floats
+        * numpy array of floats
```
maximelucas, line 760: Also works with int?
For brevity and consistency:
```suggestion
        Iterable of numerical values (floats)

        * numpy array
        * list
        * dict
```

xgi/drawing/draw.py
```diff
@@ -739,7 +740,30 @@ def _color_arg_to_dict(arg, ids, cmap):
 
     Parameters
     ----------
-    arg : str, dict, iterable, or NodeStat/EdgeStat
+    arg : There are several different valid arguments. They are:
```
maximelucas, line 743: Should we rename the variable `arg`? It does not feel very intuitive. `color`?
For brevity:
```suggestion
    arg : Several formats are accepted:
```

xgi/drawing/draw.py
```diff
@@ -756,21 +757,55 @@ def _color_arg_to_dict(arg, ids, cmap):
...
+    if isinstance(arg, (Iterable, IDStat)):
+        # set max and min of interpolation based on color map
         if isinstance(cmap, ListedColormap):
-            vals = np.interp(arg.asnumpy(), [arg.min(), arg.max()], [0, cmap.N])
+            minval = 0
+            maxval = cmap.N
         elif isinstance(cmap, LinearSegmentedColormap):
-            vals = np.interp(arg.asnumpy(), [arg.min(), arg.max()], [0.1, 0.9])
+            minval = 0.1
+            maxval = 0.9
```
maximelucas, line 813: how are these decided?

xgi/drawing/draw.py
```diff
@@ -756,21 +757,55 @@ def _color_arg_to_dict(arg, ids, cmap):
...
+                    for v, id in zip(vals, arg.keys())
+                    if id in ids
+                }
+
+            if isinstance(arg, (list, np.ndarray)) and isinstance(arg[0], (int, float)):
+                vals = np.interp(arg, [np.min(arg), np.max(arg)], [minval, maxval])
+            else:
+                raise TypeError("Argument must be an iterable of floats.")
+
         return {id: np.array(cmap(vals[i])).reshape(1, -1) for i, id in enumerate(ids)}
```
maximelucas, line 840: this line could be directly at the end of `if isinstance(arg, IDStat):` after line 787?

xgi/drawing/draw.py
nwlandry: I'll rename it to color_arg since it's not always a color

xgi/drawing/draw.py
nwlandry: Yep! I added a little example of the format.

xgi/drawing/draw.py
nwlandry: These values are so that the colors don't fade into the background but I suppose we could add them as an arg in the future?

xgi/drawing/draw.py
nwlandry: good catch! yep, I will add that to the docstring.

xgi/drawing/draw.py
```diff
@@ -755,22 +779,65 @@ def _color_arg_to_dict(arg, ids, cmap):
     ------
     TypeError
         If a string, dict, iterable, or NodeStat/EdgeStat is not passed.
+
+    Notes
+    -----
+    For the iterable of values, we do not accept numpy arrays or tuples,
+    because there is the potential for ambiguity.
```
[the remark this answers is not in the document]
nwlandry: Basically the ambiguity is that a color can be specified as a tuple of floats (R,G,B) or (R,G,B,alpha) and if you had a hypergraph with 3 or 4 nodes, it would break if you also accepted tuples of floats in the case where we map float values to a color map.

nwlandry:
> Nice good job!
> 
> Just a few questions to make sure the docs are clear. Also, I found the "Returns" a bit confusing ("An ID: attribute dictionary."), I wasn't sure what was returned. Maybe "An {ID: attribute} dictionary." or "A dictionary keyed by node/edge IDs and values as attributes. Also attribute->arg/color since we never use "attribute" in the function I think. Same for _scalar_to_dict.
> 
> Btw: do you think we have similar problems (cmap/sizes not used from list) in _scalar_to_dict?

I updated `_scalar_arg_to_dict()` while I was at it.

Issue #362, opened 2023-05-16
Issue in the stats notebook


nwlandry:
In the Stats tutorial notebook, it says that "The order of the degrees in the output respects the order of the nodes in the underlying
view", but when I compute
```
H.nodes([3, 4, 5]).degree.aslist(), H.nodes([4, 5, 3]).degree.aslist()
```
I get
```
([3, 2, 2], [3, 2, 2])
```

leotrs:
Ah damn this used to work...

leotrs:
Accidentally pushed a fix to this directly to main instead of via a PR. Welp. See commit [here](https://github.com/xgi-org/xgi/commit/efbcdab4dd7debacdd559efced25c8efc6ceca2c).

PR #406, opened 2023-06-19
Fix IDView.ids type

leotrs:
Currently, H.edges.ids sometimes returns a dict and sometimes returns a set:

```python3
>>> import xgi
>>> H  = xgi.Hypergraph([{1, 2, 3}, {4}, {5, 6}, {6, 7, 8}])
>>> H.edges.ids
{0: {1, 2, 3}, 1: {4}, 2: {5, 6}, 3: {6, 7, 8}}
>>> H.edges({0, 1, 2}).ids
{0, 1, 2}
```
This is terrible. Furthermore, the dict version can be obtained via members:
```python3
>>> H.edges.members(dtype=dict)
{0: {1, 2, 3}, 1: {4}, 2: {5, 6}, 3: {6, 7, 8}}
```

This PR makes it so that it always returns a set.
```python3
>>> import xgi
>>> H  = xgi.Hypergraph([{1, 2, 3}, {4}, {5, 6}, {6, 7, 8}])
>>> H.edges.ids
{0, 1, 2, 3}
```

I also added a couple of tests and changed an old test that relied on `ids` returning a dict (yuck).

### DIFF
--- a/tests/classes/test_reportviews.py
+++ b/tests/classes/test_reportviews.py
@@ -331,3 +331,16 @@ def test_maximal(edgelist5, edgelist8):
     # try strict=True
     m = H.edges.maximal(strict=True)
     assert set(m) == {5}
+
+
+def test_ids_type(edgelist5):
+    H = xgi.Hypergraph(edgelist5)
+    assert H.edges.ids == {0, 1, 2, 3}
+    assert H.edges([0, 1, 2, 3]).ids == {0, 1, 2, 3}
+    assert H.edges({0, 1, 2, 3}).ids == {0, 1, 2, 3}
+
+
+def test_ids_are_immutable(edgelist5):
+    H = xgi.Hypergraph(edgelist5)
+    H.edges.ids.add(42)
+    assert H.edges.ids == {0, 1, 2, 3}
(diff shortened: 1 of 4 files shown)


nwlandry:
Nice work! Can you comment on Issue #362? For now maybe you could remove the comment in the notebook about order being preserved?

nwlandry:
Okay, one more thing. I noticed that in the constructor of `IDView`, `self._id` doesn't have to be a set (line 91). Can you fix that before merging?

leotrs:
Could you be more specific? [L91](https://github.com/xgi-org/xgi/blob/24b6c736e603fe9d9de0983640075bea3dc93731/xgi/classes/reportviews.py#L91) and subsequent lines guarantee that `self._ids` is always set, regardless of the value of `ids`.

nwlandry:
> Could you be more specific? [L91](https://github.com/xgi-org/xgi/blob/24b6c736e603fe9d9de0983640075bea3dc93731/xgi/classes/reportviews.py#L91) and subsequent lines guarantee that `self._ids` is always set, regardless of the value of `ids`.

In this PR, I don't believe that `self._ids` is always a set if you construct an IDView with a) `ids=None` or b) (for example) `ids=[1, 2, 3]`. I would do something like
```{python}
self._ids = set(self._id_dict) if ids is None else self._ids = set(ids)
```
on L91, to enforce this.

leotrs:
Perhaps we're not seeing the same code? On this PR, L91 and subsequent lines look like this:
```python3
        if ids is None:
            self._ids = self._id_dict
        else:
            self._ids = ids
```

nwlandry:
I tried out a few things and I was mistaken. I take back my prior comment :)

Issue #356, opened 2023-05-10
move content of functions.py to more natural place


maximelucas:
We have many useful functions in `classes/function.py`. But they're location in the `classes` folder is probably not very natural: one would go there to look for them. We should decided where would be a better place. Maybe different functions should be move to different places.

nwlandry:
I totally agree with you. Some initial ideas:

* Rename the "classes" module to "core", rename "reportviews" to "views", and rename "hypergraphviews" to "globalviews"
* set_node_attributes", "get_node_attributes", "set_edge_attributes", "get_edge_attributes" can all go in IDView
* "freeze", "is_frozen", "convert_labels_to_integers", "density", and "is_empty" can go in Hypergraph
* Delete "create_empty_copy"

leotrs:
Agree with all Nich said except with moving more functions to Hypergraph. Right now, the Hypergraph class is already big and I made some changes a while ago to slim it down. In my head, the Hypergraph class provides functionality to build a hypergraph and store its node/edge/network attributes, and all functionality to compute statistics about it goes in the stats module or in standalone functions. I feel kinda strongly about this :)

PR #412, opened 2023-07-08
Refactor core data structures

nwlandry:
Fixes issue #356.

* I renamed the `classes` module to `core`, renamed `reportviews` to `views`, renamed `direportviews` to `diviews`, and renamed `hypergraphviews` to `globalviews`.
* I deleted `create_empty_copy`, `is_empty` because I don't think that either of these functions add much functionality.
* I was thinking originally that `set_node_attributes`, `get_node_attributes`, `set_edge_attributes`,  and `get_edge_attributes` could all go in IDView, but then I realize that views don't modify the structure of the hypergraph.
* I moved the `subfaces` function to the utils module.
* I moved `num_edges_order`, `max_edge_order`, `is_possible_order`, `is_uniform`, `edge_neighborhood`, `degree_counts`, `degree_histogram`, `unique_edge_sizes`, `density`, and `incidence_density` to a module inside the algorithms module titled `properties`.
* I moved `freeze` and `is_frozen` into the `Hypergraph`, `SimplicialComplex` and `DiHypergraph` classes.
* I moved `convert_labels_to_integers` to the utilities file.
* I deleted the `get_edge_attributes` and `get_node_attributes` because the NodeStats and EdgeStats classes handle this better.
* I made `set_edge_attributes` and `set_node_attributes` methods of the `Hypergraph` and `DiHypergraph` classes.


leotrs (approved):

maximelucas:
Looks good, thanks!
Only question: what's the reason for renaming `classes` to `core`?

nwlandry:
> Looks good, thanks! Only question: what's the reason for renaming `classes` to `core`?

When I started this PR, I wasn't sure whether I would be able to move all the functions in function.py out of the classes module and in that case, `classes` felt like a misnomer. But, in the end, I was able to move all the functions in function.py. So I'm happy to change it back to classes. Let me know what you'd like to do.

maximelucas:
That makes sense. Originally I was thinking `classes` would be slightly better because more specific. 
But there are some classes (the Stats) that are not in that folder, so maybe `core` is better.
I'm probably fine with both.

Btw, @leotrs the stats classes are in an __init__.py, any chance we could move them to more normal file?

nwlandry:
> That makes sense. Originally I was thinking `classes` would be slightly better because more specific. But there are some classes (the Stats) that are not in that folder, so maybe `core` is better. I'm probably fine with both.
> 
> Btw, @leotrs the stats classes are in an **init**.py, any chance we could move them to more normal file?

Okay, I think I will keep it as `core` but we can easily change it back in the future, since it's not user-facing.

nwlandry:
Thanks for the reviews, @maximelucas and @leotrs! 🥳

maximelucas:
Thank you, this was well needed!

PR #423, opened 2023-07-18
Reorganize convert

nwlandry:
Made `convert.py` a package, with each module dedicated to a particular type of conversion.

Issue #413, opened 2023-07-13
Add line graph generalizations


tlarock:
I would like to propose the addition of functionality for computing a few additional line graph structures. We try to justify why these structures might be of interest in [this preprint](https://arxiv.org/abs/2307.04613). Opening an issue to discuss what this might look like before I implement. Happy to hear thoughts on this, or to work on a PR independently and propose a concrete implementation :)

These are what I have in mind:
* Weighted intersection graph: Edges exist between any two hyperedges if they share at least 1 node, with weight corresponding to the size of the intersection. 
* Encapsulation DAG: Edges exsist between a larger hyperedge and a smaller hyperedge (or potentially vice-versa) if the smaller hyperedge is a subset of the larger hyperedge.

For the intersection graph, we could also add a normalization option, where the intersection size is normalized by the minimum size of the two hyperedges. It may also be useful to add edge filtering based on edge sizes. For example, we could add an option to include only edges between adjacent hyperedge sizes in the encapsulation DAG, or an option to return the transitively reduced DAG, which are both structures we study in the paper.

As of now, the line graph signature looks like `to_line_graph(H, s=1)`, where `s` is a parameter that controls the minimum intersection size. I think there are two reasonable options for adding the structures I propose:
1. (i) add an option for including (potentially normalized) weights, as in `to_line_graph(H, s=1, weights=False, normalize_weights=False)`. The return value would be a `networkx.Graph` object, with `weight` edge attributes if `weights=True`. And (ii) add a separate function `to_encapsulation_DAG(H, max_level_diff=float('inf'))` which computes and returns an encapsulation DAG. When the parameter `max_level_diff`is less than float infinity, we only include edges between hyperedges with difference in hyperedge size less than `max_level_diff`. This should allow for filtering the edges as I described above. The return value would be a `networkx.DiGraph`.
3. Instead of a separate function for the encapsulation DAG, add a `graph_type` or `relation_type` or `relation_function` option to `to_line_graph` that could be set to something like `"subset"` or `"encapsulation"`, and pass DAG parameters as well. We could also do some type checking to allow for different options without adding so many paramters, but I worry that this might make the function a bit too overloaded/confusing to use. Also, the `weights` semantics are not necessarily the same for the DAG and the intersection graph.

[...]

PR #425, opened 2023-07-24
Move `to_line_graph` from `convert.py` and add weight options

tlarock:
First in a series of PRs relating to #413. It does 2 main things:

1. Moves `to_line_graph` from `convert.py` into a new file called `line_graphs.py`, modifies files that refer to `convert.to_line_graph`, and adds `.line_graphs` to `__init__.py`
3. Adds `weighted` and `normalize_weight` options to `to_line_graph`

I decided to keep the default behavior as returning an unweighted line graph for now. So when `weighted=False`, the function behaves exactly the same as before (and `normalize_weights` has no effect). If `weighted=True`, the returned graph also has a `weight` attribute corresponding to the size of the intersection between the hyperedges. This adds no further computation, since the interesection size is needed anyway, but adds an integer attribute to every edge. If `weighted=True` and `normalize_weights=True`, the `weight` attribute is normalized by the minimum size between the two hyperedges (meaning it is a float between 0.0 and 1.0).

Instead of the two boolean options, we could either:
1. Return absolute weights by default. In this case we wouldn't need `weighted`, just `normalize_weights`.
2. Use a string input for normalization, something like `weight_type` with defualt `'unweighted'` and other options `'absolute'` and `'normalized'`.

Will also need to check again that everything works once #423 is merged.

### DIFF
--- a/xgi/__init__.py
+++ b/xgi/__init__.py
@@ -16,6 +16,7 @@ from .utils import *
 from .classes import *
 from .algorithms import *
 from .convert import *
+from .line_graphs import *
 from .drawing import *
 from .dynamics import *
 from .generators import *
--- a/xgi/algorithms/centrality.py
+++ b/xgi/algorithms/centrality.py
@@ -8,7 +8,7 @@ from numpy.linalg import norm
 from scipy.sparse.linalg import eigsh
 
 from ..classes import convert_labels_to_integers, is_uniform
-from ..convert import to_line_graph
+from ..line_graphs import to_line_graph
 from ..exception import XGIError
 from ..linalg import clique_motif_matrix, incidence_matrix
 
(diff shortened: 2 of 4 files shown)


nwlandry:
@tlarock I finished merging the changes to the convert module - let me know if you need anything from me when resolving merge conflicts.

leotrs:
My 2¢: the default should return no weights (less memory, and dare I assume that most people won't use the weights) AND use a string parameter with default None (instead of "unweighted") and possible options "adsolute" and "normalized".

tlarock:
Okay since @nwlandry's reorganization of `convert` into a module already handled breaking line graphs out into their own file, I am going to just close this PR and open a clean one that just deals with adding the `weighted` options.

PR #427, opened 2023-07-25
Added weights option to to_line_graph function.

tlarock:
Addresses discussion in abandoned PR #425, now integrated with new `convert` module.

Adds a single `weights` parameter to `to_line_graph` with options `None` (no weights), `'absolute'` (size of intersection between hyperedges), and `'normalized'` (size of intersection divided by size of smaller). Raises an `XGIError` if a bad choice is given.

### DIFF
--- a/tests/convert/test_line_graph.py
+++ b/tests/convert/test_line_graph.py
@@ -56,6 +56,15 @@ def test_abs_weighted_line_graph(edgelist1, hypergraph1, hypergraph2):
     assert L.edges[("e2","e3")]['weight'] == 1
     assert sum([dat["weight"] for _,_,dat in L.edges(data=True)]) == 3
 
+
+    L = xgi.to_line_graph(hypergraph1, s=2, weights="absolute")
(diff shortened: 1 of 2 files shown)


tlarock:
Does codecov want me to add some tests? I don't understand the failed check.

nwlandry:
Hi @tlarock - yep, it will trigger a warning if the unit test coverage goes down by more than 0.1%. If you're able to add a few tests for the new functionality, that's great, but otherwise we can deal with it later. Thanks for the PR :)

nwlandry (approved):
Thanks for this PR, @tlarock! Looks good! If everything is good from your end, I'll go ahead and merge it. One thing that I would be interested in getting you opinion on is where a breadth-first search would be more efficient down the road, but I think that's not a high priority right now.

tlarock:
Good to be merged from my end @nwlandry! Not sure off the top of my head where BFS could be more efficient but happy to discuss.

maximelucas:
Thanks Tim!!

PR #440, opened 2023-08-01
Added cleanup method to all classes

### LINKED ISSUE #195: cleanup() not working for simplicial complexes
Because `cleanup()` uses HyperGraph specific methods and functions:
`XGIError: Cannot remove_edges_from to SimplicialComplex, use remove_simplices_from instead`

nwlandry:
Fixes #195.

This PR
* Adds the `cleanup()` method to `SimplicialComplex` and `DiHypergraph` and fixes several bugs in the `cleanup()` method for `Hypergraph`.
* Adds unit tests.
* Fixes the strong node removal in Issue #445.
* Adds the `isolates()` method to the `DiNodeView` class.
* Adds the `connected` kwarg to `cleanup()` for hypergraphs and simplicial complexes, which removes nodes/edges not in the largest component.
* `is_frozen` is now a property.


xgi/utils/utilities.py
```diff
@@ -370,28 +370,48 @@ def convert_labels_to_integers(H, label_attribute="label"):
...
+    node_dict = dict(zip(net.nodes, range(net.num_nodes)))
+    edge_dict = dict(zip(net.edges, range(net.num_edges)))
 
-    temp_H.add_nodes_from((id, deepcopy(H.nodes[n])) for n, id in node_dict.items())
-    temp_H.set_node_attributes(
+    temp_net = net.__class__()
+    temp_net._hypergraph = deepcopy(net._hypergraph)
+
+    temp_net.add_nodes_from((id, deepcopy(net.nodes[n])) for n, id in node_dict.items())
+    temp_net.set_node_attributes(
```
maximelucas, line 382: can this not just be `net.copy()`?

tests/core/test_simplicialcomplex.py
```diff
@@ -508,4 +508,88 @@ def test_freeze(edgelist1):
...
+
+def test_remove_node(edgelist1):
+    S = xgi.SimplicialComplex(edgelist1)
+    assert 1 in S
+    S.remove_node(1)
+    assert 1 not in S
+    assert 0 not in S.edges
+
+
+def test_issue_445(edgelist1):
```
maximelucas, line 589: issue #445 was with a Hypergraph was it not? 
Also since `remove_node` did not yet exist for SC

tests/core/test_simplicialcomplex.py
```diff
@@ -508,4 +508,88 @@ def test_freeze(edgelist1):
...
+    assert cleanSC.num_edges == 5
+    simplices = cleanSC.edges.members()
+    assert frozenset({0, 1, 2}) in simplices
+    assert frozenset({3, 4}) in simplices
+    assert frozenset({0, 1}) in simplices
+    assert frozenset({0, 2}) in simplices
+    assert frozenset({1, 2}) in simplices
+
+
+def test_remove_node(edgelist1):
```
maximelucas, line 581: what's the difference with the test below? 
I thought the problem with #445 was that `H._node` should be `{}`, but I cannot see if we're testing that now?

tests/core/test_hypergraph.py
```diff
@@ -528,6 +528,15 @@ def test_remove_node_strong(edgelist1):
     assert 0 not in H.edges
 
 
+def test_issue_445(edgelist1):
```
maximelucas, line 531: Oh i see you also have it here for hypergraph.

maximelucas:
Thanks Nich, looks good. 

It's a bit long so I didn't check every single line. Just a few comments.
Also same usual comment, not to fix now: it feels like we have a lot of (almost) redundant code between SC, HG, and DiHG.

xgi/utils/utilities.py
nwlandry: Unfortunately, because we are renaming all the node IDs we can't do that, because `copy()` keeps the same node and edge IDs.

tests/core/test_simplicialcomplex.py
nwlandry: This is a preemptive test for simplicial complex. As you already mentioned, there is an identical test for `Hypergraph`

nwlandry: Happy to merge this into the `test_remove_node` function.

tests/core/test_simplicialcomplex.py
nwlandry: I test that `S._node` is the dual of `S._edge` below, but I will add an explicit test for the example in #445.

nwlandry:
> Thanks Nich, looks good. Good idea this `connected` argument!
> 
> It's a bit long so I didn't check every single line. Just a few comments. Also same usual comment, not to fix now: it feels like we have a lot of (almost) redundant code between SC, HG, and DiHG.

I agree - my goal is to get things working and then we can focus on inheritance/sharing functionality. One of the issues that I ran into was that because `SimplicialComplex` uses frozen sets, we can only do strong node removal. At that point, it made sense to just implement `remove_ node` and `remove_nodes_from` in `SimplicialComplex`.

maximelucas:
Thanks for the explanations, looks good :)

maximelucas (approved):

nwlandry:
Thanks for the review, @maximelucas!! I appreciate it :)

Issue #437, opened 2023-07-27
easy colorbar plotting


maximelucas:
Add an easy way to plot a colorbar when coloring nodes by a stat. 
Right now, in networkx is pretty convoluted to do it, see: https://stackoverflow.com/questions/26739248/how-to-add-a-simple-colorbar-to-a-network-graph-plot-in-python

maximelucas:
A bit of theory after some digging to see what's the natural of doing this: 
- a simple `plt.colorbar()` does not work in networkx because `nx.draw()` returns `None`. 
- a simple `plt.colorbar(ax=ax)` does not work in xgi because `xgi.draw()` returns an `ax` but not a [mappable](https://matplotlib.org/stable/api/cm_api.html#matplotlib.cm.ScalarMappable) (which is needed for a colobar, and is often an [`AxesImage`](https://matplotlib.org/stable/api/image_api.html)). For example, `plt.imshow()` returns an `AxesImage`, not an axis. 
- Importantly for us, a [`Collection`](https://matplotlib.org/stable/api/collections_api.html#matplotlib.collections.Collection) is a mappable. This includes the output of `plt.scatter()` (a `PathCollection`) which we use in `xgi.draw_nodes()`.
- I think currently our dyads (and hyperedges) are plotted as single `Line2D` (and `Polygon`) elements, which are not mappables, but could be made collections to be mappable?

---

How much do we want from this implementation? 
- Currently, we have 4 cmaps that can be set in `xgi.draw()`, and each of them could have its own colobar: `node_fc_cmap`, `node_ec_cmap`, `dyad_color_cmap`, `edge_fc_cmap`. 
- In each case, it would be nice to be able to set a `vmin` and `vmax` arguments, e.g. to have a single colorar for different plots that have different mins and maxs (see first arguments of [`imshow`](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.imshow.html)).
- For comparison, [`networkx`](https://networkx.org/documentation/stable/reference/generated/networkx.drawing.nx_pylab.draw_networkx.html) seems to have 2 cmaps (`cmap` and `edge_cmap`) with each their own vmin/vmax (`vmin,vmax` and `edge_vmin,edge_vmax`).

--- 

Potential implementations for the node colours:

1. Based on [stackoverflow](https://stackoverflow.com/questions/26739248/how-to-add-a-simple-colorbar-to-a-network-graph-plot-in-python), we can do 

      ```python
[...]

Issue #410, opened 2023-06-30
Making `_scalar_arg_to_dict` consistent


nwlandry:
The inconsistency when converting to colors (Issue #389) was fixed in #402. In the case of `_color_arg_to_dict`, it helps that colors are specified with tuples or strings so that if there is an iterable of scalars, it is clear that the scalars must be mapped to a color. In contrast, for the `_scalar_arg_to_dict` method, it's unclear whether an iterable of scalars are the desired values, or whether those values should be interpolated between preset (or user-defined) min and max allowable values. For example:
```python
>>> import xgi
>>> H = xgi.Hypergraph([{1, 2, 3}, {1, 2}, {1, 4}])
>>> d = H.nodes.degree.asnumpy()
>>> xgi.drawing.draw._scalar_arg_to_dict(d, [1, 2, 3, 4], 10, 30)
{1: 3.0, 2: 2.0, 3: 1.0, 4: 1.0}
>>> xgi.drawing.draw._scalar_arg_to_dict(H.nodes.degree, [1, 2, 3, 4], 10, 30)
{1: 30.0, 2: 20.0, 3: 10.0, 4: 10.0}
```

*I ran the `_scalar_arg_to_dict` function by pasting into a notebook and running it directly, because there was some weirdness with the importing.

nwlandry:
I think this can be closed. If this was implemented, there would be no way for users to directly specify drawing parameters.

maximelucas:
We decided to have both cases behave as the second case (stats one).

maximelucas:
Dependency on `_scalar_arg_to_dict` removed for `draw_dihypergraph` by #476. `draw_multilayer` remains.

Issue #280, opened 2023-02-16
Node positions for drawing could have an improved data structure


nwlandry:
The current data structure is inherited from NetworkX and is key:val pairs of node ID:[x, y]. In the drawing functions, it can be inefficient to draw with a for-loop and in several places, the data structure is converted. I propose that the position be stored as `pos = {"ids": np.array(ids), "x": np.array(x), "y": np.array(y)}`.

nwlandry:
Once #456 is merged, this might be easier to implement.

PR #441, opened 2023-08-01
feat: return mappable to allow colorbar

maximelucas:
Trying to solve #437. So far, only tests.


maximelucas:
Following what I described in #441 , I have rewritten `xgi.draw_nodes()` to use `plt.scatter()`'s full capabilities.

In particular: 
1. I'm not using `_color_arg_to_dict` or `_scalar_arg_to_dict` anymore. All the scaling happens in `plt.scatter()`.
2. Dict is not anymore an accepted format for `node_fc` and other similar arguments . Everything is now internally using arrays. I don't think we loose much with this.  
3. There is no `node_ec_cmap` anymore. `node_ec` can only be specified as a color or a sequence of color (no NodeStat or floats). This comes directly from what's allowed in `plt.scatter`. I don't think we loose much with this. 
4. `node_fc_cmap` is now an explicit argument (and cmaps can be specified as strings now, e.g. "Reds").
5. default params for min/max node size and lw actually work now, soling #442 and #443 for that function.

This implementation has the following advantages: 
- we get easy colormaps
- the implementation is simpler 
- we can easily add even more arguments in the future, that are built in `scatter`: `norm`, `node_shape` (as `marker`), `alpha`

Question before I fix tests: are you okay with this? 
If yes, should `xgi.draw_nodes()` return `(ax, collection)` or just `collection`?

If we agree on this, we could later look into `xgi.draw_hyperedges()` to see if we can do something similar.

nwlandry:
Okay some initial thoughts:
* I really like the thought process behind this and trying to make everything consistent.
* I also like the idea of bringing the API more into line with the matplotlib interface.
* Why do we have to eliminate the dictionary option? Can't we just convert dicts to numpy arrays? We could put this functionality in the `_scalar_arg_to_dict` `_color_arg_to_dict` functions and simplify them. I do get your point thought which, if I understand correctly is that (a) the settings should get respected and (b) that `scatter()` should handle all the coloring and sizing. This does leave me with a few questions:
* Will it be too complicated for users to specify values for the node size if they might have to change the default settings? For example, I think that the values for `s` that scatter accepts are pixel values and aren't subject to max/min values. From the code, it looks like the values do not get resized to the min/max range specified, but are clipped. This might be unexpected behavior if we aren't careful.

Other issues that are potentially related to this are #174, #272, #280, and #410.

maximelucas:
> * Why do we have to eliminate the dictionary option? Can't we just convert dicts to numpy arrays?   

Sure we could do that. I just wasn't sure if it would be a common use case?

> We could put this functionality in the `_scalar_arg_to_dict` `_color_arg_to_dict` functions and simplify them. 

Yes, I thought we might put some of the new code back into simplified versions of those. It was similar to start from scratch at first to know what I really needed. I would probably do that in another PR where we also deal with drawing hyperedges, because if I change `_scalar_arg_to_dict` it will be break them unless we adapt those too.

> * Will it be too complicated for users to specify values for the node size if they might have to change the default settings? For example, I think that the values for `s` that scatter accepts are pixel values and aren't subject to max/min values. From the code, it looks like the values do not get resized to the min/max range specified, but are clipped. This might be unexpected behavior if we aren't careful.

You're right, I'm clipping the node size values with min/max. I thought that's what we were trying to do before, but we were trying to rescale them between min/max indeed. Rescaling is a better idea I agree. I wonder if we want to do it by default or not though. If you compare two hypergraphs, you can see relative difference between nodes but not absolute ones. Should we let the user deal with this, or just have it as an option? Same with `min/max_node_lw`. 

> Other issues that are potentially related to this are #174, #272, #280, and #410.

Thanks I'll look into them.

maximelucas:
Okay Nich I think I did everything we discussed:
- refactored with a new `_draw_arg_to_arr()` function
- from clipping to interpolating with a new `_interp_draw_arg()` function
- added a `rescale_sizes` argument and made the docs more explicit. For now, `False` by default but we can change.

I think it's ready to merge.

nwlandry (approved):
Looks good, Max! Thanks so much for this PR. My only small comment is that perhaps you should be consistent with the documentation on `_draw_arg_to_arr`.  I think a more accurate descriptor would be that it makes args pyplot compliant instead of converting everything to an array?

maximelucas:
Great, changed it. Thanks so much for reviewing this big PR Nich!

Issue #443, opened 2023-08-01
default settings in `draw()` ignored if one of them is passed as argument


maximelucas:
Running 

```python
H = xgi.random_hypergraph(10, ps=[0.6, 0.05, 0.001], seed=3)
settings = {"min_node_size": 5}
xgi.draw_nodes(H, settings=settings)
```

raises `KeyError: 'node_fc_cmap'` for `settings["node_fc_cmap"]`. 

This is because the function starts with 

```python
if settings is None:
        settings = {
            "min_node_size": 10.0,
            "max_node_size": 30.0,
            "min_node_lw": 1.0,
            "max_node_lw": 5.0,
            "node_fc_cmap": cm.Reds,
            "node_ec_cmap": cm.Greys,
        }
```

so that if you pass a `settings` argument with a single one of those (or that is just not `None`), all other default values are never defined. 

I guess default settings should always be defined, and only updated by the arguments?

maximelucas:
https://github.com/xgi-org/xgi/pull/441 solved this for drawing nodes. Still needs fixing for drawing hyperedges.

maximelucas:
#456 fixed if for `draw_hyperedges()`. Still needs fixing for `draw_dihypergraph()`.

Issue #442, opened 2023-08-01
`min/max_val` not working in `_scalar_arg_to_dict()`


maximelucas:
Running 

```python
import xgi
from xgi.drawing.draw_utils import _scalar_arg_to_dict

H = xgi.random_hypergraph(5, ps=[0.5, 0.05])

_scalar_arg_to_dict(50, H.nodes, min_val=10, max_val=20)
# --> {0: 50, 1: 50, 2: 50, 3: 50, 4: 50}   # all values are 50 > 20, wrong!
```

As a consequence, the default settings in `draw_nodes()`
```python
      "min_node_size": 10.0,
      "max_node_size": 30.0,
      "min_node_lw": 1.0,
      "max_node_lw": 5.0,
```

and all other default min/max settings going through `_scalar_arg_to_dict` are ignored.

maximelucas:
https://github.com/xgi-org/xgi/pull/441 solved this for drawing nodes. Still needs fixing for drawing hyperedges.

maximelucas:
#456 fixed this for hyperdges. The functions `_color_arg_to_dict()` and `_scalar_arg_to_dict()` are not used in the new versions, but are still used in `draw_dihypergraph()`, `draw_multilayer()`.

maximelucas:
Fixed by #476 for `draw_dihypergraph` by removing dependency. `draw_multilayer` remains.

PR #456, opened 2023-08-24
Rewrite `draw_hyperedges()`

maximelucas:
- rewrites draw_hyperedges using matplotlib's Collections, related to #174 (only slight speedup observed when looking at cleaned up enron). This makes the code shorter, cleaner, a bit faster, more in line with networkx, and allows us to have colorbars (#437). It also makes it more consistent with our `draw_nodes()` after PR #441.
- addresses #442 and #410 for this function with a new simpler `_parse_color_arg()` function
- addresses #443 for this function
- new arguments `dyad_style`, `dyad_vmin/max`, `edge_vmin/max`, `rescale_sizes`. 
- new tuto demonstrating all the arguments and their possible formats

Example customisation:

```python
ax, (dyad_collection, edge_collection) = xgi.draw_hyperedges(
    H,
    pos=pos,
    dyad_color=[3, 1, 4],
    edge_fc=H.edges.size,
    dyad_color_cmap="viridis",
    edge_fc_cmap="tab10"
)

plt.colorbar(dyad_collection, label="dyads")
plt.colorbar(edge_collection, label="edges")
```
Screenshot 2023-08-25 at 00 10 17


xgi/drawing/draw_utils.py
```diff
@@ -76,6 +83,89 @@ def _update_lims(pos, ax):
...
+
+    if not is_color_like(colors) and len(colors) != xsize:
+        raise ValueError(
+            f"The input color argument must be a single color or its length must match the number of plotted elements ({xsize})."
+        )
+
+    return colors, colors_are_mapped
+
+
 def _draw_arg_to_arr(arg):
```
nwlandry, line 169: should this be called `parse_scalar_arg`?

nwlandry:
Great work, @maximelucas, and thanks for implementing such an awesome update to the drawing functionality! I've had a look and I think it is time to merge this. Before you do, though, can you correct the tutorial errors and change the `draw_simplices` function?

maximelucas:
I think this is ready now. I updated `draw()` and `draw_simplices()` so that they are consistent with `draw_hyperedges()`. 

The functions `_color_arg_to_dict()` and `_scalar_arg_to_dict()` are not used in the new versions, but are still used in `draw_dihypergraph()`, `draw_multilayer()`, and `draw_hypergraph_hull()`. These functions might need to be updated for consistency. Also the way they handle the `kwargs`. I think this PR is already big enough so I would l create new issues and leaves this other PRs. 

The only thing left: decide if we keep the current default cmap (crest_r) for edges or not.
I managed to create another cmap candidate which is a sequential blue going from not too light to not too dark:
Screenshot 2023-10-16 at 12 45 59

Here are 3 examples of different sizes with different densities and edges sizes with this cmap:

Screenshot 2023-10-16 at 12 46 16
Screenshot 2023-10-16 at 12 46 11
Screenshot 2023-10-16 at 12 46 06

Note how the original cmap is more saturated that what is shown there: that's because we use a default `alpha=0.4` when plotting. In any case, I think this cmap might look better for the larger examples, but it looks a bit too light/transparent in the smaller example. 

Let me know what you think on the cmap.

nwlandry:
Looks great, @maximelucas!! Thanks for your hard work on this!

nwlandry (approved):

PR #457, opened 2023-08-25
Attempt at integrating the tutorials in API

thomasrobiglio:


maximelucas:
Nice work, thanks so much @thomasrobiglio !! Honestly I think it looks good.
We could start being picky with some of the spacing but I'd say let's do that only later when we change to the other documentation theme. 

Minor things:
- the links to other tutorials link to the github notebooks, instead of the documentation tutorial.
- The names of the tutorials don't look consistent. That's coming from the titles inside the notebooks I think?
Screenshot 2023-09-29 at 17 46 01

Finally I was just wondering how much of this is automatised/customisable? What's the workflow when we add new tutorials?

thomasrobiglio:
Thanks @maximelucas for the feedback! 
Regarding the spacing I need to look if/how I can customize di appearance of everything... for now nbsphinx does everything on its own. 

Regarding links and titles I will modify in a consinstent way next week. (yes nb sphinx takes the first markdown section in the tutorial as the title).

Regarding the automatisation it is quite straightforward, every tutorial/recipe inside the correct directory can be referenced in the corresponding .rst file and is displayed automatically.

thomasrobiglio:
I am updating the links, since I am 'guessing' them I might need to open another pr once the version of the readthedocs with the tutos is stable

thomasrobiglio:
Ok, everything we discussed last week is done. I have also added a recipe for the multilayer draw (it is not shown anywhere else otherwise). 

This is now ready for reviewing and merging.

Screenshot 2023-10-16 alle 14 40 04

nwlandry:
Looks good @thomasrobiglio and ready to merge if @maximelucas agrees. The only small detail is to change the "-" to "." in the title of Recipe 12.

maximelucas:
Looks great, thanks so much Thomas!!! This is a big step for our docs.

maximelucas (approved):

PR #462, opened 2023-08-28
Fix #461

### LINKED ISSUE #461: lshift does not preserve node order
Even if `H1` and `H2` have the same nodes in the same order, `H1 << H2` has nodes in a different order.

This is because lshift adds the edges before the nodes.

maximelucas:
Keep a consistent node order with lshift, solving #461.


nwlandry (approved):
good call!

PR #475, opened 2023-10-16
Missing `focus tutorials` in API

thomasrobiglio:


nwlandry:
Thanks @thomasrobiglio! Looks great! Two quick question before you merge: (1) can we remove the top-level tutorials folder now? (2) can we get rid of the quickstart notebook? Thanks for all the work on this!

nwlandry:
Oops! I was looking at the wrong branch! Disregard (1).

thomasrobiglio:
Yes, we can remove the quickstart tuto. If we don't have time constraints for a new release I'd like to take 2-3 more days to check for imprecisions before merging.

nwlandry:
Sounds great! I'll approve it and you can merge it when you think it's ready.

nwlandry (approved):

maximelucas:
Nice! I added another In Depth tuto in #456 but somehow I cannot see in main now.
Could you add it?
https://github.com/xgi-org/xgi/blob/4e513734dfa4a22bc3095fe524398b8ae7a8d6cd/docs/source/api/tutorials/In%20Depth%202%20-%20Drawing%20hyperedges.ipynb

thomasrobiglio:
@maximelucas will do tomorrow. To see it in the API you should also put the reference in the corresponding .rst file (https://github.com/xgi-org/xgi/blob/main/docs/source/api/tutorials/in_depth.rst#in-depth-tutorials in this case).

Idk if there is an automated way to have all the .ipynb file but the way we are doing it now has some upsides e.g. we can choose the order in which tutos are displayed and create the different categories.

Issue #433, opened 2023-07-27
`sep` argument not working multilayer plotting function


maximelucas:
It seems to have no effect

thomasrobiglio:
uhm I will have a look at it! thank you @maximelucas

nwlandry:
Hi @thomasrobiglio - I started adding some features to `draw_multilayer` - can I push this first?

maximelucas:
Okay nice :) This is with the new version?
Then maybe just document this yes. Could even be nice to give a rough idea/example of when it starts not having an effect.

thomasrobiglio:
>> This is with the new version?

Yes, but it should work the same in the stable version

>> Could even be nice to give a rough idea/example of when it starts not having an effect.

Good idea, let's keep the issue open and I will work on this

PR #476, opened 2023-10-17
Fixing things in drawing functions

### LINKED ISSUE #443: default settings in `draw()` ignored if one of them is passed as argument

maximelucas:
- addressed #433 @thomasrobiglio can you check?
- fixed #443 (one function was missing)
- I tried to make `draw_dihypergraph()` more consistent with the other (updated) functions. I solved some bugs on the way, but making it fully consistent (and bug free) will require a bigger rewrite. I'll open an issue.
- Added some tests for `draw_dihypergraph()`. More needed.


thomasrobiglio (approved):

xgi/drawing/draw.py
```diff
@@ -3,11 +3,12 @@
 from inspect import signature
 from itertools import combinations
 
+import matplotlib as mpl
 import matplotlib.pyplot as plt
 import numpy as np
 import seaborn as sb  # for cmap "crest"
 from matplotlib import cm
-from matplotlib.patches import FancyArrow
+from matplotlib.patches import FancyArrow, FancyArrowPatch
```
thomasrobiglio, line 11: Looks like we don't need anymore the `FancyArrow` import, right?

xgi/drawing/draw.py
```diff
@@ -887,6 +885,8 @@ def draw_node_labels(
     clip_on_nodes : bool, optional
         Turn on clipping of node labels at axis boundaries.
         By default, True.
+    zorder : int, optional
+        The layer on which to draw the nodes.
```
thomasrobiglio, line 889: The vertical order on which to draw the labels. 
If None (default) the labels are drawn on the top level.

xgi/drawing/draw.py
```diff
@@ -1665,6 +1672,29 @@ def draw_dihypergraph(
...
+        Line width of the hyperedges (lines). If int or float, use the same width for
+        all hyperedges. If a dict, must contain (hyperedge_id: width) pairs. If other
+        iterable, assume the widths are specified in the same order as the hyperedges
+        are found in DH.edges. By default, 1.5.
+    arrowstyle : str, optional
+        By default: '-\|>'. See `matplotlib.patches.ArrowStyle` for more options.
+    arrowsize : int (default=10)
+        Size of the arrow head's length and width. See `matplotlib.patches.FancyArrowPatch`
+        for attribute `mutation_scale` for more info.
+    connectionstyle : string (default="arc3")
```
thomasrobiglio, line 1693: Maybe we can add an example of this (and the thing above for the `arrowstyle` parameter) in the in depth tuto?

thomasrobiglio (commented):
Thanks @maximelucas this looks super good! I have added minor comments. One more small thing:

I have noticed that we were inconsistent in the naming of the in-depth tutos (see pic). Can you: 
In depth 2 - Drawing hyperedges -> In depth: drawing hyperedges
In Depth - Drawing DiHypergraphs -> In depth: drawing DiHypergraphs
(or something else to make the three names consistent?)
Screenshot 2023-10-25 alle 11 43 26

xgi/drawing/draw.py
maximelucas: Good catch, done

xgi/drawing/draw.py
maximelucas: Good idea, done

maximelucas:
Good points Thomas, thanks! All addressed now I think.

nwlandry (approved):

nwlandry:
Great work, @maximelucas!! Much needed changes and thanks for pushing this through!

Issue #477, opened 2023-10-17
make drawing dihypergraph and multilayer consistent with other drawing functions


maximelucas:
`draw_dihypergraph` and `draw_multilayer` have not been updated since we updated the drawing of nodes #441  and hyperedges  #456 . We should update them for consistency by using collections. We should also remove their dependencies on the "old" functions `_scalar_arg_to_dict()` and `_color_arg_to_dict` that have problems #442 #410, to use the newer  `_parse_color_arg` , `_draw_arg_to_arr`, `_interp_draw_arg`. This is the last place the "old" ones are used.

For `draw_dihypergraph`, we could try to use more existing functions. The drawing of the "edge nodes" could be done via the `draw_nodes` function.

Issue #388, opened 2023-06-07
improve default coloring in viz


maximelucas:
By default, hyperedges are drawn with a shade (of a base color) proportional to their size.

In small and simple cases, it doesn't look great.  Can we improve this?
Example from #387 , the triplet edge looks very pale: 
![image](https://github.com/xgi-org/xgi/assets/7493360/0ad74147-7aeb-4a98-b544-c53c59d46c08)

maximelucas:
#456 solved this for hyperedges. Dihyperedges not done yet.

maximelucas:
#476 fixed this for dihypergraphs. Last one to change is multilayer I think.

PR #486, opened 2023-10-27
Rewrote the multilayer function

maximelucas:
This makes it more consistent with other (see #477).

- removes dependency on `_scalar_arg_to_dict` solving #442 and #410.
- makes progress on #388.

I couldn't make `sep` work (related to #433) @thomasrobiglio can you check?
Related to that I would maybe remove the `height` and `width` arguments as users can control that externally and pass an `ax` of their choice to the function. 

The function in general could some more tests but I would leave that for a future PR. I created a new notebook that showcases many use cases and they work.


thomasrobiglio:
🚀 @maximelucas incredible! 

About the `sep` parameter you mean have a look if it works in the original function?

About removing the `height` and `width` params and allowing the user to pass an `ax` the problem (and reason why I went in that direction originally) is that the `ax` input parameter needs to be a `Axes3D` object and not a `Axes` (as in the other functions for potting)

maximelucas:
> About the `sep` parameter you mean have a look if it works in the original function?

So in the original function I remember I thought it wasn't working but then you managed to see it work by using small values I think. Here I'm not managing to see any change when I change `sep`. Can you maybe check if you manage to see changes like you did before, or otherwise help find why mine is different from yours for that? I'm using `sep` everywhere to set the height.. I looked but I haven't understood why it wouldn't move yet. I even played with `height`.
 
> About removing the `height` and `width` params and allowing the user to pass an `ax` the problem (and reason why I went in that direction originally) is that the `ax` input parameter needs to be a `Axes3D` object and not a `Axes` (as in the other functions for potting)

Sorry I wasn't with what I meant. Right now we have `width` and `height` as arguments and 
```python
if ax is None:
        _, ax = plt.subplots(figsize=(width, height), subplot_kw={"projection": "3d"})
```

what I mean is we could replace it by this which works as well:

```python
if ax is None:
        _, ax = plt.subplots(subplot_kw={"projection": "3d"})
```
and then we can also remove the `width` and `height` arguments. Then indeed, if people want to use a custom axis an control width and height they need to pass a 3d one, but that's okay it's in the doc, they should know. In other words, we don't have `width` and `height` arguments for any other draw functions, and other that needing to know it should be 3d, this function doesn't to be different I think?

maximelucas:
Okay I know what's happening with the `sep`. It works, but we are rescaling the z-axis so we don't see any difference (I temporarily displayed the spines):

`sep = 0.1`:
Screenshot 2023-10-27 at 19 35 59

`sep = 1`:
Screenshot 2023-10-27 at 19 37 13

It appears that changing the `height` of the axis does not change much: matplotlib ensures the 3d plot looks like a cube.

We can fix this by doing `ax.set_aspect("equal")`: 

`sep = 1`:

Screenshot 2023-10-27 at 19 42 50

`sep = 0.4`:
Screenshot 2023-10-27 at 19 43 57

`sep = 0.1`:
Screenshot 2023-10-27 at 19 45 10

thomasrobiglio:
Thanks for the clarifications @maximelucas , good catch about `sep` and also removing the two parameters for the figure size looks like a much better solution than what i'd originally did.

maximelucas:
Ok it's ready I somebody approves then we can merge

nwlandry:
I had a quick look, and great work! I will defer final approval to Thomas since he has more expertise here, but two quick questions: (1) `draw_hypergraph_hull` is the last function that references the `_color_arg_to_dict` function, right? (2) does it make sense to remove the tutorials folder here or in a separate PR?

maximelucas:
Thanks Nich! (1) Yes! (2) Done.

thomasrobiglio (approved):
Looks good @maximelucas ! thank you very much 🚀

PR #491, opened 2023-11-02
Drawing hull as an option in draw_hyperedges

maximelucas:
`draw_hypergraph_hull` was last function to make consistent #477. I noticed the edge-plotting part was doing exactly the same as `draw_hyperedges` except for the few lines drawing the hulls. So I:
- added a `hull=False` and `radius` argument to `draw_hyperedges`, which needed just 10 more lines of code to work. So everything works out of the box because `draw_hyperedges` works.
- added this option to `draw` so we can now do `xgi.draw(H, hull=True)`. 
- removed `draw_hypergraph_hull` and `_draw_hull` as they are not used anymore. 
- fixed tests and updates notebooks
- Change the default node_size everywhere to 7 (was 15). This was the size in the hull function, larger ones look bad. I think it also generally looks better for the non-hull ones. The only thing is that nodes now are by default too small for drawing node_labels, so the user needs to set a larger size if they want to do that but I think that's ok. 
- removed `_scalar_arg_to_dict` as not used anymore


thomasrobiglio (changes requested):
Looks good @maximelucas, thank you.

Minor fixes:
1. remove ref. in the API xgi.drawing.draw.rst
2. I think the tutorial [Tutorial 7 - Convex hulls hypergraph plotting.ipynb](https://github.com/xgi-org/xgi/blob/hull/docs/source/api/tutorials/Tutorial%207%20-%20Convex%20hulls%20hypergraph%20plotting.ipynb) is better to put it in the in depth (as a separate one) or at the end of [In Depth 2 - Drawing hyperedges.ipynb](https://github.com/xgi-org/xgi/blob/hull/docs/source/api/tutorials/In%20Depth%202%20-%20Drawing%20hyperedges.ipynb)
3. About the nodes' size change I think it is nice, looks cleaner this way. Maybe we should add a line somewhere in the docs telling the thing for the labels (that the user has to change the nodes' size manually)... or we can implement a thing that automatically draws the labels prop to node size (it's like that in graph tool), but I think in a new PR

maximelucas:
Good ideas Thomas! Should be all good now. 

For 3) I added a note in the docstrings. At first I thought we could just do the following to change it automatically like you said 
```python
if node_labels and node_size is None:
    node_size = 15
```
but by default `node_size = 7` not `None` in the function call. So if we want to do it like this in another PR we would maybe need to set it to `None` and do inside of the function `if node_size is None: node_size = 7`...

thomasrobiglio (approved):

PR #492, opened 2023-11-03
Add the ability to draw hypergraphs as a bipartite graph

nwlandry:
This PR does the following:
* Removes `draw_dihypergraph` and moves its contents to `draw_bipartite`.
* The `draw_bipartite` now handles `Hypergraph` and `DiHypergraph` objects.
* Added the `bipartite_spring_layout` position function.


maximelucas:
Just a thought: conceptually, could this function just be the same as the `draw_dihypergraph` with the edge markers, except that the links are drawn without arrows and the layout would be different?

nwlandry:
Mmmmm. This is a great point. I think you're right. Would it make sense to group them together, call it `draw_bipartite` and have different handling for directed and undirected hypergraphs?

maximelucas:
I thought that could be a possibility. Worth thinking about. The current dihypergraph one only make sense as a bipartite if the edge markers are drawn though, I'd say.

Maybe directedness and "bipartite-ness" are two somewhat independent elements here.
Could the arrow drawing be included as a `directed` option in `draw_hyperedges`? Not sure that would include the edge markers. 

And then the `draw_bipartite` could just be about the layout really. Either the current layout we have in the  dihypergraph one, or a more bipartitey layout, the user could choose. And draw directed or undirected edges.

xgi/drawing/draw.py
```diff
@@ -5,7 +5,6 @@
 import matplotlib as mpl
 import matplotlib.pyplot as plt
 import numpy as np
-import seaborn as sb  # for cmap "crest"
```
maximelucas, line 8: I think we need to keep this even if it's "unused", because without it I was getting errors for using the cmap "crest" which is only provided by seaborn (and not matplotlib)

nwlandry:
I have now implemented all of your suggestions, @maximelucas. Should be ready to re-review.

docs/source/api/tutorials/In Depth 3 - Drawing DiHypergraphs.ipynb
```diff
@@ -13,7 +13,7 @@
    "id": "ea8aad6a",
```
maximelucas, line 13: <p>The output was <code>(ax, node_coll, edge_coll)</code> and is now <code>(ax, node_coll, edge_coll, dyad_coll) </code>if I got that right. But the dyad_coll is a list of <code>FancyArrowPatch</code> and not an actual python <code>Collection</code>. </p><p> </p><p>The reason to output the collection was (i) for easy colorbars and (ii) to easily change parameters of the whole collection (like alpha value). What use did you have in mind for outputting the new list? Maybe we could make it an actual collection instead?</p><p> </p><p>I think my initial thought for not needing to output is that I thought either the arrows would be multicolored but consistent with the edges, in which case the edge_coll is sufficient, or be a single neutral color like black, in which case maybe returning them is less useful.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/xgi-org/xgi/pull/492/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

maximelucas, line 13: <p>This case looks pretty bad now. How is it different from before? Should we not display it if it looks that bad?</p><p> </p><p>Also I thought you mentioned to change the pos into a single tuple argument, but maybe it's just the notebook that's not up to date, I haven't checked everything yet.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/xgi-org/xgi/pull/492/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='29'/>

maximelucas, line 13: <p>same here</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/xgi-org/xgi/pull/492/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='33'/>

docs/source/api/drawing/xgi.drawing.draw.rst
```diff
@@ -8,7 +8,6 @@ xgi.drawing.draw
    .. rubric:: Functions
    
    .. autofunction:: draw
-   .. autofunction:: draw_dihypergraph
    .. autofunction:: draw_multilayer
```
maximelucas, line 11: draw_bipartite should be added?

xgi/drawing/draw.py
```diff
@@ -1608,47 +1591,87 @@ def draw_dihypergraph(
...
         Alternate default values. Values that can be overwritten are the following:
-        * "min_node_size" (default: 5)
-        * "max_node_size" (default: 30)
-        * "min_node_lw" (default: 0)
-        * "max_node_lw" (default: 5)
-        * "min_lines_lw" (default: 0)
-        * "max_lines_lw" (default: 50)
-        * "min_source_margin" (default: 0)
-        * "min_target_margin" (default: 0)
+        * min_node_size
```
maximelucas, line 1646: add default values like before

xgi/drawing/draw.py
```diff
@@ -1608,47 +1591,87 @@ def draw_dihypergraph(
...
+        * max_node_size
+        * min_node_lw
+        * max_node_lw
+        * min_edge_marker_size
+        * max_edge_marker_size
+        * min_edge_marker_lw
+        * max_edge_marker_lw
+        * node_fc_cmap
```
maximelucas, line 1654: not needed here, already a input argument

xgi/drawing/draw.py
```diff
@@ -1608,47 +1591,87 @@ def draw_dihypergraph(
...
+        * node_ec_cmap
```
maximelucas, line 1655: not used, is it?

xgi/drawing/draw.py
```diff
@@ -1608,47 +1591,87 @@ def draw_dihypergraph(
...
+        * edge_marker_fc_cmap
```
maximelucas, line 1656: not needed here, already a input argument

xgi/drawing/draw.py
```diff
@@ -1608,47 +1591,87 @@ def draw_dihypergraph(
...
+        * edge_marker_ec_cmap
```
maximelucas, line 1657: i think we don't use this for now

xgi/drawing/draw.py
```diff
@@ -1608,47 +1591,87 @@ def draw_dihypergraph(
...
+        * min_dyad_lw
+        * max_dyad_lw
+        * dyad_color_cmap
```
maximelucas, line 1660: not needed here, already a input argument

xgi/drawing/draw.py
```diff
@@ -1658,45 +1681,86 @@ def draw_dihypergraph(
...
+        "max_edge_marker_lw": 5,
+        "min_edge_marker_size": 0,
+        "max_edge_marker_size": 5,
+        "min_dyad_lw": 1,
+        "max_dyad_lw": 5,
+        "node_fc_cmap": cm.Reds,
+        "node_ec_cmap": cm.RdBu,
+        "dyad_color_cmap": cm.Blues,
+        "edge_marker_fc_cmap": cm.Greys,
+        "edge_marker_ec_cmap": cm.Blues,
```
maximelucas, line 1716: already in arguments above, or not used i think

xgi/drawing/draw.py
```diff
@@ -1658,45 +1681,86 @@ def draw_dihypergraph(
...
 
-    # convert to hypergraph in order to use the augmented projection function
-    H_conv = convert.to_hypergraph(DH)
+    node_settings = {
+        "min_node_lw": settings["min_node_lw"],
+        "max_node_lw": settings["max_node_lw"],
+        "min_node_size": settings["min_node_size"],
+        "max_node_size": settings["max_node_size"],
+        "node_fc_cmap": settings["node_fc_cmap"],
+        "node_ec_cmap": settings["node_ec_cmap"],
```
maximelucas, line 1729: see above

xgi/drawing/draw.py
```diff
@@ -1658,45 +1681,86 @@ def draw_dihypergraph(
...
+    }
+
+    edge_marker_settings = {
+        "min_node_lw": settings["min_edge_marker_lw"],
+        "max_node_lw": settings["max_edge_marker_lw"],
+        "min_node_size": settings["min_edge_marker_size"],
+        "max_node_size": settings["max_edge_marker_size"],
+        "node_fc_cmap": settings["edge_marker_fc_cmap"],
+        "node_ec_cmap": settings["edge_marker_ec_cmap"],
```
maximelucas, line 1738: see above

xgi/drawing/draw.py
```diff
@@ -1707,111 +1771,154 @@ def draw_dihypergraph(
...
-    # draw "edge" phantom nodes
-    if edge_marker_toggle:
+    if node_labels is not None:
+        draw_node_labels(
+            H,
+            pos=node_pos,
+            node_labels=node_labels,
+            ax_nodes=ax,
+            zorder=3,
+        )
```
maximelucas, line 1812: this can be passed simply to draw_nodes like before, right?

xgi/drawing/draw.py
```diff
@@ -1707,111 +1771,154 @@ def draw_dihypergraph(
...
-            zorder=2,
-            params=settings,
-            rescale_sizes=rescale_sizes,
+    if hyperedge_labels is not None:
+        draw_node_labels(
+            H.dual(),
+            pos=edge_pos,
+            node_labels=hyperedge_labels,
+            ax_nodes=ax,
+            zorder=3,
```
maximelucas, line 1837: same question

xgi/drawing/draw.py
```diff
@@ -1707,111 +1771,154 @@ def draw_dihypergraph(
...
-    edges = DH.edges.filterby("size", 2, "geq")
+    # compute axis limits
+    _update_lims(pos, ax)
 
-    nodes_id = list(DH.nodes)  # in original order, to find index for arrays
+    return ax, (node_collection, edge_marker_collection, dyad_collection)
+
+
+def _draw_directed_edges(
+    edge_ids,
```
maximelucas, line 1900: maybe second after ax, don't remember what we do in other functions

xgi/drawing/draw.py
```diff
@@ -1707,111 +1771,154 @@ def draw_dihypergraph(
...
+    ax,
+    H,
+    settings,
```
maximelucas, line 1903: is usually last it ink

nwlandry:
Thanks for the suggestions, @maximelucas! It looks like I need to take more time to address these concerns, and I'll let you know when I'm done!

docs/source/api/drawing/xgi.drawing.draw.rst
nwlandry: Fixed. Good catch!

xgi/drawing/draw.py
nwlandry: Good catch! You're right, we can just use the `draw_nodes` function.

xgi/drawing/draw.py
nwlandry: Yep! Good catch!

xgi/drawing/draw.py
nwlandry: Fixed.

xgi/drawing/draw.py
nwlandry: Fixed. I deleted this arg and added `max_order` as an arg.

xgi/drawing/draw.py
nwlandry: Fixed.

xgi/drawing/draw.py
nwlandry: Fixed.

xgi/drawing/draw.py
```diff
@@ -1608,47 +1591,87 @@ def draw_dihypergraph(
...
+        are found in DH.edges. By default, 1.
     node_labels : bool or dict, optional
         If True, draw ids on the nodes. If a dict, must contain (node_id: label) pairs.
         By default, False.
     hyperedge_labels : bool or dict, optional
         If True, draw ids on the hyperedges. If a dict, must contain (edge_id: label)
         pairs.  By default, False.
+    max_order : int, optional
+        Maximum of hyperedges to plot. If None (default), plots all orders.
+    arrowsize : int (default=10)
```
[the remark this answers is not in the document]
nwlandry: Fixed.

xgi/drawing/draw.py
nwlandry: Fixed.

xgi/drawing/draw.py
nwlandry: Fixed.

xgi/drawing/draw.py
nwlandry: You're right! Fixed.

xgi/drawing/draw.py
nwlandry: Correct. Fixed.

docs/source/api/tutorials/In Depth 3 - Drawing DiHypergraphs.ipynb
nwlandry: This is now removed and the `pos` issue is fixed.

nwlandry: This example is now removed.

nwlandry:
@maximelucas, I believe I addressed all of your comments! Thanks for such a thorough and helpful review! I did the following:
* I standardized all of the function arguments in response to your review.
* I made `draw_undirected_dyads` and `draw_directed_dyads` standalone functions and made them available to the user.
* Added default values to the settings and made them match other functions.

tests/drawing/test_draw.py
```diff
@@ -663,7 +673,46 @@ def test_draw_bipartite_with_str_labels_and_isolated_nodes():
     assert len(node_coll4.get_offsets()) == 6  # number of original nodes
     assert len(edge_marker_coll4.get_offsets()) == 2  # number of original edges
     assert len(ax4.patches) == 7  # number of lines
-    plt.close()
+    plt.close("all")
+
+
+# 1922, 1925, 1928-1929, 1936, 1942, 1949-1952, 1956, 1960, 2101, 2104, 2111, 2117, 2124-2126, 2130, 2134, 2163, 2166, 2173, 2189, 2199, 2210, 2236, 2247
```
maximelucas, line 679: is this needed?

xgi/drawing/draw.py
```diff
@@ -1978,13 +1986,7 @@ def draw_undirected_dyads(
...
         )
 
-    node_pos, edge_pos = pos
-
-    dyads = to_bipartite_edgelist(H)
-    dyad_pos = np.asarray(
-        [(node_pos[e[0]], edge_pos[e[1]]) for e in dyads if e[1] in edge_ids]
-    )
-
+    print(len(dyad_pos), len(dyad_color))
```
maximelucas, line 1989: remove?

xgi/drawing/draw.py
```diff
@@ -1947,24 +1936,43 @@ def draw_undirected_dyads(
...
+
+    if isinstance(dyad_color, np.ndarray):
+        dyad_color = np.array(
+            list(
+                chaini(
+                    [dc] * int(s) for s, dc in zip(H.edges.size.aslist(), dyad_color)
+                ),
+            )
+        )
+
```
maximelucas: what does this do? Multiplies the number of elements of linewidth by the size of each edge?

maximelucas:
Thanks! Those are quick small comments for the last changes, I'll have a better look at the entire code soon.

xgi/drawing/draw.py
nwlandry: Oops! Thanks for finding this!

xgi/drawing/draw.py
nwlandry: This allows you to specify a color for each edge ID and automatically apply it to all associated bipartite edges instead of specifying a color for all bipartite edges.

nwlandry: I added a comment in the code explaining what these lines do.

maximelucas:
I'll do that next week!

nwlandry:
Great! Thanks so much!

docs/source/api/tutorials/In Depth 3 - Drawing DiHypergraphs.ipynb
maximelucas: <p>Yea I was thinking maybe we don't return the list of patches. In any case I dont see a case where we would need a different colorbar for edge markers and bipartite edges. Maybe we enforce that if bipartite edges need to be colored with a cmap, it needs to be the same as that of edges markers?</p>

maximelucas, line 13: <pre class="ql-syntax" spellcheck="false">Line #1.    xgi.draw_bipartite(DH)
</pre><p>The default coloring of hyperedges is a single color, instead of the crest cmap mapping edge size that we have in other functions. Any reason for this?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/xgi-org/xgi/pull/492/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

maximelucas, line 13: <p>The default cmap for hyperedges is now "Blues", which is different from all other functions where it is "crest". Any reason for that? I would maybe keep "crest" for consistency.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/xgi-org/xgi/pull/492/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

maximelucas, line 13: <p>is there a way to have the old "iterations=0" when wanted?</p><p>If not, maybe open an issue to re-implent it.</p><p>I like that in small cases, that one had the edge_markers at the barycenters. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/xgi-org/xgi/pull/492/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='27'/>

xgi/drawing/draw.py
```diff
@@ -1578,83 +1589,116 @@ def draw_dihypergraph(
...
-        at the barycenters. This initial layout may suffer from overlap between edge-nodes.
-        After that, a spring layout is ran starting from the
-        initial layout, and each iteration makes all of the nodes overlap less.
+    aspect : {"auto", "equal"} or float, optional
+        Set the aspect ratio of the axes scaling, i.e. y/x-scale. `aspect` is passed
+        directly to matplotlib's `ax.set_aspect()`. Default is `equal`. See full
+        Set the aspect ratio of the axes scaling, i.e. y/x-scale. `aspect` is passed
+        directly to matplotlib's `ax.set_aspect()`. Default is `equal`. See full
+        description at
+        https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.set_aspect.html
```
maximelucas, line 1673:
```suggestion
        Set the aspect ratio of the axes scaling, i.e. y/x-scale. `aspect` is passed
        directly to matplotlib's `ax.set_aspect()`. Default is `equal`. See full
        description at
        https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.set_aspect.html
```

xgi/drawing/draw.py
```diff
@@ -1664,160 +1708,465 @@ def draw_dihypergraph(
...
+    -------
+    ax : matplotlib.pyplot.axes
+        The axes corresponding the visualization
+    * dyad_collection : matplotlib LineCollection
+        of bipartite edges
+
+    Raises
+    ------
+    XGIError
+        If something different than a DiHypergraph is passed.
```
maximelucas, line 1913:
```suggestion
        If something different than a Hypergraph is passed.
```

maximelucas: But I don't see that it raises an error, does it?

xgi/drawing/draw.py
```diff
@@ -1578,83 +1589,116 @@ def draw_dihypergraph(
...
-        * phantom_node_collection : matplotlib PathCollection
-            Collection containing the nodes representing edges
+        * edge_marker_collection : matplotlib PathCollection
+            Collection containing the edge markers
+        * dyad_collection : matplotlib LineCollection if undirected,
+                            list of FancyArrowPatches if not
+            Collection containing the edges
 
     Raises
     ------
```
maximelucas, line 1704: Below:

If something different than a Hypergraph is passed.

maximelucas:
Some more minor comments.. we're almost there, it's a lot of code 😄 I'll check the rest of the code later today.

nwlandry:
@maximelucas thanks so much! Let me know when you're done and I'll start addressing your comments!

xgi/drawing/draw.py
```diff
@@ -1664,160 +1708,465 @@ def draw_dihypergraph(
...
 
-    if not max_order:
-        max_order = max_edge_order(H_conv)
+    if not pos:
+        pos = bipartite_spring_layout(H)
 
-    # set default colors
-    if lines_fc is None:
-        lines_fc = H_conv.edges.size
+    node_pos, edge_pos = pos
```
maximelucas, line 1760: Since the `pos` format is different from other functions, should we raise an explicit error after checking if it's a two-dict tuple? Right now a single dict would give a generic error like "too much to unpack"

xgi/drawing/draw.py
```diff
@@ -1664,160 +1708,465 @@ def draw_dihypergraph(
...
+
+    """
+    settings = {
+        "min_dyad_lw": 1,
+        "max_dyad_lw": 10,
+    }
+    settings.update(kwargs)
+
+    if not pos:
+        pos = bipartite_spring_layout(H)
```
maximelucas, line 1928: same

xgi/drawing/draw.py
```diff
@@ -1664,160 +1708,465 @@ def draw_dihypergraph(
...
+    """
+    settings = {
+        "min_dyad_lw": 1,
+        "max_dyad_lw": 10,
+        "min_source_margin": 0,
+        "min_target_margin": 0,
+    }
+    settings.update(kwargs)
 
+    if not pos:
```
maximelucas, line 2121: same

xgi/drawing/draw.py
```diff
@@ -1664,160 +1708,465 @@ def draw_dihypergraph(
...
+        from ..convert import to_hypergraph
```
maximelucas, line 2122: is this one purpose?

maximelucas:
I think I checked all of it 😛 
I would just add an example somewhere of an undirected one, and raise an issue to create a new function with the "classic" graph bipartite layout, and maybe also with the barycenter layout. 
Thanks for all the work!

xgi/drawing/draw.py
nwlandry: Good catch! Fixed.

xgi/drawing/draw.py
nwlandry: Fixed.

xgi/drawing/draw.py
nwlandry: Fixed.

xgi/drawing/draw.py
nwlandry: To compute the layout? If so, yes.

xgi/drawing/draw.py
nwlandry: I added a case checking this.

xgi/drawing/draw.py
nwlandry: Fixed.

docs/source/api/tutorials/In Depth 3 - Drawing DiHypergraphs.ipynb
nwlandry: Fixed!

nwlandry: I added a function called `edge_positions_from_barycenters` that takes in a dictionary of nodal positions and then returns a dictionary of edge positions computed from the barycenters.

nwlandry: This is good because it is no longer tied to the specific position calculation (before, it was spring layout). Now you can use any of the node position methods (circular, spiral, Kamada-Kawai, etc.) to get edge positions from the barycenters.

nwlandry:
@maximelucas, thanks for the review! A quick overview of my changes based on your comments:
* changed all of the dyads to be colored by the edge size (with "crest_r") by default
* Added error handling in a few places
* Added a new function `edge_positions_from_barycenters` as a more general replacement of the old functionality.
* Added an example of using `draw_bipartite` on an undirected hypergraph in Tutorial 5
* Added a few examples of the new `edge_positions_from_barycenters` in Tutorial 5 and In Depth 3.

At this point, I believe that I've addressed all your comments. Let me know if there are other things you would like me to change.

PR #496, opened 2023-11-30
Xgi data docs page

nwlandry:
Added a page for XGI-DATA on readthedocs.


maximelucas:
Looks great, thanks so much. I guess it's better to wait to merge the other PR first to check if the table gets populated correctly.

maximelucas:
This is not up to date. Did we miss this in the previous PR or will it be updated automatically?

Screenshot 2023-12-01 at 14 49 34

maximelucas:
> That doesn't match what I get when I call that function from the main branch of xgi or this branch. What branch are you using?

You were too fast, you replied faster than I updated my comment after realising I just needed to upgrade :)

maximelucas (approved):

PR #498, opened 2023-12-03
Create check-urls.yml

nwlandry:
Added a Github Action to check URLs.

### DIFF
--- a/.github/workflows/check-urls.yml
+++ b/.github/workflows/check-urls.yml
@@ -12,7 +12,7 @@ jobs:
       uses: urlstechie/urlchecker-action@master
       with:
         # A subfolder or path to navigate to in the present or cloned repository
-        subfolder: docs
+        subfolder: docs, xgi
 
         # A comma-separated list of file types to cover in the URL checks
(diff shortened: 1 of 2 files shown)


maximelucas:
This amazing, thanks Nich!! 
How do we know it works?
How does it work? For each PR it's gonna tell us if some urls are broken?

nwlandry:
> This amazing, thanks Nich!! How do we know it works? How does it work? For each PR it's gonna tell us if some urls are broken?

That's exactly right. For instance, I already fixed a link because of this failed run: https://github.com/xgi-org/xgi/actions/runs/7077776967

maximelucas:
Great then I'll approve. Does this mean all other urls work fine? Or we'll see at the next PR?

maximelucas (approved):

nwlandry:
> Great then I'll approve. Does this mean all other urls work fine? Or we'll see at the next PR?

I think that if I'm correct, the link that I fixed was the only issue. And to correct what I said, the action is triggered with every push.

Issue #514, opened 2024-02-29
Add more statistics functionality


nwlandry:
It would be nice to
* have `argmin` and `argmax` as functions
* be able to also supply a lambda function in addition to the standard string filterby modes ("gt", "geq", etc.).
in the `NodeStat`/`EdgeStat` interface.

nwlandry:
another thing that would be nice to have is `argsort`, where it returns a list/array/etc. of IDs sorted by their corresponding stat value.

PR #518, opened 2024-03-02
Add stats features

nwlandry:
Partially addresses #514.

### DIFF
--- a/xgi/stats/__init__.py
+++ b/xgi/stats/__init__.py
@@ -247,8 +247,8 @@ class IDStat:
     def argmin(self):
         """The ID corresponding to the minimum of the stat
 
-        When the maximal value is not unique, returns first
-        ID corresponding to the maximal value.
+        When the minimum value is not unique, returns first
+        ID corresponding to the minimum value.
 
         Returns
         -------
@@ -261,8 +261,8 @@ class IDStat:
     def argmax(self):
         """The ID corresponding to the maximum of the stat
 
-        When the minimum value is not unique, returns first
-        ID corresponding to the minimum value.
+        When the maximal value is not unique, returns first
+        ID corresponding to the maximal value.
 
         Returns
         -------
--- a/xgi/stats/__init__.py
+++ b/xgi/stats/__init__.py
@@ -244,6 +244,34 @@ class IDStat:
         arr = self.asnumpy()
         return spmoment(arr, moment=order) if center else np.mean(arr**order)
 
+    def argmin(self):
+        """The ID corresponding to the minimum of the stat
+
+        When the maximal value is not unique, returns first
+        ID corresponding to the maximal value.
+
+        Returns
+        -------
+        hashable
+            The ID to which the minimum value corresponds.
+        """
+        d = self.asdict()
+        return min(d, key=d.get)
+
+    def argmax(self):
+        """The ID corresponding to the maximum of the stat
+
+        When the minimum value is not unique, returns first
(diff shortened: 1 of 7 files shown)


xgi/stats/__init__.py
```diff
@@ -244,6 +244,34 @@ def moment(self, order=2, center=False):
         arr = self.asnumpy()
         return spmoment(arr, moment=order) if center else np.mean(arr**order)
 
+    def argmin(self):
+        """The ID corresponding to the minimum of the stat
+
+        When the maximal value is not unique, returns first
```
maximelucas, line 250: mimumum ? x2

xgi/stats/__init__.py
```diff
@@ -244,6 +244,34 @@ def moment(self, order=2, center=False):
...
+        hashable
+            The ID to which the minimum value corresponds.
+        """
+        d = self.asdict()
+        return min(d, key=d.get)
+
+    def argmax(self):
+        """The ID corresponding to the maximum of the stat
+
+        When the minimum value is not unique, returns first
```
maximelucas, line 264: maximal? x2

maximelucas:
Looks good thanks Nich! Do you think there's a way to use numpy's argmin/max to get all indices rather than just the first? Would we want that?

nwlandry:
> Looks good thanks Nich! Do you think there's a way to use numpy's argmin/max to get all indices rather than just the first? Would we want that?

That could be good, but unfortunately numpy does the same thing (returns the first max/min). I can look more into it though!

nwlandry:
I found this: https://stackoverflow.com/questions/25762332/how-to-get-all-the-keys-with-the-same-highest-value, but I'm not sure if we want to return an iterate. Let me know what you think!

maximelucas:
> That could be good, but unfortunately numpy does the same thing (returns the first max/min). I can look more into it though!

Ah true, my bad. 

> but I'm not sure if we want to return an iterate. Let me know what you think!

Yea I'm not sure either honestly. Maybe let's keep it as it is for now (it's pretty standard to return just the first occurrence), and see later? You documented it so it should be okay for now.

Ultimately, say we're looking for the nodes that have minimum degree. We might want to get all of them, not just the first. If we don't implement this by default, would there be a way to do it by hand by iterating for the user?

maximelucas (approved):

xgi/stats/__init__.py
nwlandry: Good catch!

xgi/stats/__init__.py
nwlandry: good catch!

nwlandry:
> > That could be good, but unfortunately numpy does the same thing (returns the first max/min). I can look more into it though!
> 
> Ah true, my bad.
> 
> > but I'm not sure if we want to return an iterate. Let me know what you think!
> 
> Yea I'm not sure either honestly. Maybe let's keep it as it is for now (it's pretty standard to return just the first occurrence), and see later? You documented it so it should be okay for now.
> 
> Ultimately, say we're looking for the nodes that have minimum degree. We might want to get all of them, not just the first. If we don't implement this by default, would there be a way to do it by hand by iterating for the user? And if we return all of them by default, we should check for format consistency with other stats to avoid potential problems.

I think that we can accomplish this in two ways: (1) add the ability to `argsort()` and (2) make a recipe to show how to get all max IDs. I added a note to #514 and opened an Issue #520 corresponding to the recipe.

PR #521, opened 2024-03-04
Stats argsort

nwlandry:
Added the ability to argsort statistics.


maximelucas (approved):

maximelucas:
Crazyyy 🚀 
I feel like we should showcase this (and other stats) somewhere in the docs but not sure where yet.

nwlandry:
Sounds good! I have a few things I want to merge and then let's do another release! I agree about the docs.

PR #522, opened 2024-03-04
added recipe for maximal indices

### LINKED ISSUE #520: Add recipe for returning all IDs corresponding to the maximum.
Implement something similar to this: https://stackoverflow.com/questions/25762332/how-to-get-all-the-keys-with-the-same-highest-value

nwlandry:
Fixes #520.


docs/source/api/recipes/recipes.ipynb
```diff
@@ -732,8 +732,8 @@
     "m, n = 10, 20\n",
```
maximelucas, line 732: <p>Read my mind. Could you add an example doing it with argsort too?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/xgi-org/xgi/pull/522/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='45'/>

nwlandry: It will be challenging to find the IDs corresponding to the max efficiently using `argsort`. You could do something like
```python
ids = H.nodes.degree.argsort()
max_ids = [i for i in id's if H.degree[i] == H.degree[ids[-1]]]
```

However, you could easily answer "what is the 10th largest degree node"?

Let me know what you think!

maximelucas: <p>True it might be less efficient. Was just thinking it might be nice to showcase argsort. I like your suggestion. Another variant idea: "What are the nodes that have the 10th largest degree?</p>

nwlandry: Done!

maximelucas (approved):

PR #530, opened 2024-04-03
Updated the return type of Katz centrality

nwlandry:
This PR accomplishes the following:
* `katz_centrality` now returns a dictionary to be consistent with all other centrality measures.
* The Katz centrality is now 1-normalized consistent with other functions
* `katz_centrality` is now a nodestat as well.

### DIFF
--- a/tests/algorithms/test_centrality.py
+++ b/tests/algorithms/test_centrality.py
@@ -158,36 +158,39 @@ def ratio(r, m, kind="CEC"):
         return r ** (1.0 / m)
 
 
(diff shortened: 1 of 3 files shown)


thomasrobiglio (approved):
Looks good! We should mention that the values that are returned are normalised, can you add that?

nwlandry:
> Looks good! We should mention that the values that are returned are normalised, can you add that?

I have mentioned that centralities are 1-normalized in the `Returns` field.

nwlandry:
Thanks for the review!!

PR #531, opened 2024-04-05
feat: hypergraph random edge shuffle

doabell:
This PR implements `random_edge_shuffle` for hypergraphs, as described in _[Configuration models of random hypergraphs](https://doi.org/10.1093/comnet/cnaa018)_, Philip S C., 2020.

Given two edges, the existing `double_edge_swap` swaps **two nodes**, whereas `random_edge_shuffle` puts **all nodes** in a bucket and shuffles them.
This preserves the edge sizes and node degrees.
If the same node appears in both hyperedges, then this is still true after reshuffling.

The user can also use this to shuffle two randomly selected hyperedges, if no edge IDs are specified.


tests/core/test_hypergraph.py
```diff
@@ -478,6 +478,61 @@ def test_double_edge_swap(edgelist1):
...
+
+    # all edge sizes are preserved
+    for edge_id in H._edge:
+        assert len(H._edge[edge_id]) == len(S._edge[edge_id])
+
+    # all node degrees are preserved
+    for node_id in H._node:
+        assert len(H._node[node_id]) == len(S._node[node_id])
+
+
```
nwlandry, line 535: Add another test to make sure that the bipartite structure is preserved.

xgi/core/hypergraph.py
```diff
@@ -958,6 +959,86 @@ def double_edge_swap(self, n_id1, n_id2, e_id1, e_id2):
...
+        """
+        if len(self._edge) < 2:
+            raise ValueError("Hypergraph must have at least two edges.")
+
+        # select two random edges
+        if e_id1 is None or e_id2 is None:
+            e_id1, e_id2 = random.sample(list(self._edge), 2)
+
+        # get nodes in those edges
+        nodes_e1 = self._edge[e_id1]
```
nwlandry, line 1010: I don't care either way on this, but perhaps you can eliminate the `node_` prefix for some of these variable names for brevity?

nwlandry (approved):
This looks great! Thanks, @doabell! Two small comments to address, but otherwise, I think that it is good to go.

tests/core/test_hypergraph.py
doabell: Added these in a286f66, thanks!

xgi/core/hypergraph.py
doabell: @nwlandry OK, I guess "edges are node lists" is self-explanatory. Fixed in a286f66.

doabell:
Resolved these comments, could you take another look? Thanks.

nwlandry:
Looks great! Thanks so much! Merging now.

PR #526, opened 2024-03-27
Made exact assortativity reproducible

nwlandry:
In the `degree_assortativity` function, when using the `exact` keyword, it was not reproducible because the two degrees were randomly shuffled prior to putting into the `numpy.corrcoef` function. NetworkX handles this by copying the pairs twice: $(u, v)$ and $(v,u )$ (See this [comment](https://github.com/networkx/networkx/blob/f44fdc4d12cd36907959637961865a0325383a91/networkx/algorithms/assortativity/pairs.py#L38)). This PR fixes this issue.

PR #533, opened 2024-04-15
Warn on unknown type of degree assortativity

doabell:
When calling `xgi.degree_assortativity()` with `exact=True` and an unknown `kind`, the code raises an `UnboundLocalError` that is hard to debug.
This PR fixes that behavior.

Unknown `kind`-s are handled in `_choose_degrees`, but after #526, when `exact=True`,  `_choose_degrees` is no longer called.

I also changed the message from
> Invalid choice function!

to

> Invalid type of degree assortativity!

to better describe the error.

(I found this because I made a typo in `kind=`.)


nwlandry (approved):
Thanks for this contribution, @doabell! Looks great!

PR #535, opened 2024-04-19
Added a new drawing recipe

### LINKED ISSUE #510: Returning node positions from `xgi.draw`
I want to use the same node positions to draw a few different hypergraphs on the same set of nodes. My lazy approach was to draw the first hypergraph with `xgi.draw`, then get node positions from the return and use them for the subsequent draw calls. However, it was surprisingly difficult to get the positions, despite the fact that `xgi.draw` computes positions and returns something called `node_collection`. The difficulty is that the return object is a `matplotlib.PathCollection`, so it doesn't contain `pos` in the same semantics (at least as far as I can tell). One might be able to use `PathCollection.get_offsets()` or similar to get them, I'm not sure.

This is not urgent at all because really I should compute the layout first, then use the same positions throughout, which is also what the example in the documentation points towards. Not trying to encourage bad design, but it still might be nice to return `pos` explicitly for convenience. I think the main question is whether adding another value to the return will somehow break other pieces of code or be otherwise inconvenient.

May relate to #280 in the future.

nwlandry:
Added a recipe showing how to reuse the same nodal positions to plot many hypergraphs. Fixes #510.


docs/source/api/recipes/recipes.ipynb
```diff
@@ -10,7 +10,7 @@
   },
```
tlarock, line 10: <pre class="ql-syntax" spellcheck="false">Line #14.    pos = xgi.barycenter_spring_layout(H, seed=2)
</pre><p>Can we add a comment here that clearly states this is where the magic is happening? i.e. "#&nbsp;Compute positions for nodes that will be used for all plots"</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/xgi-org/xgi/pull/535/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='54'/>

tlarock, line 10: <pre class="ql-syntax" spellcheck="false">Line #52.    pos = xgi.pca_transform(xgi.pairwise_spring_layout(H, seed=3))
</pre><p>Actually pos is redefined here - was it used in the meantime? Do we need two different pos variables?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/xgi-org/xgi/pull/535/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='54'/>

nwlandry: You're right --- I deleted the duplicate

nwlandry: Done.

nwlandry:
@tlarock I believe I addressed all of your comments!

tlarock (approved):
Looks good!

PR #538, opened 2024-05-06
added error handling for JSON duplicate IDs

nwlandry:
If we construct a hypergraph as
```python
>>> import xgi
>>> H = xgi.Hypergraph()
>>> H.add_nodes_from(["2", 2])
>>> xgi.write_json(H, "test.json")

>>> H2 = xgi.read_json("test.json")
>>> H2.nodes
NodeView(("2"))
```
These will get mapped to a single node when we cast IDs to strings for the JSON. This will raise an error instead.


thomasrobiglio (approved):
Thank you @nwlandry, looks good!

nwlandry:
Thanks so much, @thomasrobiglio!!

PR #539, opened 2024-05-13
Update license and contributing

nwlandry:
Made some small updates:
* Added 2024 to the license files
* Removed now-unnecessary acknowledgments in the license file
* Simplified contribution guide.


thomasrobiglio:
@nwlandry let me know when you are done and I will approve this. 

(Are we removing the HypernetX and NetworkX acknowledgments simply because we no longer have code copied from them or for other reasons? thx)

nwlandry:
> @nwlandry let me know when you are done and I will approve this. 
> 
> (Are we removing the HypernetX and NetworkX acknowledgments simply because we no longer have code copied from them or for other reasons? thx)

@thomasrobiglio ready for review! At this point, we have rewritten the code that we were using from other libraries (I could be wrong about this) so I thought we could simplify our license.

thomasrobiglio (approved):

thomasrobiglio:
Good to go! thank you @nwlandry 

> At this point, we have rewritten the code that we were using from other libraries (I could be wrong about this) so I thought we could simplify our license.

That's what I thought, thank you :)

PR #540, opened 2024-05-20
Added the ability for XGI to load data collections

nwlandry:
Refactored the following to handle collections of data:
* `load_xgi_data`
* `download_xgi_data`
* `write_json`
* `read_json`

Also did the following:
* Moved the contents of `write_json` to `to_hypergraph_dict`
* Renamed `dict_to_hypergraph` to `from_hypergraph_dict`

### DIFF
--- a/xgi/convert/hypergraph_dict.py
+++ b/xgi/convert/hypergraph_dict.py
@@ -1,12 +1,60 @@
 """Method for converting from a standardized dictionary."""
 
+from collections import Counter
+
 from ..exception import XGIError
 from ..generators import empty_hypergraph
+from ..utils import get_network_type
 
-__all__ = ["dict_to_hypergraph"]
+__all__ = ["to_hypergraph_dict", "from_hypergraph_dict"]
 
 
-def dict_to_hypergraph(data, nodetype=None, edgetype=None, max_order=None):
+def to_hypergraph_dict(H):
+    data = {}
+    data["type"] = get_network_type(H)
+    # name always gets written (default is an empty string)
+    data["hypergraph-data"] = {}
+    data["hypergraph-data"].update(H._hypergraph)
+
+    # get node data
+    try:
+        data["node-data"] = {str(idx): H.nodes[idx] for idx in H.nodes}
+
+        if len(data["node-data"]) != H.num_nodes:
+            dups = [
+                item
+                for item, count in Counter([str(n) for n in H.nodes]).items()
+                if count > 1
+            ]
+            raise XGIError(
+                f"When casting node IDs to strings, ID(s) {', '.join(dups)} have conflicting IDs!"
+            )
+    except KeyError:
+        raise XGIError("Node attributes not saved!")
+
+    try:
+        data["edge-data"] = {str(idx): H.edges[idx] for idx in H.edges}
+
+        if len(data["edge-data"]) != H.num_edges:
+            dups = [
+                item
+                for item, count in Counter([str(n) for n in H.edges]).items()
+                if count > 1
+            ]
+            raise XGIError(
+                f"When casting edge IDs to strings, ID(s) {', '.join(dups)} have conflicting IDs!"
+            )
+    except KeyError:
+        raise XGIError("Edge attributes not saved!")
+
+    # hyperedge dict
+    data["edge-dict"] = {
+        str(idx): [str(n) for n in H.edges.members(idx)] for idx in H.edges
+    }
+    return data
+
+
+def from_hypergraph_dict(data, nodetype=None, edgetype=None, max_order=None):
     """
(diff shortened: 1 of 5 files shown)


xgi/convert/hypergraph_dict.py
```diff
@@ -1,12 +1,60 @@
...
+from collections import Counter
+
 from ..exception import XGIError
 from ..generators import empty_hypergraph
+from ..utils import get_network_type
+
+__all__ = ["to_hypergraph_dict", "from_hypergraph_dict"]
+
+
+def to_hypergraph_dict(H):
```
maximelucas, line 12: missing docstring

maximelucas:
Idea sounds nice! Can you elaborate a bit, what do you mean by collection and what's your intended workflow with this?

xgi/convert/hypergraph_dict.py
```diff
@@ -1,12 +1,60 @@
...
+    data = {}
+    data["type"] = get_network_type(H)
```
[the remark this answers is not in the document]
nwlandry: So eventually, we will specify the "type" of the json: "collection", "hypergraph", "dihypergraph", or "simplicialcomplex", but for now, if it is not specified, it loads as a hypergraph. But good point. I'll add in some details in the docstring.

nwlandry:
So my original intent was to figure out how to store collections similar to [this one](https://hyperbard.net/). I opened Issue [20](https://github.com/xgi-org/xgi-data/issues/20) in XGI-DATA for this issue. The choices in my mind were (1) make a huge file containing all the hypergraphs or (2) continue to have a file for each hypergraph, but in addition have a *collection* file specifying metadata of the collection as well as where to find each dataset in the collection. There might be a more elegant way, but this is the best I've come up with so far. Note that with this PR, you can call `xgi.load_xgi_data("hyperbard")` and load the hyperbard collection.

xgi/convert/hypergraph_dict.py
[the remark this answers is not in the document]
maximelucas: Either than getting it from the dataset name (which may make those names long), could we get it from a variable/key in the file? Or maybe we need it before loading.

xgi/convert/hypergraph_dict.py
nwlandry: fixed. Thanks!

xgi/convert/hypergraph_dict.py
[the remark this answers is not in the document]
nwlandry: So the new dict format has "type" as a key, and that type is assigned from the class name.

tests/readwrite/test_xgi_data.py
```diff
@@ -48,10 +48,29 @@ def test_load_xgi_data(capfd):
     assert "email-enron" in out
     assert "congress-bills" in out
 
+    # test collection
+    collection = load_xgi_data("hyperbard")
+    assert len(collection) == 37
```
maximelucas, line 53: this is the number of datasets in the collection right?

nwlandry: Correct!

maximelucas:
Just to check, in the end you chose option (2) right? And this is implemented as a single record in Zenodo for the collection containing many .json files (one for each dataset). 

`xgi.load_xgi_data("hyperbard")` loads a dict of datasets, so xgi.load_xgi_data("hyperbard")["coriolanus"] should load a single dataset if I understand correctly.

This is very nice and good for now thanks Nich. 

As I mentioned in our call, I'm wondering if don't want to be able to access them directly from something like `xgi.load_xgi_data("hyperbard-coriolanus")` to be able to iterate over all datasets in XGI-data. They could also appear in the list we get with `xgi.load_xgi_data()` this way? This could be a future PR too. 

Btw, related to #548 I don't think the loading function is in the online docs.

nwlandry:
Thanks for the review! I completely agree, but this will take a bit longer to figure out how to implement. Let me know if you'd rather I wait to implement accessing/viewing individual datasets to merge or if I should open an issue on this. What do you mean when you say that the loading function isn't in the online docs?

maximelucas:
> Let me know if you'd rather I wait to implement accessing/viewing individual datasets to merge or if I should open an issue on this.

Yes I think we can just open an issue about this!

> What do you mean when you say that the loading function isn't in the online docs?

I mean that `xgi.load_xgi_data("<dataset_name>")` is not documented anywhere (except from its own docstring). 
As far as I can tell it only appears here, but without all its arguments: https://xgi.readthedocs.io/en/stable/xgi-data.html.

maximelucas (approved):

nwlandry:
Thanks!! I added two issues from your comments.

Issue #379, opened 2023-06-01
Inheritance for hypergraph classes


nwlandry:
We should have `Hypergraph` inherit from `DiHypergraph` or vice-versa as well as `IDView`/`DiIDView` to avoid overly duplicated code.

nwlandry:
Some thoughts on this:

We should start by implementing a `HigherOrderSystem` (HOrS) class with some basic functionality.

All higher-order networks have:
* nodes (same for all networks)
* interactions (the *type* of interaction will determine how it is implemented)
For example, interactions can be modeled as a:
* set of frozen sets for simple hypergraphs, or a dictionary of sets with some checking.
* dictionary of [sorted lists](https://grantjenks.com/docs/sortedcontainers/) for loopy hypergraphs
* dictionary of dictionaries of the form `{"out": {<tail>}, "in":{<head>}}` for dihypergraphs
* dictionary of sets for multihypergraphs.

My thought is that we can implement all of the general things except for the edges in the base class and write interaction type-specific functionality in the inherited classes.

nwlandry:
PR #541 implements the third item. Item 2 is fairly low priority, but when we implement this, `Counter` might be preferable to a sorted list.

PR #541, opened 2024-05-26
Simplify dihypergraph

nwlandry:
This PR changes the internals of `DiHypergraph` so that it is more consistent with the other classes. This is an attempt to start addressing #379. This PR
* Replaces `_node_in` and `_node_out` with `_node` where `_node[n]` is of the form `{"in": {tail}, "out": {head}}`
* Replaces `_edge_in` and `_edge_out` with `_edge` where `_edge[n]` is of the form `{"in": {}, "out": {}}`
* Removes `DiIDView` since now we can simply inherit from `IDView`.
* Moves `DiNodeView` and `DiEdgeView` to `views.py` and deletes `diviews.py`
* Moves tests for directed views to `test_views.py`
* Removes unnecessary imports in the unit tests.


nwlandry:
@maximelucas - when do you think you will be able to take a look at this? Thanks!

xgi/core/views.py
```diff
@@ -790,3 +792,375 @@ def maximal(self, strict=False):
                         max_edges.update(dups[frozenset(e)])
 
         return self.from_view(self, bunch=max_edges)
+
+
+class DiNodeView(IDView):
+    """An DiIDView that keeps track of node ids.
```
maximelucas, line 798:
```suggestion
    """A DiIDView that keeps track of node ids.
```

xgi/stats/dinodestats.py
```diff
@@ -248,30 +248,30 @@ def out_degree(net, bunch, order=None, weight=None):
     >>> import xgi
     >>> H = xgi.DiHypergraph([[{1, 2}, {5, 6}], [{4}, {1, 3}]])
     >>> H.nodes.out_degree.asdict()
-    {1: 1, 2: 0, 5: 1, 6: 1, 4: 0, 3: 1}
+    {1: 1, 2: 1, 5: 0, 6: 0, 4: 1, 3: 0}
```
maximelucas, line 251: was this a mistake that was not caught by our automated checks?

xgi/core/views.py
```diff
@@ -790,3 +792,375 @@ def maximal(self, strict=False):
...
+                return [self._id_dict[key]["in"].copy() for key in self]
+            else:
+                raise XGIError(f"Unrecognized dtype {dtype}")
+
+        if e not in self:
+            raise IDNotFound(f'ID "{e}" not in this view')
+
+        return self._id_dict[e]["in"].copy()
+
+    def sources(self, e=None, dtype=list):
```
maximelucas, line 1147: Did we also have these duplicate methods with different names?

xgi/core/views.py
```diff
@@ -790,3 +792,375 @@ def maximal(self, strict=False):
...
+            Edge members.
+        dict (if dtype is dict)
+            Edge members.
+        set (if e is not None)
+            Members of edge e.
+
+        The members of an edge are the union of
+        its head and tail sets.
+
+        The
```
maximelucas, line 1030:
```suggestion

```

xgi/core/dihypergraph.py
```diff
@@ -328,9 +322,8 @@ def add_node(self, node, **attr):
         If node is already in the dihypergraph, its attributes are still updated.
 
         """
-        if node not in self._node_in:
-            self._node_in[node] = set()
-            self._node_out[node] = set()
+        if node not in self._node:
```
maximelucas, line 325: will this ever be true? since `self._node` should a 2-key dict, only `"in"` and `"out"` will be `in self._node`?

maximelucas: It if it never true, the new "empty" 2-key dict will be create each time

maximelucas: Oh wait, there's one 2-key dict per node? I thought it was `H._nodes["in"][node_id]` but it's `H._node[node_id]["in"]`.

tests/core/test_dihypergraph.py
```diff
@@ -280,7 +280,7 @@ def test_add_edges_from_format4():
     H1 = xgi.DiHypergraph([({1, 2}, {3}), ({2, 3, 4}, {1})])
     with pytest.warns(Warning):
         H1.add_edges_from([(({0, 1}, {2}), 0, {"color": "red"})])
-    assert H1._edge_out == {0: {1, 2}, 1: {2, 3, 4}}
+    assert {i: e["out"] for i, e in H1._edge.items()} == {0: {3}, 1: {1}}
```
maximelucas, line 283: this looks more complicated to access the edges out than before (`H1._edge_out`). Could we do something like `H._edge["out"]` ?

xgi/stats/dinodestats.py
nwlandry: Correct. We had in and out-degrees flipped.

xgi/core/views.py
nwlandry: Fixed.

xgi/core/views.py
nwlandry: Yes, this was a request when I originally implemented dihypergraphs for explainability. Would you like me to remove them?

tests/stats/test_dinodestats.py
```diff
@@ -17,15 +13,16 @@ def test_degree(diedgelist2):
 
 def test_in_degree(diedgelist2):
     H = xgi.DiHypergraph(diedgelist2)
-    degs = {0: 1, 1: 2, 2: 2, 4: 1, 3: 1, 5: 0}
+    degs = {0: 0, 1: 0, 2: 1, 3: 0, 4: 2, 5: 1}
```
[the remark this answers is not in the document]
nwlandry: yes. The _edge_out corresponds with _node_in and this was incorrect before. I double-checked these values.

xgi/core/dihypergraph.py
nwlandry: Correct. Yeah, I made the design decision to index by nodes first.

tests/core/test_dihypergraph.py
nwlandry: So this won't change the API, but I get what you mean. Because I chose an edge/node ID-first approach, when trying to get just the inputs or outputs, it will require some kind of list/dict comprehension.

nwlandry:
> I've had a look now, thanks! I left comments which are mostly questions for my own understanding.
> 
> The main one may be about about the choice between having `H._node[node_id]["in"]` and not `H._nodes["in"][node_id]`. I'm just wondering which one will be easier to generalise (don't know what I would choose). The first (current) option allows the call to H._[node_id] to be valid like before, but make it a bit more lengthy to access all "in" nodes. By the way, I'm just realising, what is an in-node again? 😛
> 
> If the API for the user is unchanged it's not needed, but otherwise we should document this somewhere.

Thanks for the review! I've thought quite a bit about your suggestion and I'd like to leave the structure (`H._edge[e]["in]` for example) as is for the following reasons:
* It seems natural for an edge ID to come with a list of all of its in- and out-connections
* The goal of this PR is to move us toward full inheritance and I think if we wanted to overload some algorithm definitions, it would be more natural for an edge ID to return a full edge, whether undirected or directed.

As far as your question goes, IMO it's most natural to think about it in the bipartite sense. If a node is specified as an input (in the tail) in a directed edge, that edge will add to its out-degree.

maximelucas:
Thanks!

PR #542, opened 2024-05-27
Higher-order base class

nwlandry:
This adds a base class. This is currently unused, but other classes will inherit from it in the future.


nwlandry:
Closed in favor of #555

PR #550, opened 2024-06-07
Move tutorials to top level

nwlandry:
This PR uses nbsphinx-link to move the tutorials back to the top level so that not only can nbsphinx render the notebooks for the website, but users can easily access the notebooks as well. This PR also fixes a few small extraneous things, namely:
* Now the encapsulation_dag page has the correct name
* We no longer use `pkg_resources` to determine the version and the contributing guide now reflects that.


thomasrobiglio (approved):
great, thank you @nwlandry

nwlandry:
Thanks for the review, @thomasrobiglio! :D

Issue #537, opened 2024-04-22
add back edgecolor argument to draw hulls with white face but colored contour


maximelucas:
To get something like this:

<img src="https://github.com/xgi-org/xgi/assets/7493360/cc4d909f-edcf-4659-ac3a-672417320d9d" width="35%">

nwlandry:
Can you comment on what you would like the call to be to draw the above figure? Like a minimal working example.

maximelucas:
I started a branch a few days ago, I'll push soon!

nwlandry:
Great! Looking forward to it!

PR #553, opened 2024-06-11
add back edgecolor to drawing hull 

maximelucas:
Goal is to solve #537


maximelucas:
Closing, replaced by #575

Issue #557, opened 2024-07-03
to_line_graph function leads to KeyError


pgberlureau:
Hello,

I think it would be better to iterate over 

```combinations(edge_label_dict.keys(), 2)```

instead of 

```combinations(H.edges.members(), 2)``` 

It solves a KeyError i have when i call to_line_graph on the following hypergraph:

```h = xgi.Hypergraph({0: [(6.019592999999995, 47.2350647), (6.019592999999853, 47.2350647), (6.01959299999999, 47.2350647), (6.019592999999993, 47.2350647), (6.0195929999999755, 47.2350647)], 1: [(6.019592999999995, 47.2350647), (6.019592999999853, 47.2350647), (6.01959299999999, 47.2350647), (6.019592999999993, 47.2350647), (6.0195929999999755, 47.2350647)]})```

thomasrobiglio:
Thank you @pgberlureau for pointing this out! 
I replicated your error, and from a first look (I quickly implemented it in my local version of `xgi`), it seems that the solution you suggest works well. 

We will soon open a PR to solve this bug, if you wish to do it yourself you are more than welcome to do so :)

pgberlureau:
No problem, I can do it :)

PR #558, opened 2024-07-03
Get rid of KeyErrors in to_line_graph 

pgberlureau:
Use edge_label_dict.values() as nodes and get edge1 and edge2 from edge_label_dict.keys() to avoid KeyError.
(Fix issue #557)


xgi/convert/line_graph.py
```diff
@@ -55,16 +55,16 @@ def to_line_graph(H, s=1, weights=None):
 
     edge_label_dict = {tuple(edge): index for index, edge in H._edge.items()}
 
-    LG.add_nodes_from(H.edges)
+    LG.add_nodes_from(edge_label_dict.values())
 
-    for edge1, edge2 in combinations(H.edges.members(), 2):
+    for edge1, edge2 in combinations(edge_label_dict.keys(), 2):
```
nwlandry, line 60: If you just use `edge_label_dict` it will use the keys by default.

nwlandry:
@pgberlureau thanks so much for finding this bug and for your contribution! Two quick questions:
1. Can you add a corresponding unit test in `tests/convert/test_line_graph.py` called `test_557` so that we don't break this in the future?
2. I realized that we are eliminating duplicate edges by making the tuples of the edges the keys of the label dict. Perhaps we should do something like this:
```python
for e1, e2 in combinations(H._edge, 2):
    edge_intersect = len(H._edge[e1].intersection(H._edge[e2]))
    if edge_intersect >= s:
        LG.add_edge(e1, e2)
```

nwlandry:
I saw that the unit tests are failing due to an unrelated issue which I will fix. Thanks again!!

pgberlureau:
Hey,

1. yes i can add the test, no problem :)
2. You are right, only thing is that in the obtained nx graph, nodes are now indices and the original corresponding hyperedges are  attributes of those nodes. This is not a problem to me but is this really the desired behaviour ?

nwlandry:
> Hey,
> 
> 1. yes i can add the test, no problem :)
> 2. You are right, only thing is that in the obtained nx graph, nodes are now indices and the original corresponding hyperedges are  attributes of those nodes. This is not a problem to me but is this really the desired behaviour ? (Or I didn't got what you meant sorry)

What I was realizing is that using tuples may be unexpectedly buggy. Because the order of `set` is not guaranteed, when we cast to tuples, we may be getting different permutations of the edges, e.g., `(1, 2, 3)` and `(3, 1, 2)`. I propose that we fix this using the code that I suggested, or, at the very least, putting a `sorted` statement inside the casting to tuple.

My worry about using the edges as keys is that the line graph will be the same with or without multiedges, so if a hypergraph is not a simple hypergraph, then we're not getting the accurate linegraph.

nwlandry (approved):
This looks awesome, @pgberlureau! Thanks so much! Let me know if you're done, and if so, I'll merge.

pgberlureau:
I have nothing to add, thank you for your work !

PR #559, opened 2024-07-03
Fix scipy error

nwlandry:
This fixes an error with the Scipy up-version related to `asfptype()`. In addition, it does the following:
* up-versions pylint
* Removes an unnecessary import
* Removes a warning in the unit tests


thomasrobiglio (commented):
Great, thank you @nwlandry. I was about to comment on the try/except but you already removed that.

thomasrobiglio (approved):

PR #560, opened 2024-07-03
Rename `_hypergraph` to `_net_attr`

nwlandry:
This PR renames the `_hypergraph` internal variable to `_net_attr` to be more descriptive and general. This change does not affect the API.

### DIFF
--- a/xgi/core/hypergraph.py
+++ b/xgi/core/hypergraph.py
@@ -1,4 +1,5 @@
 """Base class for undirected hypergraphs."""
+
 import random
 from collections import defaultdict
 from collections.abc import Hashable, Iterable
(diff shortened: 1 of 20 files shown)


nwlandry:
@maximelucas --- do you mind taking a quick look? This PR is solely renaming `_hypergraph` to `_net_attr`

xgi/core/hypergraph.py
```diff
@@ -66,11 +67,12 @@ class Hypergraph:
...
 
     """
+
     _node_dict_factory = IDDict
     _node_attr_dict_factory = IDDict
-    _hyperedge_dict_factory = IDDict
-    _hyperedge_attr_dict_factory = IDDict
-    _hypergraph_attr_dict_factory = dict
+    _edge_dict_factory = IDDict
+    _edge_attr_dict_factory = IDDict
```
maximelucas, line 74: what's the reasoning between _hyperedge -> _edge?
i vaguely remember an old conversation where people wanted to keep hyperedge to make it explicit that's it's not pairwise, but might be wrong

maximelucas (approved):

xgi/core/hypergraph.py
nwlandry: I changed it so that it matches the convention across the rest of the library e.g., `H.edges`, `edge_fc`, etc.

PR #561, opened 2024-07-11
Fix #552

### LINKED ISSUE #552: Possible error in stats
In this line in the constructor of IDStat, there is this line:
```python
self.kwargs = {} if args is None else kwargs
```
Shouldn't it be `kwargs`?

nwlandry:
This fixes #552. I can't create an example where the prior implementation will create issues (there's a weird overlap of issues which leads to the correct behavior), but I think that this is more explainable.


nwlandry:
@leotrs --- is this something you're able to review? (Same question for #562)

maximelucas (approved):

PR #565, opened 2024-07-26
Allow empty edges

### LINKED ISSUE #403: Consistency in dealing with empty edges
Currently, some XGI methods support empty edges and some do not.

Ones that do:
* `dual()`: the dual of a hypergraph with isolated nodes must have empty edges.
* `add_edges_from()` so long as the empty edges is not first and if it is, it raises an IndexError as it tries to access `list(first_edge)[0]` (in the case of a list of sets)

Ones that do not:
* `add_edge()`: this raises an XGIError for an empty edge.
* `remove_nodes_from()`, `remove_node()`, and `remove_node_from_edge()`: these functions delete edges if removing a node creates an empty edge.

nwlandry:
This PR enables empty edges. This has long been an inconsistency, since `add_edges_from` allowed empty edges and, for example, `dual` could create empty edges. This PR does the following:
* removes the exception raised for empty edges from the `add_edge` methods in `Hypergraph` and `DiHypergraph` and from `add_simplex` in `SimplicialComplex`.
* adds an `allow_empty` keyword in the `remove_node` methods to handle discarding empty edges created from weakly removing nodes.
* makes all the keywords in `remove_node` accessible as options in the `remove_nodes_from` methods for consistency.

This fixes #403.


nwlandry:
@tlarock and @thomasrobiglio --- let me know if you have the bandwidth to review this. The URL workflow fails, but this is corrected in an already-merged PR.

thomasrobiglio:
Hi @nwlandry, thank you for this. I am offline for these two weeks, I will review once I'm back.

nwlandry:
Enjoy your vacation! That sounds good.

xgi/core/dihypergraph.py
```diff
@@ -406,17 +408,26 @@ def remove_node(self, n, strong=False):
...
 
         Parameters
         ----------
         nodes : iterable
             An iterable of nodes.
+        strong : bool (default False)
+            Whether to execute weak or strong removal.
+        remove_empty : bool, optional
+            Whether to remove empty edges (0 members in both head and tail).
+            By default, False.
```
tlarock, line 430: Same question - default appears to be `True`, but listed in docs as `False`

xgi/core/hypergraph.py
```diff
@@ -1136,6 +1139,8 @@ def remove_node_from_edge(self, edge, node):
             The edge ID
         node : hashable
             The node ID
+        remove_empty : bool, optional
+            Whether empty edges are removed. By default, True.
```
tlarock, line 1143: Please check that all docstrings match signatures

xgi/core/dihypergraph.py
```diff
@@ -375,9 +375,11 @@ def remove_node(self, n, strong=False):
         ----------
         n : node
             A node in the dihypergraph
-
         strong : bool (default False)
             Whether to execute weak or strong removal.
+        remove_empty : bool, optional
+            Whether to remove empty edges (0 members in both head and tail).
+            By default, False.
```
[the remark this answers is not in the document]
nwlandry: Good catch! I didn't want to break the API, so the default should be True. I'll fix that.

tlarock (approved):

nwlandry:
@thomasrobiglio --- are you still able to review this? No worries if not!

Issue #566, opened 2024-08-01
strange datatype of unique_edge_size()


maximelucas:
Running 
```python
H = xgi.load_xgi_data("email-enron")
xgi.unique_edge_sizes(H)
```

yields 

```python
[np.int64(1),
 np.int64(2),
 np.int64(3),
...
 np.int64(36),
 np.int64(37)]
```

I don't remember these elements being `np.int64`. There's probably nothing wrong with it but it looks unnecessary.
This might be due to the new numpy version 2.0?

nwlandry:
Yes, that's correct! If you want to make a quick PR casting the values to int and adding a short test confirming the data type, I'll approve it!

PR #567, opened 2024-08-01
fix: #566

maximelucas:
fix: #566 by casting edge sizes to a list instead of a numpy array


nwlandry (approved):

nwlandry:
Good catch and thanks so much for this PR, @maximelucas!!

PR #569, opened 2024-08-05
simplify cleanup functions

### LINKED ISSUE #568: homogenise large_connected_hypergraph implementations
To get the GCC, we have the `large_connected_hypergraph()` function:

 https://github.com/xgi-org/xgi/blob/e01ddd55fc288219ce12ea622091814691c8e866/xgi/algorithms/connected.py#L226-L230

And we also get the GCC in `cleanup()` when `connected=True`, but it's implemented slightly differently:

https://github.com/xgi-org/xgi/blob/e01ddd55fc288219ce12ea622091814691c8e866/xgi/core/hypergraph.py#L1471-L1474

If there's no reason for this, I think we should make them the same, or even better, call the first one in the second.
It will be easier to maintain and check that both do the same.

nwlandry:
This PR condenses unnecessary code and fixes #568.

### DIFF
--- a/xgi/core/hypergraph.py
+++ b/xgi/core/hypergraph.py
@@ -1462,38 +1462,38 @@ class Hypergraph:
 
         """
         if in_place:
-            H = self
+            _H = self
         else:
-            H = self.copy()
+            _H = self.copy()
 
         if not multiedges:
-            H.merge_duplicate_edges()
+            _H.merge_duplicate_edges()
         if not singletons:
-            H.remove_edges_from(H.edges.singletons())
+            _H.remove_edges_from(_H.edges.singletons())
         if not isolates:
-            H.remove_nodes_from(H.nodes.isolates())
+            _H.remove_nodes_from(_H.nodes.isolates())
         if connected:
             from ..algorithms import largest_connected_hypergraph
 
-            largest_connected_hypergraph(H, in_place=True)
+            largest_connected_hypergraph(_H, in_place=True)
         if relabel:
             from ..utils import convert_labels_to_integers
 
-            temp = convert_labels_to_integers(H).copy()
+            temp = convert_labels_to_integers(_H).copy()
 
             nn = temp.nodes
             ee = temp.edges
 
-            H.clear()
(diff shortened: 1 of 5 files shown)


xgi/core/hypergraph.py
```diff
@@ -1462,49 +1462,38 @@ def cleanup(
...
-            if not isolates:
-                H.remove_nodes_from(H.nodes.isolates())
-            if connected:
-                from ..algorithms import largest_connected_component
-
-                H.remove_nodes_from(H.nodes - largest_connected_component(H))
-            if relabel:
-                from ..utils import convert_labels_to_integers
-
-                H = convert_labels_to_integers(H)
```
maximelucas, line 1505: Was that wrong? Since we now use the longer relabeling version in both cases

xgi/core/hypergraph.py
```diff
@@ -1462,49 +1462,38 @@ def cleanup(
...
+
+            temp = convert_labels_to_integers(_H).copy()
+
+            nn = temp.nodes
+            ee = temp.edges
+
+            _H.clear()
+            _H.add_nodes_from((n, deepcopy(attr)) for n, attr in nn.items())
+            _H.add_edges_from(
+                (e, id, deepcopy(temp.edges[id]))
```
maximelucas, line 1490: can use ee instead of temp.edges

maximelucas (approved):

maximelucas:
Looks good thanks! Just a few questions

xgi/core/hypergraph.py
nwlandry: Good catch! I have fixed this and moved all of the heavy lifting to the `convert_labels_to_integers` function.

xgi/core/hypergraph.py
nwlandry: This is fixed by moving all of the logic to the `convert_labels_to_integers` function.

nwlandry: So this becomes a one-liner again.

nwlandry:
Thanks for the review, @maximelucas! I moved the in-place vs. new network logic to the `convert_labels_to_integers` function and I think it streamlines things.

PR #572, opened 2024-08-22
HIF functionality (new)

nwlandry:
This PR adds:
* read/write capability of json files according to the [HIF standard](https://github.com/pszufe/HIF_validators/blob/main/schemas/hif_schema_v0.1.0.json).
* The `__add__` method for `IDDict`
* Additional bipartite edgelist functionality.
This replaces PR #563


maximelucas:
What's the difference between the HIF and our own format?

nwlandry:
> What's the difference between the HIF and our own format?

Lots of things. The HIF standard is being developed with 3 other higher-order libraries (and hopefully more in the near future) which means that this will be much more standard. Secondly, HIF has far more features than the format used in XGI-DATA (empy edges, edge-dependent node attributes, directed hypergraphs, etc.) so my hope is to finalize the format with the group developing this and then convert the files in XGI-DATA to this sometime down the road.

maximelucas:
Great, sounds good, thanks a lot.

PR #573, opened 2024-08-25
Switch from setup.py to pyproject.toml

nwlandry:
According to [PEP 621](https://peps.python.org/pep-0621/), projects are encouraged to migrate to a pyproject.toml file.


README.md
```diff
@@ -58,9 +58,38 @@ pip install -e .["all"]
...
+```bash
+$ pip install -U -r requirements/default.txt
+```
+
+### Running the tests
+
+```bash
+$ pip install -U -r requirements/default.txt
+$ pip install -U -r requirements/test.txt
+```
```
maximelucas, line 92: I feel like this is precise and good but it would take a lot of space in our Readme for something that is gonna be useful to only rare users, and is probably somewhat standard pip behavior. Could we for example reduce by only listing the 6 options and link to a python page explaining this?

pyproject.toml
```diff
@@ -0,0 +1,104 @@
...
+]
+dependencies = [
+    "numpy>=1.20",
+    "scipy>=1.8",
+    "pandas>=1.3",
+    "networkx>=2.7",
+    "requests>=2.0",
+    "matplotlib>=3.4",
+    "seaborn>=0.10",
+]
```
maximelucas, line 53: we don't have more? can we import them directly from the requirement file?

pyproject.toml
```diff
@@ -0,0 +1,104 @@
...
+    "github-changelog",
+]
+test = [
+    "pytest>=7.2",
+    "pytest-cov>=4.0",
+]
+tutorial = [
+    "jupyter>=1.0",
+    "ipython<=8.12.0",
+]
```
maximelucas, line 97: same question about importing from requirement files

pyproject.toml
```diff
@@ -0,0 +1,104 @@
+[build-system]
+build-backend = "setuptools.build_meta"
+requires = ["setuptools>=61.2", "wheel"]
+
+[project]
```
maximelucas, line 5: do we not need a version number like we had in setup.py (imported from another file)

maximelucas:
Thanks a lot Nich! Looks good. Just a few questions above.

README.md
nwlandry: I like this idea. I think I can move it to the readme in the requirements folder.

pyproject.toml
nwlandry: I copied this directly from `default.txt`, so this is the accurate list of requirements. In PEP621, the goal is for all of the project info to be contained in pyproject.toml and use this file to generate all peripheral files. See for example what networkx [does](https://github.com/networkx/networkx/blob/main/tools/generate_requirements.py). In fact, I think that in the future, we might be able to remove the requirements files all together since I can't think of an example where one wishes to only import the dependencies and not the package itself. But that's a future problem, I think.

pyproject.toml
nwlandry: See above. I will add a `generate_requirements.py` file as NetworkX does.

pyproject.toml
nwlandry: This is now handled with
```
[tool.setuptools.dynamic.version]
attr = "xgi.__version__"
```

pyproject.toml
maximelucas: Ah I see so this should ultimately replace the requirement files.

pyproject.toml
maximelucas: Sounds like a good idea so that they're defined and modified in a single place

pyproject.toml
maximelucas: Thanks I missed that

README.md
nwlandry: Done.

nwlandry:
@maximelucas Should have addressed all of your comments, thanks!

pyproject.toml
nwlandry: Yep!

requirements/tutorial.txt
```diff
@@ -1,4 +1,4 @@
+# Generated via tools/generate_requirements.py and pre-commit hook.
+# Do not edit this file; modify pyproject.toml instead.
 jupyter>=1.0
-matplotlib>=3.3
-seaborn>=0.10
-ipython<=8.12.0
\ No newline at end of file
+ipython<=8.12.0
```
maximelucas, line 4: matplotlib not needed anymore because already in the default?

nwlandry: Correct, it was redundant before.

maximelucas (approved):

PR #575, opened 2024-08-30
Added `edge_ec` argument in `draw` to specify edge colors

### LINKED ISSUE #537: add back edgecolor argument to draw hulls with white face but colored contour

maximelucas:
This should fix #537 

Excepted usage is mostly when using `draw()`  with `hull=True` and white facecolors for edges with `edge_fc="white"`.
Colors can be specified in all usual formats (same as `edge_fc`). 

Technical trick: matplotlib does not automatically map numeric values to colors for `edgecolors` in `PatchCollection` so I had to map them manually with a `ScalarMappable`. The `vmin`, `vmax`; and `cmap` used for the mapping are the same as those from `edge_fc`.


xgi/drawing/draw_utils.py
```diff
@@ -78,7 +78,7 @@ def _parse_color_arg(colors, ids, id_kind="edges"):
 
     This function is needed to handle the input formats not naturally
     handled by matploltib's Collections: IDStat, dict, and arrays of
-    floats. All those are converted to arrays of floats and.
+    floats. All those numerical foramtsare converted to arrays of floats.
```
nwlandry, line 81: typo

xgi/drawing/draw_utils.py
```diff
@@ -103,8 +103,8 @@ def _parse_color_arg(colors, ids, id_kind="edges"):
     --------
     colors : single color or ndarray
         Processed color values for plotting.
-    colors_are_mapped : bool
-        True if the colors are mapped and need special handling. This
+    colors_to_map : bool
```
nwlandry, line 106: can this just be called `map_colors`?

nwlandry (approved):
Two minor comments, but good to merge when you're ready. If you merge today, I can release 0.8.9 in time for your presentation.

xgi/drawing/draw_utils.py
maximelucas: Yea that would work too I guess, but the colors are only mapped later in the draw function

PR #581, opened 2024-09-06
fix #580 + related tests

### LINKED ISSUE #580: `remove_simplex_ids()` raises error when removing all simplices from a SC
Originally noted in #578. I am working on fixing it.

```
>>> tetra = xgi.SimplicialComplex([[1,2,3,4]])
>>> bunch = tetra.edges.filterby("order", 0, "gt")
>>> tetra.remove_simplex_ids_from(bunch)
  File "/Users/thomasrobiglio/GitHub/xgi/xgi/utils/utilities.py", line 40, in __getitem__
    return dict.__getitem__(self, item)
KeyError: 4

The above exception was the direct cause of the following exception:

  File "/Users/thomasrobiglio/GitHub/xgi/xgi/core/simplicialcomplex.py", line 749, in remove_simplex_id
    supfaces_ids = self._supfaces_id(self._edge[id])
  File "/Users/thomasrobiglio/GitHub/xgi/xgi/utils/utilities.py", line 42, in __getitem__
    raise IDNotFound(f"ID {item} not found") from e
xgi.exception.IDNotFound: 'ID 4 not found'

The above exception was the direct cause of the following exception:

  File "<stdin>", line 1, in <module>
  File "/Users/thomasrobiglio/GitHub/xgi/xgi/core/simplicialcomplex.py", line 779, in remove_simplex_ids_from
    self.remove_simplex_id(id)
  File "/Users/thomasrobiglio/GitHub/xgi/xgi/core/simplicialcomplex.py", line 757, in remove_simplex_id
    raise XGIError(f"Simplex {id} is not in the Simplicialcomplex") from e
xgi.exception.XGIError: Simplex 4 is not in the Simplicialcomplex
```

quoting @maximelucas:

> I think it fails when trying to remove edge 4: frozenset({1, 2, 3}) which does not exist anymore already because it already removed 2: frozenset({1, 2}), and in doing so had to remove 4: frozenset({1, 2, 3}) already to ensure downward closure.

thomasrobiglio:


maximelucas:
Thanks Thomas!

I think I would move the logic of this to `remove_simplex_ids_from()` instead. Because in general if a user wants to use `remove_simplex_id()`, this `removed_ids()` option is a bit surprising. And it's really a problem that only internal to `remove_simplex_ids_from()`  because we remove several at one and it needs to be self-consistent. So maybe just similarly keep track of removed ids there, and then decide when to call `remove_simplex_id` based on that.

What do you think?

tests/core/test_simplicialcomplex.py
```diff
@@ -462,7 +462,7 @@ def test_remove_simplex_id(edgelist6):
     assert set(S.edges.members()) == set(edges)
 
 
-def test_remove_simplex_ids_from(edgelist6):
+def test_remove_simplex_ids_from(edgelist6, edgelist2):
```
maximelucas, line 465:
```suggestion
def test_remove_simplex_ids_from(edgelist6, edgelist4):
```

`edgelist2` is actually not failing before the PR, `edgelist4` is

tests/core/test_simplicialcomplex.py
```diff
@@ -480,6 +480,12 @@ def test_remove_simplex_ids_from(edgelist6):
     ]
     assert set(S.edges.members()) == set(edges)
 
+    # test issue 580
+    S1 = xgi.SimplicialComplex(edgelist2)
```
maximelucas, line 484:
```suggestion
    S1 = xgi.SimplicialComplex(edgelist4)
```

xgi/core/simplicialcomplex.py
```diff
@@ -745,12 +745,12 @@ def remove_simplex_id(self, id):
         remove_edges_from : remove a collection of edges
         """
         try:
-            # remove all simplices that contain simplex
+            # Remove all simplices that contain the given simplex
             supfaces_ids = self._supfaces_id(self._edge[id])
             for sup_id in supfaces_ids:
-                self._remove_simplex_id(sup_id)
+                self.remove_simplex_id(sup_id)
```
maximelucas, line 751: I don't this was wrong before so I wouldn't touch it, unless you think it's more efficient?

thomasrobiglio: good catch, thx

maximelucas (approved):

PR #582, opened 2024-09-06
Website edits

### LINKED ISSUE #341: Fix `See Also` links
Some of the functions listed in the `See Also` field in the docstrings aren't functioning links when they exist in another module. Perhaps a way to fix this: https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html

nwlandry:
This PR
* Fixes the color switching issues
* Removes unnecessary files
* Fixes sphinx errors from the migration to the PyData theme
* Fixes #341


xgi/convert/incidence.py
```diff
@@ -37,7 +37,7 @@ def from_incidence_matrix(d, create_using=None, nodelabels=None, edgelabels=None
 
     See Also
     --------
-    incidence_matrix
+    ~xgi.linalg.hypergraph_matrix.incidence_matrix
     to_incidence_matrix
```
maximelucas, line 41: how come the first line needs the long reference and the second one does not?

nwlandry: You need the long reference if it's not in the same file.

docs/source/user_guides.rst
```diff
@@ -55,13 +51,15 @@ User Guides
     	To know everything there is to know about a specific feature of XGI 
     	+++
 
-        .. button-ref:: indepth_tutorials
+        .. button-ref:: api/tutorials/in_depth
```
maximelucas, line 54: Now you're using the path to the rst file rather than the anchor right? 
Is it better practice or what's idea behind it?

docs/source/installing_more.rst
```diff
@@ -1,8 +1,8 @@
-.. _installing_more:
+:orphan:
```
maximelucas, line 1: what does :orphan: does? better than having nothing?

docs/source/installing.rst
```diff
@@ -1,8 +1,6 @@
-.. _installing:
-
-***************
-Installing
-***************
+**************
+Installing XGI
```
maximelucas, line 2: I feel like adding "XGI" makes it longer and the menu doesn't have so much space, so I think I wouldn't add it

maximelucas:
Nice work Nich! Just a few questions. I like the color switch and that you added the case studies back in.

docs/source/user_guides.rst
nwlandry: Hard to know what best practices are. But, FWIW, matplotlib does it with relative paths: https://github.com/matplotlib/matplotlib/blob/main/doc%2Findex.rst

docs/source/installing_more.rst
nwlandry: :orphan: suppresses the "document is not part of any toctree" in sphinx. This is the only solution I found to that problem.

docs/source/installing.rst
nwlandry: Sounds good. I'll change it back.

nwlandry: Fixed.

docs/source/user_guides.rst
maximelucas: Probably easier to track back the file it looks to this way

nwlandry:
> Nice work Nich! Just a few questions. I like the color switch and that you added the case studies back in. 

Thanks so much for the review! I believe I addressed all your comments.

maximelucas (approved):

Issue #585, opened 2024-09-07
add dark version of logo to online docs


maximelucas:
It looks like this:

![image](https://github.com/user-attachments/assets/8bc37f61-9b4f-4688-bfdb-3a91193fdd85)

Maybe have both versions with transparent background, and the dark theme one with "GI" in white?

PR #586, opened 2024-09-08
Add version switcher, favicon, and context-dependent logo

### LINKED ISSUE #584: add version switcher to online docs
We used to be able to switch between the "stable" and "latest" version of the online docs. 
Now we can only do it by changing the url manually. 

It would be nice to add the switcher back, following the [explanations](https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/version-dropdown.html) from PyData theme.

nwlandry:
Fixes #584 and #585.

This PR
* Adds a favicon
* Creates a dark and light version of the logo
* Adds a version switcher


maximelucas:
Amazing Nich!! Favicon is a sweet plus.

Two things: 
1. On my laptop the menu is now looking a bit strange because on two lines
2. the light theme logo doesn't have a margin anymore so it's stuck to the "new release" button (dark theme logo doesn't have that problem)

image

I'm not fully sure how to solve 1., I think we need to find a way to control the width of the[ 3 navbars/zones (left, middle, right)](https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/layout.html)? Can the version switcher be made less wide by changing the default text from "choose version" to "vesion" or simply "stable" by default?

nwlandry:
In response to the review I
* reverted the logo back
* adjusted the spacing in the navbar
* Added a call to the Github API to get the date of the latest release

I believe I addressed all your comments, @maximelucas!

docs/source/conf.py
```diff
@@ -115,9 +118,18 @@
 
 # There are two options for replacing |today|: either, you set today to some
 # non-false value, then it is used:
-# today = ''
+r = requests.get("https://api.github.com/repos/xgi-org/xgi/releases/latest")
+
+if r.ok:
+    release_date = (
+        datetime.fromisoformat(r.json()["published_at"]).date().strftime("%B %d, %Y")
```
maximelucas, line 125:
```suggestion
        datetime.fromisoformat(r.json()["published_at"]).date().strftime("%b %d, %Y")
```

docs/source/conf.py
```diff
@@ -184,6 +201,13 @@
             "icon": "fa-brands fa-mastodon",  # Font Awesome icon
         },
     ],
+    "header_links_before_dropdown": 4,
```
maximelucas, line 204: Anyway of having 5 and keeping the Gallery visible (not in dropdown)?

docs/source/conf.py
maximelucas: just short month version for space

maximelucas:
You're on fire, the automatic date retrieval is great!!
Two tiny comments but I'll let you choose, approving already.

maximelucas (approved):

docs/source/conf.py
nwlandry: The issue was that I wanted everything on one line even when the user wasn't in full-screen, which is basically impossible unless the navbar is slimmed down. So I will keep it like it is for now, and then change it later if need be.

nwlandry:
Thanks for the review!

PR #587, opened 2024-09-08
Added simpliciality measures

nwlandry:
This PR adds all the functionality from this [repo](https://github.com/nwlandry/the-simpliciality-of-higher-order-networks).

### DIFF
--- a/xgi/algorithms/simpliciality.py
+++ b/xgi/algorithms/simpliciality.py
@@ -241,7 +241,7 @@ def mean_face_edit_distance(H, min_size=2, exclude_min_size=True, normalize=True
         if len(e) >= min_size:
             s = _count_subfaces(t, e, min_size=min_size)
             m = _max_number_of_subfaces(min_size, len(e))
-            d = m - s # missing subfaces
+            d = m - s  # missing subfaces
             if normalize and m != 0:
                 d *= 1.0 / m
             avg_d += d / len(max_faces)
--- a/xgi/algorithms/simpliciality.py
+++ b/xgi/algorithms/simpliciality.py
@@ -36,6 +36,12 @@ def edit_simpliciality(H, min_size=2, exclude_min_size=True):
     -------
     float
         The edit simpliciality
+
+    References
+    ----------
+    "The simpliciality of higher-order order networks"
+    by Nicholas Landry, Jean-Gabriel Young, and Nicole Eikmeier,
+    *EPJ Data Science* **13**, 17 (2024).
     """
     edges = H.edges.filterby("size", min_size, "geq").members()
 
@@ -100,6 +106,12 @@ def simplicial_edit_distance(H, min_size=2, exclude_min_size=True, normalize=Tru
     -------
     float
         The edit simpliciality
+
+    References
+    ----------
+    "The simpliciality of higher-order order networks"
+    by Nicholas Landry, Jean-Gabriel Young, and Nicole Eikmeier,
+    *EPJ Data Science* **13**, 17 (2024).
     """
     edges = H.edges.filterby("size", min_size, "geq").members()
 
@@ -165,6 +177,12 @@ def face_edit_simpliciality(H, min_size=2, exclude_min_size=True):
     -------
     float
         The face edit simpliciality
+
+    References
+    ----------
+    "The simpliciality of higher-order order networks"
+    by Nicholas Landry, Jean-Gabriel Young, and Nicole Eikmeier,
+    *EPJ Data Science* **13**, 17 (2024).
     """
     edges = (
         H.edges.maximal().filterby("size", min_size + exclude_min_size, "geq").members()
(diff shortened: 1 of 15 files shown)


xgi/algorithms/simpliciality.py
```diff
@@ -0,0 +1,434 @@
...
+    sub_edges = list(_powerset(face, min_size=min_size, max_size=len(face) - 1))
+    count = 0
+    for e in sub_edges:
+        if t.search(e):
+            count += 1
+
+    return count
+
+
+def _count_subfaces(t, face, min_size=1):
```
tlarock, line 346: Is this a repeat of the above function with the same signature?

xgi/algorithms/simpliciality.py
```diff
@@ -0,0 +1,434 @@
...
+                    # faces so no inclusions.
+                    if not t.search(c):
+                        redundant_missing_faces.add(frozenset(c))
+
+        nm = _max_number_of_subfaces(min_size, len(e))
+        nf = _count_subfaces(t, e, min_size)
+        rmf = len(redundant_missing_faces)
+        ms += nm - nf - rmf
+
+    try:
```
tlarock, line 79: Do we have a style preference for a try/catch block here rather than checking for 0 denominator in a normal conditional? Since `s` and `ms` should both be integers, could we just check `if (ms+s) > 0` rather than relying on exception handling?

tlarock: I think this comment applies to the returns of all of these functions. Practically the outcome should be the same using either approach, so in my view it is a style/readability question. My personal preference is for a conditional, but open to leaving it as is if it is consistent with style elsewhere or desireable for another reason.

tlarock (changes requested):
Looks good overall and tests appear to pass. Worth considering how we structure the return of the functions, since using exception handling rather than conditionals is a design choice. One function appears to be duplicated. Otherwise good to go!

nwlandry:
Thanks for the review, @tlarock! Someone pointed out a minor error in my simpliciality paper, so I'm determining what to do with that first before making your suggested changes.

xgi/algorithms/simpliciality.py
nwlandry: Great point. I fixed this according to your suggestion.

xgi/algorithms/simpliciality.py
nwlandry: Oops --- great catch. Fixed now!

nwlandry:
Hi @tlarock --- I believe I addresed all of your comments. The URL checker fails, but this was addressed in another PR so it will go away once we merge into the main branch.

xgi/algorithms/simpliciality.py
```diff
@@ -0,0 +1,369 @@
...
+    Parameters
+    ----------
+    H : xgi.Hypergraph
+        The hypergraph of interest
+    min_size: int, default: 2
+        The minimum hyperedge size to include when
+        calculating whether a hyperedge is a simplex
+        by counting subfaces.
+    exclude_min_size : bool, optional
+        Whether to include minimal simplices when counting simplices, by default True
```
tlarock, line 70: The name is `exclude_min_size`, but the documentation reads "Whether to _include_...default `True`", which makes my head spin a bit! Maybe reword for clarity?

xgi/algorithms/simpliciality.py
```diff
@@ -0,0 +1,369 @@
...
+    "The simpliciality of higher-order order networks"
+    by Nicholas Landry, Jean-Gabriel Young, and Nicole Eikmeier,
+    *EPJ Data Science* **13**, 17 (2024).
+    """
+    return 1 - simplicial_edit_distance(
+        H, min_size=min_size, exclude_min_size=exclude_min_size
+    )
+
+
+def simplicial_edit_distance(H, min_size=2, exclude_min_size=True, normalize=True):
```
tlarock, line 55: I don't think I could fully understand from the documentation alone what exactly the parameters `min_size` and `exclude_min_size` control. I had to go read the code, especially for the latter parameter. I'm not sure if there is a straightforward solution without either (a) having longer more expressive parameter names, or (b) a more thorough description in the main docstring that explains explicitly how they change the calculation, but maybe something worth thinking about a bit since I am somewhat familiar with the measures and couldn't quite understand.

tlarock (commented):
Very close! Sorry to be a pain about the documentation, but hopefully easier to clairfy now than later :) Of course feel free to take or leave

xgi/algorithms/simpliciality.py
nwlandry: On it! Thanks for bringing this up!

xgi/algorithms/simpliciality.py
nwlandry: Good call. I will correct everywhere.

tlarock (approved):

PR #590, opened 2024-09-09
Fix version switching

nwlandry:
This PR fixes the version dropdown


docs/source/conf.py
```diff
@@ -173,12 +174,33 @@
...
+if version_match == "dev":
+    version = "dev"
+
 # documentation.
 html_theme_options = {
     "logo": {
         "image_light": "../../logo/logo.svg",
         "image_dark": "../../logo/logo_white.svg",
     },
+    "navbar_start": ["navbar-logo", "navbar-version"],
```
maximelucas, line 203: is having the version on the left with the logo not redundant with having it displayed on the right in the switcher?

maximelucas:
Not sure exactly what caused this but now the New release button says "dev" instead of the version number:

image

nwlandry:
Mmm. Good catch! Sorry about that. The only issue with putting the version in the version switcher is that you need to edit by hand, which is totally fine, but I was trying to be clever and have the website be completely "hands-free" so it would auto import the version, date, etc. but obviously there are some downsides. Happy to make a little PR removing the logic I added.

PR #596, opened 2024-09-16
Add ability to generate ER hypergraphs without multiedges

### LINKED ISSUE #339: uniform_erdos_renyi_hypergraph produces way too many edges
Unless I'm seriously mistaken as to how this is supposed to work, the number of edges produced by `uniform_erdos_renyi_hypergraph` is off by a factor of $m!$.

```python
import xgi

p = 1e-6
n = 1000
m = 3
plt.hist([len(xgi.uniform_erdos_renyi_hypergraph(n, m, p, p_type="prob").edges) for i in range(1000)],
    label="Num. of edges")
plt.axvline(math.comb(n, m)*p, label="Exp. number of edges")
plt.legend()
```
download

Possible solutions: lower `q`  by $m!$ or even better, use [nth combination](https://docs.python.org/3/library/itertools.html#itertools-recipes:~:text=def%20nth_combination) instead of n-th product when defining `_index_to_edge`.

nwlandry:
This PR fixes #339.


nwlandry:
@arashbm --- does this fix the issue you raised?

nwlandry:
@maximelucas --- are you able to review this?

xgi/generators/uniform.py
```diff
@@ -346,6 +369,47 @@ def _index_to_edge(index, n, m):
...
+        while r - comb(n - 1 - cs, m - s, exact=True) > 0:
+            r -= comb(n - 1 - cs, m - s, exact=True)
+            cs += 1
+        c.append(cs)
+        j = cs
+    return c
+
+
 def _index_to_edge_partition(index, partition_sizes, m):
     """Generate a hyperedge given an index in the list of possible edges
```
maximelucas, line 414: given those things, how is it generated?

xgi/generators/uniform.py
```diff
@@ -346,6 +369,47 @@ def _index_to_edge(index, n, m):
     return [(index // (n**r) % n) for r in range(m - 1, -1, -1)]
 
 
+def _index_to_edge_comb(index, n, m):
+    """Generate a hyperedge given an index in the list of possible edges.
```
maximelucas, line 373: given the index, how is the hyperedge generated?

xgi/generators/uniform.py
```diff
@@ -346,6 +369,47 @@ def _index_to_edge(index, n, m):
...
+        The index of the hyperedge in the list of all possible hyperedges.
+    n : int > 0
+        The number of nodes
+    m : int > 0
+        The hyperedge size.
+
+    Returns
+    -------
+    list
+        The reconstructed hyperedge
```
maximelucas, line 390: Would there be a better word than "reconstructed"? I'm not sure exactly what is meant

xgi/generators/uniform.py
```diff
@@ -277,7 +278,16 @@ def uniform_erdos_renyi_hypergraph(n, m, p, p_type="degree", seed=None):
         Mean expected degree if p_type="degree" and
```
maximelucas, line 278: Maybe I'd but the case "prob" first as it's the most classic one, and maybe add "(Default)" next to it too.

xgi/generators/uniform.py
```diff
@@ -277,7 +278,16 @@ def uniform_erdos_renyi_hypergraph(n, m, p, p_type="degree", seed=None):
...
         probability of an m-hyperedge if p_type="prob"
     p_type : str
-        "degree" or "prob", by default "degree"
+        "degree" or "prob", by default "prob"
```
maximelucas, line 281: add something like "changes the way `p` is interpreted (see `p` for detail)

btw this option is not available in the other random_hypergraph() functions, should we make it?

xgi/generators/uniform.py
```diff
@@ -277,7 +278,16 @@ def uniform_erdos_renyi_hypergraph(n, m, p, p_type="degree", seed=None):
...
+    multiedges : bool, optional
+        Whether or not to allow multiedges. If True, there
+        can be significant speedups but at the cost of creating
+        (potentially unwanted artifacts). When multiedges=True,
```
maximelucas, line 285:
```suggestion
        (potentially unwanted) artifacts. When multiedges=True,
```

xgi/generators/uniform.py
```diff
@@ -299,28 +308,42 @@ def uniform_erdos_renyi_hypergraph(n, m, p, p_type="degree", seed=None):
     H.add_nodes_from(node_labels)
 
     if p_type == "degree":
-        q = p / (m * n ** (m - 1))  # wiring probability
+        if multiedges:
+            q = p / (m * n ** (m - 1))  # wiring probability
+        else:
+            q = p * n / (m * comb(n, m))
```
maximelucas, line 314: add a ref if there is one?

xgi/generators/uniform.py
```diff
@@ -260,7 +261,7 @@ def uniform_HPPM(n, m, k, epsilon, rho=0.5, seed=None):
     return uniform_HSBM(n, m, p, sizes, seed=seed)
 
 
-def uniform_erdos_renyi_hypergraph(n, m, p, p_type="degree", seed=None):
+def uniform_erdos_renyi_hypergraph(n, m, p, p_type="prob", multiedges=False, seed=None):
```
maximelucas, line 264: also, this function uses an optimization, maybe add a comment about it and a ref

xgi/generators/uniform.py
```diff
@@ -299,28 +308,42 @@ def uniform_erdos_renyi_hypergraph(n, m, p, p_type="degree", seed=None):
...
-        raise XGIError("Probability not in [0,1].")
+        raise XGIError("Probability not in [0, 1].")
+
+    if multiedges:
+        max_index = n**m
+        index = np.random.geometric(q) - 1  # -1 b/c zero indexing
+        f = _index_to_edge_prod
+    else:
+        max_index = comb(n, m, exact=True)
+        index = np.random.geometric(q)
```
maximelucas, line 329: why not -1 here like in multiedge case?

maximelucas:
Mainly a few comments about clarifying the code with comments!

xgi/generators/uniform.py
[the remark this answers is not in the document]
nwlandry: Fixed!

xgi/generators/uniform.py
nwlandry: Done.

xgi/generators/uniform.py
nwlandry: I fixed it so they're both zero-indexed.

xgi/generators/uniform.py
nwlandry: Fixed.

xgi/generators/uniform.py
nwlandry: Fixed.

xgi/generators/uniform.py
```diff
@@ -299,28 +308,42 @@ def uniform_erdos_renyi_hypergraph(n, m, p, p_type="degree", seed=None):
...
+        f = _index_to_edge_comb
 
-    index = np.random.geometric(q) - 1  # -1 b/c zero indexing
-    max_index = n**m
-    while index < max_index:
-        e = set(_index_to_edge(index, n, m))
+    while index <= max_index:
+        e = set(f(index, n, m))
         if len(e) == m:
```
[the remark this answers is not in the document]
nwlandry: fixed by explaining more.

xgi/generators/uniform.py
nwlandry: Fixed.

xgi/generators/uniform.py
nwlandry: Fixed.

xgi/generators/uniform.py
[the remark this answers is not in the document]
nwlandry: Fixed.

nwlandry: Fixed.

nwlandry: Done.

nwlandry:
@maximelucas --- I believe I addressed all of your comments!

xgi/generators/uniform.py
```diff
@@ -260,24 +299,40 @@ def uniform_HPPM(n, m, k, epsilon, rho=0.5, seed=None):
...
+    :math:`\mathcal{O}(m(N + |E|))`. See the references
+    for more details.
+
     Parameters
     ----------
     n : int > 0
         Number of nodes
     m : int > 0
         Hyperedge size
     p : float or int > 0
```
maximelucas, line 321:
```suggestion
    p : float or int >= 0
```

xgi/generators/uniform.py
```diff
@@ -260,24 +299,40 @@ def uniform_HPPM(n, m, k, epsilon, rho=0.5, seed=None):
     return uniform_HSBM(n, m, p, sizes, seed=seed)
 
 
-def uniform_erdos_renyi_hypergraph(n, m, p, p_type="degree", seed=None):
+def uniform_erdos_renyi_hypergraph(n, m, p, p_type="prob", multiedges=False, seed=None):
     """Generate an m-uniform Erdős–Rényi hypergraph
 
     This creates a hypergraph with `n` nodes where
     hyperedges of size `m` are created at random to
     obtain a mean degree of `k`.
```
maximelucas, line 307:
```suggestion
    hyperedges of size `m` are created at random 
    with probability (or to obtain a mean degree of) `p`.
```

xgi/generators/uniform.py
```diff
@@ -260,24 +299,40 @@ def uniform_HPPM(n, m, k, epsilon, rho=0.5, seed=None):
...
     n : int > 0
         Number of nodes
     m : int > 0
         Hyperedge size
     p : float or int > 0
-        Mean expected degree if p_type="degree" and
-        probability of an m-hyperedge if p_type="prob"
-    p_type : str
-        "degree" or "prob", by default "degree"
+        probability of an m-hyperedge if p_type="prob" and
```
maximelucas, line 322:
```suggestion
        Probability of an m-hyperedge if p_type="prob" and
```

xgi/generators/uniform.py
```diff
@@ -260,24 +299,40 @@ def uniform_HPPM(n, m, k, epsilon, rho=0.5, seed=None):
...
+        mean expected degree if p_type="degree"
+    p_type : str, optional
+        changes the way p is interpreted (see p for detail).
```
maximelucas, line 325:
```suggestion
        Determines the way p is interpreted (see p for detail).
```

xgi/generators/uniform.py
```diff
@@ -286,10 +341,26 @@ def uniform_erdos_renyi_hypergraph(n, m, p, p_type="degree", seed=None):
...
         The Erdos Renyi hypergraph
 
-
     See Also
     --------
     ~xgi.generators.random.random_hypergraph
+
+    Notes
+    -----
+    Because XGI only stores edges as sets, if self-loops are allowed,
```
maximelucas, line 350: But there's no option to allow (or not) self-loops is this function, right? I'm a bit confused. In any case, since that singleton {0} is always removed by the method, does it matter?

The multiedges option is not linked to that correct?

xgi/generators/uniform.py
```diff
@@ -286,10 +341,26 @@ def uniform_erdos_renyi_hypergraph(n, m, p, p_type="degree", seed=None):
...
+    for example, the edge (0, 0, 0) will be mapped to {0}. However,
+    because this is explicitly a *uniform* method, we discard these edges
+    so that this is the case. For sparse networks, this is a rare
```
maximelucas, line 353:
```suggestion
    because this is explicitly a *uniform* method, we discard these edges. 
    For sparse networks, this is a rare
```
if "so that this is the case" meant "so that the output is a uniform hypergraph" then I think we can remove that bit.

xgi/generators/uniform.py
```diff
@@ -286,10 +341,26 @@ def uniform_erdos_renyi_hypergraph(n, m, p, p_type="degree", seed=None):
...
+    occurrence and this method offers an order of magnitude speedup,
+    so while it is not the default behavior, this option is exposed to
```
maximelucas, line 355: The option of doing what? Sorry I don't understand how the multiedge option is related to what's above in this Note 🥲 I'll try so maybe you'll understand what I don't understand hehe:

For sparse networks, we need to discard fewer edges that are self-loops (result in in edges of size < m), simply because we produce few edges overall. Because of that, this method (meaning, the way of generating edges with the geometric series, but only with multiedge=True?) is faster (compared to the naive other random hypergraph function or compared to multiedge=False?). So, the we give the option to set multiedge=True, even though the default is multiedge=False. 

Are we saying multiedge=True, is faster, and in sparse networks we will need to discard fewer self-loops (in absolute numbers, not relatively to the total number of edges though) so that it makes proportionally even faster in that case?

xgi/generators/uniform.py
```diff
@@ -299,27 +370,69 @@ def uniform_erdos_renyi_hypergraph(n, m, p, p_type="degree", seed=None):
...
+        # if f corresponds to _index_to_edge_prod,
+        # edge ids are not guaranteed to be unique
+        # and when casting to a set, they will no
+        # longer be of size m.
+        # for instance (0, 0, 0) becomes {0}
+        # if we accept these edges, the hypergraph
+        # will not longer be uniform, so we discard them.
         if len(e) == m:
             H.add_edge(e)
         index += np.random.geometric(q)
```
maximelucas, line 411: just checking, here we don't need the -1 like above because it's not the initial number right?

xgi/generators/uniform.py
```diff
@@ -299,27 +370,69 @@ def uniform_erdos_renyi_hypergraph(n, m, p, p_type="degree", seed=None):
...
+    starting at zero. For example, (0, 0, 0) has index 0, (0, 0, 1) has index 1,
+    (0, 0, 2) has index 2, (0, 0, 3) has index 3, (0, 1, 0) has index 4, and so on.
+    This function will, for instance,
+    return (0, 0, 3) for index 3, network size 4, and edge size 3.
+
+    Because XGI only stores edges as sets, the edge (0, 0, 0) will be mapped
+    to {0}. However, because this is explicitly a *uniform* method, we discard
+    these edges so that this is the case. For sparse networks, this is a rare
+    occurrence and this method offers an order of magnitude speedup, so while
+    it is not the default behavior.
```
maximelucas, line 431: end of sentence missing i think

xgi/generators/uniform.py
```diff
@@ -299,27 +370,69 @@ def uniform_erdos_renyi_hypergraph(n, m, p, p_type="degree", seed=None):
...
+
+    In this method, it treats each edge permutation as distinct, which can
```
maximelucas, line 433:
```suggestion
    This method treats each edge permutation as distinct, which can
```

xgi/generators/uniform.py
```diff
@@ -346,13 +460,23 @@ def _index_to_edge(index, n, m):
...
+    Imagine that there is a hypergraph with 4 nodes and an edge size of 3.
+    We write out each edge incrementing the last entry first, followed by the
+    second-to-last entry and so on, with each edge corresponding to an index
+    starting at zero. For example, (0, 1, 2) has index 0, (0, 1, 3) has index 0,
+    (0, 2, 3) has index 2, and (1, 2, 3) has index 3. This function will, for instance,
+    return (0, 2, 3) for index 2, network size 4, and edge size 3.
+
+
+    In this function, we prohibit multiedges, so each index corresponds to a
+    unique edge.
```
maximelucas, line 475: But in _index_to_edge_prod too right? Should it not be, "so each edge corresponds to a unique index"?
Which is not he case in _index_to_edge_prod, eg index 1 and 4 are the same edge in your example above.

maximelucas:
Thanks for the explanations! I'm still confused about some things, see above. Sorry for being a pain 😅 As long as the advantages are clear (and ideally why they are) we're good I guess. And that each method is clear from its own documentation. 

If I understand a bit better, the `index` in "artificial/internal" in the sense that is just runs over all possible edges of size m among n nodes, with/without caring for the order (and hence allowing or not multiedges).

nwlandry:
> Thanks for the explanations! I'm still confused about some things, see above. Sorry for being a pain 😅 As long as the advantages are clear (and ideally why they are) we're good I guess. And that each method is clear from its own documentation.
> 
> If I understand a bit better, the `index` in "artificial/internal" in the sense that is just runs over all possible edges of size m among n nodes, with/without caring for the order (and hence allowing or not multiedges).

Yes, that's correct! I'll take a look at your more involved comments (hopefully today!)

xgi/generators/uniform.py
nwlandry: Exactly. If the minimum was zero (i.e., we subtracted 1 from the right-hand side), you might sample the same index multiple times. I'll add a note.

xgi/generators/uniform.py
nwlandry: Fixed.

xgi/generators/uniform.py
nwlandry: fixed.

xgi/generators/uniform.py
nwlandry: Hopefully my revision addressed this! Basically what I'm saying is that (1) allowing self-loops and multiedges is much faster and (2) when the hypergraph is sparse, not many artifacts are created anyway so it's not so bad.

maximelucas (approved):

nwlandry:
Thanks for all the comments!!

PR #597, opened 2024-09-16
Fix random hypergraph

### LINKED ISSUE #361: remove np random generators for memory reasons
@maximelucas The combinatorial explosion of the mask array makes it impossible to use in some scenarios: e.g.
```pycon
>>> import xgi
>>> xgi.random_hypergraph(1000, order=5, ps=[1e-12])  # => should create a network with ~1350 edges
numpy.core._exceptions._ArrayMemoryError: Unable to allocate 9.72 PiB for an array with shape (1368173298991500,) and data type float64
```

But that also has an easy solution of not pre-allocating mask:

```python
import xgi
import numpy as np
import itertools

def random_hypergraph(N, ps, order=None, seed=None):
    rng = np.random.default_rng(seed)

    if order is not None:
        if len(ps) != 1:
            raise ValueError("ps must contain a single element if order is an int")

    if (np.any(np.array(ps) < 0)) or (np.any(np.array(ps) > 1)):
        raise ValueError("All elements of ps must be between 0 and 1 included.")

    nodes = range(N)
    hyperedges = []

    for i, p in enumerate(ps):

        if order is not None:
            d = order
        else:
            d = i + 1  # order, ps[0] is prob of edges (d=1)

        potential_edges = itertools.combinations(nodes, d + 1)
        edges_to_add = [e for e in potential_edges if rng.random() <= p]

        hyperedges += edges_to_add

    H = xgi.empty_hypergraph()
    H.add_nodes_from(nodes)
    H.add_edges_from(hyperedges)

    return H
```

This doesn't fail but this still takes $O({{N}\choose{m}})$ time to run. 

_Originally posted by @arashbm in https://github.com/xgi-org/xgi/issues/339#issuecomment-1517780776_

Summary: go back to list comprehension rather than np.random()

nwlandry:
This fixes #361.

Before I fix the tests, is this what you were picturing @maximelucas and @arashbm?


xgi/generators/random.py
```diff
@@ -76,13 +82,74 @@ def random_hypergraph(N, ps, order=None, seed=None):
...
-        mask = np.random.random(size=n_comb) <= p  # True if edge to keep
+        size = d + 1
+        if p > 0:
+            h_uniform = uniform_erdos_renyi_hypergraph(n, size, p, seed=seed)
+            H << h_uniform
+
+    return H
+
+
+def random_hypergraph(N, ps, seed=None):
```
maximelucas, line 93:
```suggestion
def random_hypergraph(n, ps, seed=None):
```

for consistency with the others

xgi/generators/random.py
```diff
@@ -76,13 +82,74 @@ def random_hypergraph(N, ps, order=None, seed=None):
...
+    """Generates a random hypergraph
+
+    Generate N nodes, and connect any d+1 nodes
+    by a hyperedge with probability ps[d-1].
+
+    Parameters
+    ----------
+    N : int
```
maximelucas, line 101:
```suggestion
    n : int
```

xgi/generators/random.py
```diff
@@ -76,13 +82,74 @@ def random_hypergraph(N, ps, order=None, seed=None):
...
+    for i, p in enumerate(ps):
+        d = i + 1  # order, ps[0] is prob of edges (d=1)
 
-        edges_to_add = [e for e, val in zip(potential_edges, mask) if val]
+        for hyperedge in combinations(nodes, d + 1):
+            if random.random() <= p:
+                hyperedges.append(hyperedge)
 
-        hyperedges += edges_to_add
+    hyperedges += [[i] for i in nodes]  # add singleton edges
```
maximelucas, line 152: I think in generative models we don't add singletons as a general rule by default, so I'd remove?

xgi/generators/random.py
```diff
@@ -20,15 +21,15 @@
 ]
 
 
-def random_hypergraph(N, ps, order=None, seed=None):
+def fast_random_hypergraph(n, ps, order=None, seed=None):
     """Generates a random hypergraph
 
     Generate N nodes, and connect any d+1 nodes
```
maximelucas, line 27: add here or as a Note something about why this one is fast what the optimization is (or just a ref)

xgi/generators/random.py
```diff
@@ -76,13 +82,74 @@ def random_hypergraph(N, ps, order=None, seed=None):
         else:
             d = i + 1  # order, ps[0] is prob of edges (d=1)
 
-        potential_edges = combinations(nodes, d + 1)
-        n_comb = comb(N, d + 1, exact=True)
-        mask = np.random.random(size=n_comb) <= p  # True if edge to keep
+        size = d + 1
+        if p > 0:
+            h_uniform = uniform_erdos_renyi_hypergraph(n, size, p, seed=seed)
+            H << h_uniform
```
maximelucas, line 88: add brief comment reminding what that symbol does for future ref

maximelucas:
Looks good, thanks Nich! Just a few minor comments

xgi/generators/random.py
nwlandry: This was the behavior of the original function, but yeah, agreed, let's axe this.

xgi/generators/random.py
nwlandry: Fixed.

xgi/generators/random.py
nwlandry: Added a reference and a note to look at the refs.

maximelucas:
This line seems to be the one failing the tests:
https://github.com/xgi-org/xgi/blob/5e8bd4ecc689983b5e69d2c12b9daae08b0ed674/xgi/generators/random.py#L164-L166

Looks unused, I think we can remove it?

nwlandry:
Sooooo I refactored the random_hypergraph method to be a bit more flexible and I will work on documenting and fixing tests.

nwlandry:
@maximelucas --- any chance you can take one last look?

xgi/generators/random.py
```diff
@@ -1,45 +1,129 @@
...
+
+    ps, order = _check_input_args(ps, order)
+
+    nodes = range(n)
+
+    H = empty_hypergraph()
+    H.add_nodes_from(nodes)
+
+    for d, p in zip(order, ps):
+        if p == 1.0:
```
maximelucas, line 87:
```suggestion
        if p == 1:
```
don't need the .0 anymore i think

xgi/generators/random.py
```diff
@@ -1,45 +1,129 @@
...
 
     Parameters
     ----------
-    N : int
+    n : int
         Number of nodes
-    ps : list of float
+    ps : list of float, or float
         List of probabilities (between 0 and 1) to create a
         hyperedge at each order d between any d+1 nodes. For example,
```
maximelucas, line 117:
```suggestion
        hyperedge at each order d between any d+1 nodes (when `order` is None). For example,
```

xgi/generators/random.py
```diff
@@ -1,45 +1,129 @@
...
         ps[0] is the wiring probability of any edge (2 nodes), ps[1]
-        of any triangles (3 nodes).
-    order: int of None (default)
-        If None, ignore. If int, generates a uniform hypergraph with edges
-        of order `order` (ps must have only one element).
-    seed : integer or None (default)
-            Seed for the random number generator.
+        of any triangles (3 nodes). If a float, generate a uniform hypergraph
+        (in this case, order must be specified)
```
maximelucas, line 120:
```suggestion
        (in this case, order must be specified)
        See `order` for advanced options when it is not `None`.
```

maximelucas:
Just a few super minor things for the docs, otherwise approved already, thanks!

PR #600, opened 2024-10-17
Add tensor centralities

### LINKED ISSUE #505: Add tensor functionality to eigenvector centralities
https://github.com/pnnl/GENTTSV/tree/master/xgi and add license.

nwlandry:
Added tensor methods to the centrality module. Fixes #505.

This PR also fixes Issue #616, where the `node_edge_centrality` was not converging correctly.


nwlandry:
@ilyaamburg --- I rewrote the `pairwise_incidence` method and I have a few questions on this: (1) does order matter when adding (i, j) to the dictionary of pairs? (2) there was an "r" parameter for this function, but because `r` seemed to just be the maximum edge size, the loop where you added the self-loops (node, node) seems to always be true. So I removed `r` as an argument, and removed the if-statement. Let me know your thoughts!

xgi/utils/tensor.py
```diff
@@ -0,0 +1,191 @@
...
+        for i in range(1, r - l + 2):
+            _coefs.append(_coefs[-1] * a[u] / i)
+        _coefs = np.array(_coefs)
+        _coefs[0] = 0
+        coefs = convolve(coefs, _coefs)[0:r]
+    gen_fun_coef = coefs[-1]
+    return gen_fun_coef
+
+
+def ttsv1(E, H, r, a):
```
maximelucas, line 105: missing docs. also the name is not very explicit, anyway of finding another one that's not too long?

xgi/utils/tensor.py
```diff
@@ -0,0 +1,191 @@
...
+            else:
+                gen_fun_coef = get_gen_coef_fft_fast_array(
+                    edge_without_node, a, node, l, r
+                )
+            c += r_minus_1_factorial * l * gen_fun_coef / alpha
+        s[node] = c
+    return s
+
+
+def ttsv2(E, H, r, a, n):
```
maximelucas, line 130: same as ttsv1

xgi/utils/tensor.py
```diff
@@ -0,0 +1,191 @@
...
+
+    Returns
+    -------
+    float
+        the Banerjee coefficient
+    """
+    return sum(((-1) ** j) * binomial(l, j) * (l - j) ** r for j in range(l + 1))
+
+
+def get_gen_coef_subset_expansion(edge_values, node_value, r):
```
maximelucas, line 59: missing docs also a few functions after this one

xgi/utils/tensor.py
```diff
@@ -0,0 +1,191 @@
...
+    for e, edge in edgedict.items():
+        for i, j in combinations(sorted(edge), 2):
+            pairs[(i, j)].add(e)
+        for n in edge:
+            pairs[(n, n)].add(e)
+    return pairs
+
+
+def banerjee_coeff(l, r):
+    """Return the Banerjee alpha coefficient
```
maximelucas, line 42: is this a standard coefficient? googling didn't help me. Either change the name, or say what it relates to, and maybe add a ref?

xgi/utils/tensor.py
```diff
@@ -0,0 +1,191 @@
...
+
+    Parameters
+    ----------
+    l : int
+        size of given hyperedge
+    r : int
```
maximelucas, line 48: should we replace these by "s" or "size" and "s_max" to make it more explicit and consistent with the rest of the library?

xgi/utils/tensor.py
```diff
@@ -0,0 +1,191 @@
...
+
+__all__ = [
+    "pairwise_incidence",
+    "ttsv1",
+    "ttsv2",
+]
+
+
+def pairwise_incidence(edgedict):
+    """Create pairwise adjacency dictionary from hyperedge list dictionary
```
maximelucas, line 20: Function name says "incidence" but description says "adjacency", maybe pick one. 

Also, this feel like it could/should be a general function? If I got it right, we have `xgi.adjacency_matrix(H, order=1)` that associates a pair of nodes to the number of edges of which it's part, and here we associated it to the list of edge IDs, with a dict. Would it make sense to have a new one like `xgi.adjacency_dict(H, order)` somewhere (not in utils/tensor.py)?

xgi/algorithms/centrality.py
```diff
@@ -392,3 +387,161 @@ def katz_centrality(H, cutoff=100):
         c *= 1 / norm(c, 1)
     nodedict = dict(zip(range(n), H.nodes))
     return {nodedict[idx]: c[idx] for idx in nodedict}
+
+
+def h_eigenvector_tensor_centrality(H, max_iter=100, tol=1e-6):
+    """Compute the H-eigenvector centrality of a hypergraph.
```
maximelucas, line 393: maybe add one line of additional info saying for example what H stands for, or what the specificity of this centrality is?

xgi/algorithms/centrality.py
```diff
@@ -392,3 +387,161 @@ def katz_centrality(H, cutoff=100):
...
+
+    Returns
+    -------
+    dict
+        Centrality, where keys are node IDs and values are centralities. The
+        centralities are 1-normalized.
+
+    See Also
+    --------
+    clique_eigenvector_centrality
```
maximelucas, line 413:
```suggestion
    clique_eigenvector_centrality
    z_eigenvector_tensor_centrality
```

xgi/algorithms/centrality.py
```diff
@@ -392,3 +387,161 @@ def katz_centrality(H, cutoff=100):
...
+        centralities are 1-normalized.
+
+    Raises
+    ------
+    XGIError
+        If the hypergraph is not uniform.
+
+    See Also
+    --------
+    clique_eigenvector_centrality
```
maximelucas, line 491:
```suggestion
    clique_eigenvector_centrality
    h_eigenvector_tensor_centrality
```

xgi/algorithms/centrality.py
```diff
@@ -392,3 +387,161 @@ def katz_centrality(H, cutoff=100):
...
+
+    def LR_evec(A):
+        """Compute the largest real eigenvalue of the matrix A"""
+        _, v = eigsh(A, k=1, which="LM", tol=1e-5, maxiter=200)
+        evec = np.array([_v for _v in v[:, 0]])
+        if evec[0] < 0:
+            evec = -evec
+        return evec / norm(evec, 1)
+
+    def f(u):
```
maximelucas, line 526: what does this do?

maximelucas:
Thanks Nich. A few comments mainly about documentation. Let's see what @ilyaamburg has to say about your question too.

xgi/algorithms/centrality.py
ilyaamburg: This makes the eigenvector positive if you happen to compute a negative one and makes it stochastic.

xgi/algorithms/centrality.py
ilyaamburg: I'm thinking of adding a reference for H- (and Z-) eigenvectors of tensors.

xgi/utils/tensor.py
ilyaamburg: You're right, we should stick with one convention, and it should be incidence. Yeah, it might make sense to have to have an xgi.incidence_dict(H, order) function.

xgi/utils/tensor.py
ilyaamburg: I'm down with "size"!

xgi/utils/tensor.py
ilyaamburg: We should add reference to the Banerjee paper https://www.sciencedirect.com/science/article/pii/S0024379516306115?via%3Dihub

xgi/utils/tensor.py
ilyaamburg: I will add these.

xgi/utils/tensor.py
ilyaamburg: Yes.

xgi/utils/tensor.py
ilyaamburg: I will add docs. It is "tensor times same vector in all modes but one" -- ttsv is standard in the tensor literature/software world, so maybe we're ok?

ilyaamburg:
> @ilyaamburg --- I rewrote the `pairwise_incidence` method and I have a few questions on this: (1) does order matter when adding (i, j) to the dictionary of pairs? (2) there was an "r" parameter for this function, but because `r` seemed to just be the maximum edge size, the loop where you added the self-loops (node, node) seems to always be true. So I removed `r` as an argument, and removed the if-statement. Let me know your thoughts!

Hey Nick, I am liking the new pairwise_incidence that uses combinations and defaultdict with set values, which should be fine, since the order does not matter as long as you end up one of each pair, and the order of the values does not matter either. The r and corresponding if statement should stay, though, because we only want self loops in the case l (= |e|) < r; that is, we do not want to add them if the hyperedge size is maximal.

xgi/utils/tensor.py
nwlandry: Fixed.

xgi/utils/tensor.py
nwlandry: Reference added.

tests/algorithms/test_centrality.py
```diff
@@ -195,3 +167,125 @@ def test_katz_centrality(edgelist1, edgelist8):
...
+        1: 0.45497398635982933,
+        2: 0.45900452108663403,
+        3: 0.04301074627676834,
+        4: 0.04301074627676829,
+    }
+    for i in c:
+        assert np.allclose(c[i], true_c[i])
+
+
+def _ratio(r, m, kind="CEC"):
```
maximelucas, line 262: should this be in tests or in some utility file?

xgi/algorithms/centrality.py
nwlandry: Done.

tests/algorithms/test_centrality.py
nwlandry: Maybe? It really is only used to test centralities for the sunflower hypergraph, but since we only use it here, it didn't seem to make sense to move it.

xgi/utils/tensor.py
nwlandry: After looking more into this, I think that it's best to stay in the `tensor` utilities module. Updating the docstring right now.

xgi/utils/tensor.py
nwlandry: Done.

xgi/utils/tensor.py
nwlandry: I added an underscore to emphasize that we're not exposing these functions to users. I moved them to the end of the file and added a note that they are helper functions. TBH, I'm not the best person to write docstrings for these functions, but if Ilya wants to give it a go, I'm fine with that.

nwlandry:
@ilyaamburg --- I think I addressed all of your comments if you want to take a final look. From looking at the test coverage, it seems like the `_get_gen_coef_fft_fast_array` function is never called with the test hypergraphs that I used. Is there a specific hypergraph that I could use to test this?

Also, could you provide docstrings in this comment thread briefly describing the two helper functions, `_get_gen_coef_subset_expansion` and `_get_gen_coef_fft_fast_array`?

We are almost there!!

ilyaamburg:
> @ilyaamburg --- I think I addressed all of your comments if you want to take a final look. From looking at the test coverage, it seems like the `_get_gen_coef_fft_fast_array` function is never called with the test hypergraphs that I used. Is there a specific hypergraph that I could use to test this?
> 
> Also, could you provide docstrings in this comment thread briefly describing the two helper functions, `_get_gen_coef_subset_expansion` and `_get_gen_coef_fft_fast_array`?
> 
> We are almost there!!

@nwlandry thanks so much for this! Hm, _get_gen_coef_fft_fast_array is really the workhorse that makes everything scale so I'm shocked it's not being called. It should get called whenever the hyperedge size is in the purple range:
Screenshot 2024-11-12 at 10 15 04 AM
As for the docstrings, here is a draft:
def _get_gen_coef_subset_expansion(edge_values, node_value, r):    
    """Computes the generating funciton coefficient of order r using subset expansion.

    Parameters
    ----------
    edge_values : NumPy array
        Array of values from the a vector corresponding to
        nodes in the hyperedge.
    node_value : float
        The value in a corresponding to the node being processed.
    r : int
        Desired order to get coefficient for.

    Returns
    -------
    float
        Generating function coefficient of order r.

    See Also
    --------
    _get_gen_coef_fft_fast_array

    References
    ----------
    Scalable Tensor Methods for Nonuniform Hypergraphs,
    Sinan Aksoy, Ilya Amburg, Stephen Young,
    https://doi.org/10.1137/23M1584472
    """
def _get_gen_coef_fft_fast_array(edge_without_node, a, node, l, r):    
    """Computes the generating funciton coefficient of order r using the Fast Fourier Transform.

    Parameters
    ----------
    edge_without_node : list
        Array of node indices corresponding to
        all nodes in the hyperedge but the one being processed.
    a : NumPy array
        The vector to multiply the tensor by.
    node : int
        The index of the node being processed.
    l : int
        Number of nodes in the hyperedge.
    r : int
        Desired order to get coefficient for.

    Returns
    -------
    float
        Generating function coefficient of order r.

    See Also
    --------
    _get_gen_coef_subset_expansion

    References
    ----------
    Scalable Tensor Methods for Nonuniform Hypergraphs,
    Sinan Aksoy, Ilya Amburg, Stephen Young,
    https://doi.org/10.1137/23M1584472
    """

nwlandry:
Everything is now addressed, and so I will now merge this once tests pass.

Issue #544, opened 2024-06-03
Numpy 2.0 to create breaking changes


nwlandry:
https://numpy.org/news/#numpy-20-release-date-june-16

I think we can place a ceiling in the requirements and then start trying to fix any breaking changes.

maximelucas:
Sounds reasonable to me

PR #603, opened 2024-10-19
Change aggregate stats type and added `unique` method.

nwlandry:
This PR does the following:
* Makes the type of the statistic aggregate methods (`max`, `min`, etc.) the classic float type instead of the new numpy type.
* added the `unique` method to the statistics interface.

### DIFF
--- a/tests/stats/test_core_stats_functions.py
+++ b/tests/stats/test_core_stats_functions.py
@@ -980,3 +980,15 @@ def test_multi_with_attrs(hyperwithattrs):
         5: [2, "blue"],
     }
     assert multi.asdict(list) == d
+
+
+def test_aggregate_stats_types(edgelist1):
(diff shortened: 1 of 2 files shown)


leotrs (approved):
Could you add a little context? What is this new numpy float type? What versions is it available in? Do we want to pin the numpy version?

nwlandry:
> Could you add a little context? What is this new numpy float type? What versions is it available in? Do we want to pin the numpy version?

@leotrs --- thanks for the review! Numpy 2.0 created breaking changes, which resulted in Issues #544 and #566. Numpy made the decision to use their own float and int datatypes (np.int32, np.int64, etc.) (not sure why), and we made the decision that the traditional Python types are preferable.

nwlandry:
The way we are handling the Numpy interface now is largely backwards-compatible, so we will also be able to use Numpy versions <2.0.

PR #608, opened 2024-10-21
Update clustering.py

cosimoagostinelli:
Fixing local_clustering_coefficient function.

I think that the two 'for' cycles are not correct because 'range(dv)' does not identify the hyper-edges to which the node n belongs.
Since we have to go over all the possible pairs of hyperedges (without repetitions) containing the node n, I guess that one possible correct implementation may be:  'for e1, e2 in combinations(ev, 2)'.

### DIFF
--- a/tests/algorithms/test_clustering.py
+++ b/tests/algorithms/test_clustering.py
@@ -1,6 +1,7 @@
 import pytest
 
 import xgi
+import networkx as nx
 from xgi.exception import XGIError
 
 
@@ -13,8 +14,9 @@ def test_local_clustering_coefficient(edgelist8):
 
     H = xgi.random_hypergraph(3, [1, 1])
     cc = xgi.local_clustering_coefficient(H)
-    true_cc = {0: 1.0, 1: 1.0, 2: 1.0}
-    assert cc == true_cc
+    true_cc = {0: 0.333, 1: 0.333, 2: 0.333}
+    for n in cc:
+        assert round(cc[n], 3) == true_cc[n]
 
     H = xgi.random_hypergraph(3, [0, 1])
     cc = xgi.local_clustering_coefficient(H)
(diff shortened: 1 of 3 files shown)


nwlandry:
Looks good @cosimoagostinelli! Thanks for the contribution!

nwlandry:
The only thing is that can you make sure that the unit tests pass? It looks like they need to be updated.

cosimoagostinelli:
I think that the output given by my version of the function is the correct one, according to the definition proposed in "Properties of metabolic graphs: biological organization or representation artifacts?" (ref. 1 of the local_clustering function) while the test does not match this defeinition.

Indeed, in the test we have 3 nodes and 4 hyperedges:  e1=[0,1], e2=[0,2], e3=[1,2], e4=[0,1,2].
Considering the node i, to compute its local hyper clustering coefficient we have to consider all the possible pairs of hyperedges containing i, sum the extra-overlap (EO) between these pairs and then divide by the number of edges containg i.

For instance, for the node 0 we have: 

EO(e1,e2) = 1 -- because nodes 1 and 2 are connected through at least another edge.
EO(e1,e4) = 0 -- because all nodes in e1 are also elements of e4.
EO(e2,e4) = 0 -- as before.

Thus, the clustering is given by (1+0+0)/3. This is true for all the nodes as they are equivalent.
So, even if it is counter intuitive, each node of this triangle should have hyper clusterig coefficient 1/3 instead of 1.

cosimoagostinelli:
Does it make sense?
Correct me if I am wrong

nwlandry:
Hi @cosimoagostinelli --- absolutely! I agree with you. What I meant was not that your new version of the function is wrong, but rather, could you update the unit tests so that they are also correct and match your new function? Sorry about the confusion.

cosimoagostinelli:
Oh, sorry, I completely misunderstood!
I will update the test as soon as possible.  :)

tests/algorithms/test_clustering.py
```diff
@@ -33,16 +35,24 @@ def test_local_clustering_coefficient(edgelist8):
...
-        4: 0.6666666666666666,
-        5: 0.0,
-        6: 0.0,
+        0: 0.47111111111111115,
+        1: 0.44833333333333336,
+        2: 0.625,
+        3: 0.625,
+        4: 0.5833333333333334,
+        5: 1.0,
+        6: 1.0
```
maximelucas, line 44: nice this make more sense, since 6's two neighbours are connected

tests/algorithms/test_clustering.py
```diff
@@ -33,16 +35,24 @@ def test_local_clustering_coefficient(edgelist8):
...
     }
     for n in cc:
         assert round(cc[n], 3) == round(true_cc[n], 3)
+        
+    G = nx.erdos_renyi_graph(50, 0.1, seed=0)
```
maximelucas, line 49:
```suggestion
    # fix #607 
    G = nx.erdos_renyi_graph(50, 0.1, seed=0)
```

maximelucas:
Mini comment but looks good to me, thanks Cosimo! I'll let Nich approve since I think he wrote this function.

nwlandry:
Great work @cosimoagostinelli !! Merging now.

nwlandry (approved):

cosimoagostinelli:
Thank you, happy to contribute!

maximelucas:
Congrats on your first PR!!

Issue #583, opened 2024-09-07
use Sphinx-gallery


maximelucas:
To improve the [Gallery](https://xgi.readthedocs.io/en/stable/gallery.html) of our online docs, [Sphinx-gallery](https://sphinx-gallery.github.io/stable/index.html) seems like the way to go.

It's used for example by [Matplotlib](https://matplotlib.org/stable/gallery/index.html) and [NetworkX](https://networkx.org/documentation/stable/auto_examples/index.html).

It works by generating a Gallery from a collection of pure python files that are executed to produce drawings.
I think it's a good idea to populate our gallery with more simple examples that people can quickly browse (see NetworkX).
We could already take many examples from the tutorials.

PR #609, opened 2024-10-23
Initial sphinx-gallery 

maximelucas:
Addressing #583: initial try for a new gallery that's automatically generated from individual small python scripts.

Right now it's accessible through a new item "Gallery of examples" in the menu. I think it could ultimately be merged with our current Gallery by porting the existing plots into this new format.

Let me know if you like it, and we  can see what to add (now or later).


nwlandry:
Looks great @maximelucas!! Only comment is that I would remove the "Gallery" tab and examples and remove the "Algorithms" subsection.

nwlandry (approved):
Looks good! Only comment is wow there are a lot of files --- how does this compare to other libraries?

maximelucas:
Thanks Nich! Most of the files were autogenerated, I removed them. 

Also for the record I only needed to:
- add a few lines to conf.py
- add single .py files for each example

Basically that's it. And to add new examples, we just need to add single `plot_*.py` in the folder `examples/` and its subfolders. 

I put the Gallery visible in the menu too (moved away from dropdown). If you like it I'll merge.

nwlandry:
Looks great! When you merge this PR, can you open an issue to add in some more examples and can you also make sure that the unused pictures and files from the old gallery are deleted?

Issue #436, opened 2023-07-27
Nodes in multilayer 3D plot are not embedded in the layer


alpatania:
Right now nodes are plotted using mpl scatter, which keep them as a perfect circle even when in the tilted layer. we should maybe use a patch?

thomasrobiglio:
I have a (super preliminary) version of a solution to this on my local xgi. This is done following [this](https://matplotlib.org/stable/gallery/mplot3d/pathpatch3d.html). I still have to work to make it consistent with all the rest... 

<img width="511" alt="Screenshot 2024-10-02 alle 16 46 46" src="https://github.com/user-attachments/assets/22bc4414-2f2d-4b68-96a5-bdaeb3f3a19a">

There is a minor problem: with this solution we are forced to use circles to draw the nodes, thus losing the possibility to pass a `node_shape` parameter. Do we want this?

If you give me the green light on this I will work on a proper PR :)

maximelucas:
Thanks Thomas! I think it's not a major problem if we loose the ability of changing the node shape. 
Would we loose anything else?

thomasrobiglio:
Yes I thought the same about the nodes' shape. What can be more relevant is the following:

I am struggling to have a node collection returned when calling the function (which then allows to have the colorbars and the cool stuff we have for the standard drawing function). 

This is because I need to create a separate `Circle` patch that is transformed into a 3D patch for each node (in each level) separately, and I have not managed to group these objects in a collection so far. I will keep looking for a solution, if you have ideas they are most welcome, I am a bit stuck--I have opened a draft pr #612

PR #612, opened 2024-10-28
Embedded nodes in `draw_multilayer`

thomasrobiglio:
Adresses #436. I am still working on this to try to change as little as possible to behavior of the function.


maximelucas:
Quoting @thomasrobiglio : 

> Yes I thought the same about the nodes' shape. What can be more relevant is the following:
> 
> I am struggling to have a node collection returned when calling the function (which then allows to have the colorbars and the cool stuff we have for the standard drawing function).
> 
> This is because I need to create a separate Circle patch that is transformed into a 3D patch for each node (in each level) separately, and I have not managed to group these objects in a collection so far. I will keep looking for a solution, if you have ideas they are most welcome, I am a bit stuck--I have opened a draft pr https://github.com/xgi-org/xgi/pull/612

Yes, I thought that might be a problem. Matplotlib has different types of [collections](https://matplotlib.org/stable/api/collections_api.html), including a [PatchCollection](https://matplotlib.org/stable/api/collections_api.html#matplotlib.collections.PatchCollection) with an example [here](https://matplotlib.org/stable/gallery/statistics/errorbars_and_boxes.html#sphx-glr-gallery-statistics-errorbars-and-boxes-py). Would this work with the 3D patches?

thomasrobiglio:
Thank you @maximelucas! this is more or less the documentation I was looking at when working on this... let's see if I can find something that works :).

thomasrobiglio:
This could be a way out... to have the circles for the nodes drawn as polygons with many sides. It's not the best thing to do, but I could not find a way to use a collection of circle objects in 3d. 

With a couple of tweaks, I should be able to make it work from outside as the function we have in the stable version now.

maximelucas:
Could you show us what the viz would look like? Is it like your example in the issue you opened?

thomasrobiglio:
Something like this---if you have more >20 points to build the polygon you cannot see that it's not an actual circle:

try

maximelucas:
@thomasrobiglio nice, even if we view it from above? Do you think the colors of the nodes are still visible enough (for example when we plot stats)?

thomasrobiglio:
thanks for the ping @maximelucas! tbh this had fallen into oblivion on my to-do list... I'll try to make some further progress

maximelucas:
Apparently you could do something like this instead:
Build real Circle patches (true circles, no polygon hack), put them in one PatchCollection with the colormap-array, then convert to 3D (with `art3d.patch_collection_2d_to_3d`), mirroring the pattern of edge_collection already uses at draw.py:843:

```python
  # one Circle per node per layer
  patches = [Circle((x, y), radius) for ... ]
  node_collection = PatchCollection(
      patches,
      array=node_fc_arr,      # values → mapped by the collection (enables colorbar)
      cmap=node_fc_cmap,
      edgecolors=node_ec,
      linewidths=node_lw,
  )
  art3d.patch_collection_2d_to_3d(node_collection, zs=z_per_patch, zdir="z",
                                  depthshade=False)   # keep stat colors faithful
  ax.add_collection3d(node_collection)
```

Pros compared to current PR solution:
  1. Real circles, not 30-gons.
  2. One ScalarMappable collection → plt.colorbar(node_collection) and stat-based coloring work, like the 2D draw_nodes.
  3. You can collect circles across all layers into a single collection (pass per-patch zs), so draw_multilayer can
  return one node_collection matching the 2D API, instead of an unreturned per-layer collection.
  4. It reuses the exact idiom already in draw.py (the edge_collection block), so it's consistent.
  5. depthshade=False keeps colormap colors accurate.
  
  Worth having a look

PR #614, opened 2024-10-28
Fix docs

nwlandry:
This PR:
* Removes "experimental" from the `DiHypergraph` class
* Makes the formatting requirements for PRs optional
* Updates the publishing process to reflect what we currently do.


maximelucas:
Thanks Nich! I think ultimately, we could aim at an even simpler process for PRs, even if just a guide for first contributors, next to the more complete pipeline. But for now this is good!

maximelucas (approved):

PR #618, opened 2024-11-12
Add docstring to watts-strogatz hypergraph

### LINKED ISSUE #615: Documentation missing for `watts_strogatz_hypergraph`
I was looking at the `generators` module and noticed that there is no documentation for [`watts_strogatz_hypergraph`](https://github.com/xgi-org/xgi/blob/98a46e9d128a4eeff3f370d2bc3d37d3d9b370d9/xgi/generators/random.py#L436). While reading the source is enough to broadly understand the generation process, I am not familiar with the origins of this model, so it would be helpful to add a description and reference. It may also be useful to give the parameters more meaningful names (or just explain that they are inputs to `ring_lattice`, plus a rewire probability).

nwlandry:
Fixes #615.

### DIFF
--- a/xgi/generators/random.py
+++ b/xgi/generators/random.py
@@ -434,6 +434,38 @@ def dcsbm_hypergraph(k1, k2, g1, g2, omega, seed=None):
 
 
 def watts_strogatz_hypergraph(n, d, k, l, p, seed=None):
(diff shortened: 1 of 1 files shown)


tlarock (approved):
Looks great, thanks @nwlandry !

PR #620, opened 2024-11-13
Remove `id` variable from XGI

### LINKED ISSUE #619: Remove instances of variable name `id` across the codebase
Some of our code uses `id` as a variable name. This is bad practice because [`id` is a built-in python function](https://docs.python.org/3/library/functions.html#id). We should change any use of `id` to something safe. My personal practice is to use `idx` or simply `i`, but even better is to use something more contextually relevant where possible.

For example, in [`uniform_hypergraph_configuration_model`](https://github.com/xgi-org/xgi/blob/85f74e7daff5bf0639c22281ca814c481566fee8/xgi/generators/uniform.py#L88) (as referenced in #617 that led to this issue), we could change `id` to `node_id`, since that is what the variable represents in this context.

One useful thing would be for someone to create a checklist of places where we need to make this change - can try to do it myself at some point, but also happy if someone beats me to it :)

nwlandry:
This fixes #619. Note that this breaks the API, because now the `add_edge` method accepts the `idx` parameter, and not the `id` parameter.


tests/readwrite/test_bipartite_edgelist.py
```diff
@@ -48,10 +48,10 @@ def test_read_bipartite_edgelist(file_string, extra_kwargs):
 
     H = xgi.read_bipartite_edgelist(filename, nodetype=int, **extra_kwargs)
     int_edgelist = [{0, 1, 2, 3}, {4}, {5, 6}, {6, 7, 8}]
-    assert [H.edges.members(id) for id in H.edges] == int_edgelist
+    assert [H.edges.members(e) for e in H.edges] == int_edgelist
```
maximelucas, line 51: maybe more specific by using `edge_id` to avoid possible confusion with edge set of members?

nwlandry: Good point! Changed to `eid`.

tests/readwrite/test_bipartite_edgelist.py
```diff
@@ -48,10 +48,10 @@ def test_read_bipartite_edgelist(file_string, extra_kwargs):
...
     H = xgi.read_bipartite_edgelist(filename, nodetype=str, **extra_kwargs)
     str_edgelist = [{"0", "1", "2", "3"}, {"4"}, {"5", "6"}, {"6", "7", "8"}]
-    assert [H.edges.members(id) for id in H.edges] == str_edgelist
+    assert [H.edges.members(e) for e in H.edges] == str_edgelist
```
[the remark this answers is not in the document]
nwlandry: Same.

PR #626, opened 2024-11-24
Modify changelog generator

nwlandry:
This PR copies the source code (with attribution) from [this repo](https://github.com/cfpb/github-changelog) to
* Change the formatting of the entries
* Add the author automatically to the entry

### DIFF
--- a/.github/workflows/check-urls.yml
+++ b/.github/workflows/check-urls.yml
@@ -14,6 +14,8 @@ jobs:
         # A comma-separated list of file types to cover in the URL checks
         file_types: .md,.py,.rst,.ipynb
 
+        exclude_files: tools/*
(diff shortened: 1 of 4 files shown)


maximelucas:
Nice! This will update the changelog, but we won't see that it works until it's merged and we make a new release?

nwlandry:
> Nice! This will update the changelog, but we won't see that it works until it's merged and we make a new release?

This isn't an GitHub Action, so you can simply run it with the command that you see in the updated `HOW_TO_CONTRIBUTE.md` and see how it works. And the output is printed to the command line so you still will have to copy-and-paste the output into the changelog.

maximelucas (approved):

Issue #625, opened 2024-11-22
Support DiHypergraph in from_bipartite_graph


colltoaction:
Hi, do you think it's possible to support `DiHypergraph`s in this API?

https://github.com/xgi-org/xgi/blob/139a2bc5f1cff7f77090d1e7b882821833b7fe0b/xgi/convert/bipartite_graph.py#L12

I can try to implement it but I wonder what the API would look like. Likely something compatible with HIF.

nwlandry:
@colltoaction --- this is a great idea! Do you want to take the lead on this? If not, happy to give it a try.

colltoaction:
Definitely! I'm thinking I can use the nx.Graph attributes like the HIF top-level metadata:

https://github.com/xgi-org/xgi/blob/0726fe1ca99b80fd57058d569ed8f6e1a2ee9b01/xgi/readwrite/hif.py#L138-L146

I would like to implement DiHypergraphs for from_bipartite at the moment, but there is more work to do for feature parity.

PR #633, opened 2024-11-25
Support DiHypergraph in from_bipartite_graph

colltoaction:
See #625.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -137,17 +137,16 @@ def bipartite_graph4():
 
 @pytest.fixture
 def bipartite_dihypergraph1():
-    G = nx.Graph()
-    G.graph["network-type"] = "directed"
+    G = nx.DiGraph()
     G.add_nodes_from([1, 2, 3, 4], bipartite=0)
     G.add_nodes_from(["a", "b", "c"], bipartite=1)
     G.add_edges_from([
-        (1, "a"),
-        (1, "b", {"direction": "tail"}),
-        (2, "b"),
-        (2, "c", {"direction": "tail"}),
-        (3, "c"),
-        (4, "a", {"direction": "tail"})])
+        ("a", 1),
+        ("b", 1, {"direction": "tail"}),
+        ("b", 2),
+        ("c", 2, {"direction": "tail"}),
+        ("c", 3),
+        ("a", 4, {"direction": "tail"})])
     return G
 
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -140,13 +140,16 @@ def bipartite_dihypergraph1():
     G = nx.DiGraph()
     G.add_nodes_from([1, 2, 3, 4], bipartite=0)
     G.add_nodes_from(["a", "b", "c"], bipartite=1)
-    G.add_edges_from([
-        ("a", 1),
-        ("b", 1, {"direction": "tail"}),
-        ("b", 2),
-        ("c", 2, {"direction": "tail"}),
-        ("c", 3),
-        ("a", 4, {"direction": "tail"})])
+    G.add_edges_from(
+        [
+            ("a", 1),
+            ("b", 1, {"direction": "tail"}),
+            ("b", 2),
+            ("c", 2, {"direction": "tail"}),
+            ("c", 3),
+            ("a", 4, {"direction": "tail"}),
+        ]
+    )
     return G
 
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -136,6 +136,22 @@ def bipartite_graph4():
 
 
 @pytest.fixture
+def bipartite_dihypergraph1():
+    G = nx.Graph()
+    G.graph["network-type"] = "directed"
+    G.add_nodes_from([1, 2, 3, 4], bipartite=0)
+    G.add_nodes_from(["a", "b", "c"], bipartite=1)
+    G.add_edges_from([
+        (1, "a"),
+        (1, "b", {"direction": "tail"}),
+        (2, "b"),
+        (2, "c", {"direction": "tail"}),
+        (3, "c"),
+        (4, "a", {"direction": "tail"})])
+    return G
+
+
+@pytest.fixture
 def attr0():
     return {"color": "brown", "name": "camel"}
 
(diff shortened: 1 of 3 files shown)


nwlandry:
@colltoaction --- thanks so much for the contribution! My biggest question is whether the default should be to read/write from a DiGraph. I think that what you've implemented makes sense as well --- maybe we could handle this with a kwarg or something?

colltoaction:
My understanding is that we always need a MultiGraph, except maybe for half-edges which I don't know.
> we mean “pseudographs” in the greatest level of generality, allowing for loops, multiple edges between vertices, and dangling half-edges
https://ncatlab.org/nlab/show/hypergraph#hypergraphs_as_2colored_graphs

I guess using other classes of graphs we get something different than a hypergraph, is this a known structure?

tlarock:
Thanks for the contribution @colltoaction ! Really useful functionality, I had actually been wondering about how to do this recently.

Wondering if it is possible to achieve this using the original `create_using` semantics, rather than conditioning on the user correctly setting `G.graph.get("network-type")` in the bipartite graph? I think we could check if `create_using` is of type `xgi.DiHypergraph` (others may reasonably disagree with this approach!), then maybe validate that `direction` is in the edge-data of the bipartite graph and raise an error if not.

My reasoning is that as written, we ask the user to specify they want a directed hypergraph in 3 different ways: (1) they have to set the "network-type" variable in the bipartite graph to the string `"directed"`; (2) they need to include `"direction"` in the edge data; and finally (3) they pass `create_using=xgi.DiHypergraph`.  My suggestion would remove (1) and validate that they have done (2).

On the latter, my potentially wrong understanding is that if a user has not done (2) correctly (e.g., if they accidentaly typed `"drection"` as the edge-data in the bipartite graph), they will get a DiHypergraph with all heads without warning - is that correct?

Related, but separate: Do we want to allow users to construct simplicial complexes from bipartite graphs? Is that implemented separately? Related because if someone passed `create_using=xgi.SimplicialComplex` I think it would run fine, but it is not clear whether we intend that or not (to me, the documentation all reads as hypergraphs, not simplicial complexes). @nwlandry let me know if you think I should open a new issue for this one or if it is dealt with somehow.

nwlandry:
@colltoaction @tlarock --- I thought more about the API for this and here is what I would propose:
* have a `create_using` keyword with default `None`
* If `create_using` is None, then `Graph` and `MultiGraph` get mapped to `Hypergraph` and `DiGraph` and `MultiDiGraph` get mapped to `DiHypergraph`
* This default behavior can be overridden with `create_using`, but maybe there are limitations like, for instance, DiGraph variants can't be mapped to Hypergraph?

tests/conftest.py
```diff
@@ -135,6 +135,21 @@ def bipartite_graph4():
     return G
 
 
+@pytest.fixture
+def bipartite_dihypergraph1():
+    G = nx.DiGraph()
+    G.add_nodes_from([1, 2, 3, 4], bipartite=0)
+    G.add_nodes_from(["a", "b", "c"], bipartite=1)
+    G.add_edges_from([
```
colltoaction, line 143: I switched the order of the edge pair so it works with `.edges`, otherwise we need to change to `.in_edges`

xgi/convert/bipartite_graph.py
```diff
@@ -73,14 +74,24 @@ def from_bipartite_graph(G, create_using=None, dual=False):
         else:
             raise XGIError("Invalid type specifier")
 
-    if not bipartite.is_bipartite_node_set(G, nodes):
-        raise XGIError("The network is not bipartite")
+    if G.is_directed():
+        H = empty_dihypergraph(create_using)
+    else:
+        if not bipartite.is_bipartite_node_set(G, nodes):
```
colltoaction, line 80: this function doesn't work with digraphs

xgi/convert/bipartite_graph.py
```diff
@@ -73,14 +74,24 @@ def from_bipartite_graph(G, create_using=None, dual=False):
...
+            raise XGIError("The network is not bipartite")
+
+        H = empty_hypergraph(create_using)
 
-    H = empty_hypergraph(create_using)
     H.add_nodes_from(nodes)
     for edge in edges:
-        nodes_in_edge = list(G.neighbors(edge))
-        H.add_edge(nodes_in_edge, idx=edge)
+        for u, v, d in G.edges(edge, data="direction"):
```
colltoaction, line 87: `edges` matches `out_edges` in the directed case

colltoaction (commented):
I updated the PR and would like to point out some comments below

xgi/convert/bipartite_graph.py
```diff
@@ -73,17 +74,37 @@ def from_bipartite_graph(G, create_using=None, dual=False):
...
+
     H.add_nodes_from(nodes)
     for edge in edges:
-        nodes_in_edge = list(G.neighbors(edge))
-        H.add_edge(nodes_in_edge, idx=edge)
+        for u, v, d in G.edges(edge, data="direction"):
+            if isinstance(H, xgi.DiHypergraph):
+                if d == "tail":
+                    H.add_node_to_edge(u, v, direction="in")
+                else:
```
tlarock, line 91: Could add:
```
elif d == "head":
   H.add_node_to_edge(u, v, direction="out")
else:
    raise XGIError("direction property not set in bipartite graph")
```

nwlandry:
@colltoaction --- I made some changes to simplify the API and to add the ability to convert a DiHypergraph to a bipartite DiGraph. Let me know what you think!

colltoaction:
@nwlandry your changes worked perfectly in my library after small tweaks. I switched from a MultiDiGraph to a DiGraph, so that makes me think what information could be missing.

Anyways it would be very helpful to have this!

nwlandry:
> @nwlandry your changes worked perfectly in my library after small tweaks. I switched from a MultiDiGraph to a DiGraph, so that makes me think what information could be missing.
> 
> Anyways it would be very helpful to have this!

Glad to hear! XGI handles multiedges with unique edge IDs, so we don't need multiedges in the bipartite directed graph. But I suppose that could be good to handle in the distant future!

PR #648, opened 2025-01-24
Fix/normalized hypergraph laplacian

### LINKED ISSUE #647: Bug: Negative eigenvalues from `normalized_hypergraph_laplacian`
# Summary
The normalized hypergraph Laplacian of a hypergraph with non-uniformly sized edges is capable of returning negative eigenvalues.

# Example
Below is a (minimal?) working example
```python
import xgi
import numpy as np
from scipy.linalg import eigh

E = [
    {1, 2, 3},
    {1, 4, 5},
    {1, 6, 7, 8},
    {4, 9, 10, 11, 12},
    {1, 13, 14, 15, 16},
    {4, 17, 18}
]
H = xgi.Hypergraph(E)

L = xgi.normalized_hypergraph_laplacian(H, sparse=False)
evals, evecs = eigh(L)

assert np.all(evals >= 0)
```

# Suspected problem 
The [referenced paper](https://proceedings.neurips.cc/paper/2006/hash/dff8e9c2ac33381546d96deea9922999-Abstract.html) [1] in the implementation states that the Laplacian should have a trivial eigenvalue, eigenvector pair with eigenvalue 0 and eigenvector $(\sqrt{k_i})_i$. Since the Laplacian is real symmetric and positive semi-definite, it should also have non-negative real eigenvalues so I believe that eigenvalues are impossible.

Looking through the [source implementation](https://github.com/xgi-org/xgi/blob/main/xgi/linalg/laplacian_matrix.py#L186), I suspect the problem is that the current implementation utilizes a simplification of the Laplacian that is only valid in 2-uniform hypergraphs as derived in [1].

# Suggested fix
I have been able to circumvent this error by relying on the definition directly in [1], i.e., 
$$\mathcal{L} = I - D_v^{-\frac{1}{2}}HD_e^{-1}H^{T}D_v^{-\frac{1}{2}}$$ (for unweighted hypergraphs).

I am happy to create a pull request!

Also, this is my first time submitting an issue to a public repo - sorry if I have done anything wrong with it!

---

[1] “Learning with Hypergraphs: Clustering, Classification, and Embedding” by Dengyong Zhou, Jiayuan Huang, Bernhard Schölkopf Advances in Neural Information Processing Systems (2006)

kaiser-dan:
# Summary
Rewrote `normalized_hypergraph_laplacian` implementation to match reference implementation. Fixes #647 .

# Description
Rewrote the matrix calculations in `normalized_hypergraph_laplacian` to match the implementation referenced here: https://github.com/xgi-org/xgi/issues/647#issuecomment-2612087221. Also adds a `weighted` Boolean parameter to allow for differentially weighting edges in the Laplacian calculation, drawn from each edge's "weight" attribute if available (default to 1 if unavailable).

# Concerns
I have only added barebone unit tests for the _effect_ of edge weights (only on their acceptable definition). I am unsure how or what the expected behavior would be for non-uniform edge weights.

# Other
~~Currently only supports strictly positive weights - a weight of zero would cause some issues with rank. That said, an edge weight of 0 could be interpreted as the subgraph with that edge removed and could be a nice way to handle zero weights in the future?~~ Deprecated with latest commit [573eaad](https://github.com/xgi-org/xgi/pull/648/commits/573eaad7c41708e9c0e640c613d3746588147014).

### DIFF
--- a/tests/linalg/test_matrix.py
+++ b/tests/linalg/test_matrix.py
@@ -608,18 +608,21 @@ def test_normalized_hypergraph_laplacian():
     assert isinstance(L2, np.ndarray)
     assert np.all(L1.toarray() == L2)
 
+    # Eigenvalues are all non-negative
     evals = eigh(L2, eigvals_only=True)
     negative_evals = list(filter(lambda e: e < 0, evals))
     assert (not negative_evals) or (np.allclose(negative_evals, 0))
 
     L3, d = xgi.normalized_hypergraph_laplacian(H, index=True)
     assert d == {0: 1, 1: 2, 2: 3, 3: 4, 4: 5, 5: 6, 6: 7, 7: 8}
+    assert np.allclose(L3.toarray(), L2)
 
     # sqrt(d) is eigenvector with eigenvalue 0
     sqrt_d = np.array([np.sqrt(d) for d in H.nodes.degree.aslist()])
     assert np.allclose(L3 @ sqrt_d, 0)
 
-    true_L = np.array(
+    # Exact Laplacian calculation
+    true_L3 = np.array(
         [
             [0.666667, -0.333333, -0.333333, -0.0, -0.0, -0.0, -0.0, -0.0],
             [-0.333333, 0.666667, -0.333333, -0.0, -0.0, -0.0, -0.0, -0.0],
@@ -631,9 +634,11 @@ def test_normalized_hypergraph_laplacian():
             [-0.0, -0.0, -0.0, -0.0, -0.0, -0.235702, -0.333333, 0.666667],
         ]
     )
-    assert np.allclose(true_L, L2)
+    assert np.allclose(true_L3, L3.toarray())
 
-    el_mwe = [
+
+def test_fix_647():
+    el = [
         {1, 2, 3},
         {1, 4, 5},
         {1, 6, 7, 8},
@@ -641,22 +646,25 @@ def test_normalized_hypergraph_laplacian():
         {1, 13, 14, 15, 16},
         {4, 17, 18},
     ]
-    H_mwe = xgi.Hypergraph(el_mwe)
-    L_mwe = xgi.normalized_hypergraph_laplacian(H_mwe, sparse=False)
-    evals_mwe = eigvalsh(L_mwe)
+    H = xgi.Hypergraph(el)
+    L = xgi.normalized_hypergraph_laplacian(H, sparse=False)
+
+    # Eigenvalues non-negative
+    evals_mwe = eigvalsh(L)
     assert np.all(evals_mwe >= 0)
 
     # Weights error handling
-    ## Default
-    L_mwe_wtd = xgi.normalized_hypergraph_laplacian(H_mwe, weighted=True, sparse=False)
-    assert np.allclose(L_mwe, L_mwe_wtd)
+    ## Default value when "weight" attribute unavailable
+    L_wtd = xgi.normalized_hypergraph_laplacian(H, weighted=True, sparse=False)
+    assert np.allclose(L, L_wtd)
 
     ## Uniform weight
-    H_mwe.set_edge_attributes(2, name="weight")
-    L_mwe_wtd_uni = xgi.normalized_hypergraph_laplacian(
-        H_mwe, weighted=True, sparse=False
+    H_wtd = H.copy()
+    H_wtd.set_edge_attributes(2, name="weight")
+    L_wtd_uniform = xgi.normalized_hypergraph_laplacian(
+        H_wtd, weighted=True, sparse=False
     )
-    assert np.allclose(2 * L_mwe_wtd - np.eye(H_mwe.num_nodes), L_mwe_wtd_uni)
+    assert np.allclose(2 * L_wtd - np.eye(H_wtd.num_nodes), L_wtd_uniform)
 
 
 def test_empty_order(edgelist6):
--- a/tests/linalg/test_matrix.py
+++ b/tests/linalg/test_matrix.py
@@ -1,6 +1,6 @@
 import numpy as np
 import pytest
-from scipy.linalg import eigh
+from scipy.linalg import eigh, eigvalsh
 from scipy.sparse import csr_array
 from scipy.sparse.linalg import norm as spnorm
 
@@ -615,6 +615,24 @@ def test_normalized_hypergraph_laplacian():
     L3, d = xgi.normalized_hypergraph_laplacian(H, index=True)
     assert d == {0: 1, 1: 2, 2: 3, 3: 4, 4: 5, 5: 6, 6: 7, 7: 8}
 
+    # sqrt(d) is eigenvector with eigenvalue 0
+    sqrt_d = np.array([np.sqrt(d) for d in H.nodes.degree.aslist()])
+    assert np.allclose(L3 @ sqrt_d, 0)
+
+    true_L = np.array(
+        [
+            [0.666667, -0.333333, -0.333333, -0.0, -0.0, -0.0, -0.0, -0.0],
+            [-0.333333, 0.666667, -0.333333, -0.0, -0.0, -0.0, -0.0, -0.0],
+            [-0.333333, -0.333333, 0.666667, -0.0, -0.0, -0.0, -0.0, -0.0],
+            [-0.0, -0.0, -0.0, 0.0, -0.0, -0.0, -0.0, -0.0],
+            [-0.0, -0.0, -0.0, -0.0, 0.5, -0.353553, -0.0, -0.0],
+            [-0.0, -0.0, -0.0, -0.0, -0.353553, 0.583333, -0.235702, -0.235702],
+            [-0.0, -0.0, -0.0, -0.0, -0.0, -0.235702, 0.666667, -0.333333],
+            [-0.0, -0.0, -0.0, -0.0, -0.0, -0.235702, -0.333333, 0.666667],
+        ]
+    )
+    assert np.allclose(true_L, L2)
+
     el_mwe = [
         {1, 2, 3},
         {1, 4, 5},
@@ -625,7 +643,7 @@ def test_normalized_hypergraph_laplacian():
     ]
     H = xgi.Hypergraph(el_mwe)
     L = xgi.normalized_hypergraph_laplacian(H, sparse=False)
-    evals = eigh(L, eigvals_only=True)
+    evals = eigvalsh(L)
     assert np.all(evals >= 0)
 
     # Weights error handling
--- a/tests/linalg/test_matrix.py
+++ b/tests/linalg/test_matrix.py
@@ -610,7 +610,7 @@ def test_normalized_hypergraph_laplacian():
 
     evals = eigh(L2, eigvals_only=True)
     negative_evals = list(filter(lambda e: e < 0, evals))
-    assert (not negative_evals) | (np.allclose(negative_evals, 0))
+    assert (not negative_evals) or (np.allclose(negative_evals, 0))
 
     L3, d = xgi.normalized_hypergraph_laplacian(H, index=True)
     assert d == {0: 1, 1: 2, 2: 3, 3: 4, 4: 5, 5: 6, 6: 7, 7: 8}
@@ -623,9 +623,9 @@ def test_normalized_hypergraph_laplacian():
         {1, 13, 14, 15, 16},
         {4, 17, 18},
     ]
-    H = xgi.Hypergraph(E)
+    H = xgi.Hypergraph(el_mwe)
     L = xgi.normalized_hypergraph_laplacian(H, sparse=False)
-    evals, evecs = eigh(L)
+    evals = eigh(L, eigvals_only=True)
     assert np.all(evals >= 0)
 
     # Weights error handling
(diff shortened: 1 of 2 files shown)


tests/linalg/test_matrix.py
```diff
@@ -605,24 +607,47 @@ def test_normalized_hypergraph_laplacian():
 
     assert isinstance(L2, np.ndarray)
     assert np.all(L1.toarray() == L2)
-    assert np.all(np.diag(L2) == 0.5)
 
-    L3, d = xgi.normalized_hypergraph_laplacian(H, index=True)
+    evals = eigh(L2, eigvals_only=True)
```
maximelucas, line 611: if you don't need the eigenvalues, you can use `eigvalsh`, which outputs them in ascending order. I think it's supposed to be faster

xgi/linalg/laplacian_matrix.py
```diff
@@ -221,16 +234,60 @@ def normalized_hypergraph_laplacian(H, sparse=True, index=False):
             "Every node must be a member of an edge to avoid divide by zero error!"
         )
 
-    D = degree_matrix(H)
-    A, rowdict = clique_motif_matrix(H, sparse=sparse, index=True)
+    if weights is not None:
+        if weights is not list:
```
maximelucas, line 238: i don't think this works? better to use isinstance to check for type

xgi/linalg/laplacian_matrix.py
```diff
@@ -221,16 +234,60 @@ def normalized_hypergraph_laplacian(H, sparse=True, index=False):
...
     else:
-        Dinvsqrt = np.diag(np.power(D, -0.5))
+        Dv_invsqrt = np.diag(np.power(Dv, -0.5))
+
+        De_inv = np.diag(list(
+            map(
+                lambda x: 1/x,
+                np.sum(H_, axis=0)
+            )
+        ))
```
maximelucas, line 278:
```suggestion
        De = np.sum(H_, axis=0)
        De_inv = np.diag(1 / De)
```
A bit more legible and hopefully a bit faster

xgi/linalg/laplacian_matrix.py
```diff
@@ -221,16 +234,60 @@ def normalized_hypergraph_laplacian(H, sparse=True, index=False):
...
+            raise XGIError("Edge weights must be given as a list!")
+        if len(weights) != H.num_edges:
+            raise XGIError("There must be as many edge weights as there are edges!")
+        if np.any(weights <= 0):
+            raise XGIError("Edge weights must be strictly positive!")
+
+
+    Dv = degree_matrix(H)
+    (
+        H_,
```
maximelucas, line 248: I was confusing H_ with a hypergraph, maybe replace by `incidence`  or other (even if they call it H in the paper

xgi/linalg/laplacian_matrix.py
```diff
@@ -221,16 +234,60 @@ def normalized_hypergraph_laplacian(H, sparse=True, index=False):
...
+            W = np.identity(H.num_edges)
+
         eye = np.eye(H.num_nodes)
 
-    L = 0.5 * (eye - Dinvsqrt @ A @ Dinvsqrt)
+
+    # PERF: There is a faster way to do this calculation if unweighted.
+    # W can be ignored entirely if unweighted, but it adds a couple conditionals and complicates the code.
+    # It is untested, but I suspect the performance change is negligible.
+    L = (eye - (Dv_invsqrt @ H_ @ W @ De_inv @ np.transpose(H_) @ Dv_invsqrt))
```
maximelucas, line 291: does this np.transpose function work when H_ is a sparse array?

tests/linalg/test_matrix.py
```diff
@@ -605,24 +607,47 @@ def test_normalized_hypergraph_laplacian():
...
+    negative_evals = list(filter(lambda e: e < 0, evals))
+    assert (not negative_evals) or (np.allclose(negative_evals, 0))
 
+    L3, d = xgi.normalized_hypergraph_laplacian(H, index=True)
```
maximelucas, line 615: values or at least eigenvalues need to be tested in this case (only sparse case)

xgi/linalg/laplacian_matrix.py
```diff
@@ -221,16 +234,60 @@ def normalized_hypergraph_laplacian(H, sparse=True, index=False):
...
     if sparse:
-        Dinvsqrt = csr_array(diags(np.power(D, -0.5)))
+        Dv_invsqrt = csr_array(diags(np.power(Dv, -0.5)))
+
+        De_inv = diags(list(
+            map(
+                lambda x: 1/x,
+                np.sum(H_, axis=0)
+            )
+        ))
```
maximelucas, line 261: maybe try to simplify similarly to the non-sparse case. Also, is np.sum(H_, axis=0) still sparse?

maximelucas:
Thanks for the PR!

It looks good, I only left a few comments to check that everything is okay in the sparse case, and to simplify a few lines. 

About the weights, I'm thinking maybe we want the argument to be boolean instead: `weighted=False`. When it's True, we could just get the weights from the "weight" attribute of the hyperedges encoded in the hypergraph.
What do you think? Or do you see a case where one would want to use other weights?

That's how we have it for the adjacency matrix now.

xgi/linalg/laplacian_matrix.py
nwlandry: The performance comment should be moved to an issue or solved with this PR.

kaiser-dan: I will remove it from the next update and make an improvement issue if/when this PR is approved

xgi/linalg/laplacian_matrix.py
kaiser-dan: Will be obsolete with move to boolean indicator for weights and edge attributes

kaiser-dan:
Okie I believe this is ready for re-review. I have addressed most of your comments @maximelucas  by directly implementing the suggestions, however, to clarify on the remaining, here are two observations:
- the `np.transpose` of a SciPy sparse array is still a sparse array
- `np.sum` of a sparse matrix is _not_ sparse, so in the middle of the `De` calculations a dense array is made, then cast back into a sparse diagonal matrix. This is introducing an additional memory allocation of $\Theta(m)$, where m is the number of hyperedges. These default to 64-bit floats in each entry.

xgi/linalg/laplacian_matrix.py
```diff
@@ -183,13 +188,15 @@ def multiorder_laplacian(
...
-def normalized_hypergraph_laplacian(H, sparse=True, index=False):
+def normalized_hypergraph_laplacian(H, weighted=False, sparse=True, index=False):
     """Compute the normalized Laplacian.
 
     Parameters
     ----------
     H : Hypergraph
         Hypergraph
+    weighted : bool, optional
+        whether or not to use hyperedge weightr, by default False (every edge weighted as 1).
```
maximelucas, line 199:
```suggestion
        whether or not to use hyperedge weights, by default False (every edge weighted as 1).
```

xgi/linalg/laplacian_matrix.py
```diff
@@ -221,16 +228,39 @@ def normalized_hypergraph_laplacian(H, sparse=True, index=False):
...
+        rowdict,
+        _,  # Discard edge name-index mapping
+    ) = incidence_matrix(H, sparse=sparse, index=True)
+
+    Dv = degree_matrix(H)
+    De = np.sum(incidence, axis=0)
 
     if sparse:
-        Dinvsqrt = csr_array(diags(np.power(D, -0.5)))
+        Dv_invsqrt = csr_array(diags(np.power(Dv, -0.5)))
```
maximelucas, line 241:
```suggestion
        Dv_invsqrt = diags_array(np.power(Dv, -0.5), format="csr")
```
Just saw [this](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.diags_array.html#scipy.sparse.diags_array) does it in one function and seems to be the new scipy go-to. Check that it works

xgi/linalg/laplacian_matrix.py
```diff
@@ -221,16 +228,39 @@ def normalized_hypergraph_laplacian(H, sparse=True, index=False):
...
-    D = degree_matrix(H)
-    A, rowdict = clique_motif_matrix(H, sparse=sparse, index=True)
+    (
+        incidence,
+        rowdict,
+        _,  # Discard edge name-index mapping
+    ) = incidence_matrix(H, sparse=sparse, index=True)
+
+    Dv = degree_matrix(H)
+    De = np.sum(incidence, axis=0)
```
maximelucas, line 238:
```suggestion
    De = np.sum(incidence, axis=0)
    if weighted:
        weights = [H.edges[edge_idx].get("weight", 1) for edge_idx in H.edges]
    
```

xgi/linalg/laplacian_matrix.py
```diff
@@ -221,16 +228,39 @@ def normalized_hypergraph_laplacian(H, sparse=True, index=False):
...
+        Dv_invsqrt = csr_array(diags(np.power(Dv, -0.5)))
+        De_inv = diags(1 / De)
+
+        if weighted:
+            W = diags([H.edges[edge_idx].get("weight", 1) for edge_idx in H.edges])
+        else:
+            W = identity(H.num_edges)
+
         eye = csr_array((H.num_nodes, H.num_nodes))
         eye.setdiag(1)
```
maximelucas, line 250:
```suggestion
        eye = eye_array(H.num_nodes, H.num_nodes), format="csr")
```
Just realised this does the same in one line

xgi/linalg/laplacian_matrix.py
```diff
@@ -221,16 +228,39 @@ def normalized_hypergraph_laplacian(H, sparse=True, index=False):
...
     else:
-        Dinvsqrt = np.diag(np.power(D, -0.5))
+        Dv_invsqrt = np.diag(np.power(Dv, -0.5))
+        De_inv = np.diag(1 / De)
+
+        if weighted:
+            W = np.diag([H.edges[edge_idx].get("weight", 1) for edge_idx in H.edges])
```
maximelucas, line 256:
```suggestion
            W = np.diag(weights)
```

xgi/linalg/laplacian_matrix.py
```diff
@@ -221,16 +228,39 @@ def normalized_hypergraph_laplacian(H, sparse=True, index=False):
...
+        else:
+            W = np.identity(H.num_edges)
```
maximelucas, line 258:
```suggestion
            W = np.eye(H.num_edges)
```

xgi/linalg/laplacian_matrix.py
```diff
@@ -221,16 +228,39 @@ def normalized_hypergraph_laplacian(H, sparse=True, index=False):
...
+
         eye = np.eye(H.num_nodes)
 
-    L = 0.5 * (eye - Dinvsqrt @ A @ Dinvsqrt)
+    L = eye - (
+        Dv_invsqrt @ incidence @ W @ De_inv @ np.transpose(incidence) @ Dv_invsqrt
+    )
```
maximelucas, line 264:
```suggestion
    L = eye - Dv_invsqrt @ incidence @ W @ De_inv @ incidence.T @ Dv_invsqrt
```
Hopefully will stay on one line, .T should work on sparse and non-sparse

xgi/linalg/laplacian_matrix.py
```diff
@@ -47,10 +47,15 @@
 from warnings import warn
 
 import numpy as np
-from scipy.sparse import csr_array, diags
+from scipy.sparse import csr_array, diags, identity
```
maximelucas, line 50: update imports with new functions

tests/linalg/test_matrix.py
```diff
@@ -605,25 +607,57 @@ def test_normalized_hypergraph_laplacian():
...
```
maximelucas, line 615: Sorry what I meant is I think we're not checking explicitly any of the sparse cases :) 
L2 is the sparse version of L3, maybe just make sure they're the same

xgi/linalg/laplacian_matrix.py
```diff
@@ -107,7 +106,7 @@ def laplacian(H, order=1, sparse=False, rescale_per_node=False, index=False):
         return (L, {}) if index else L
 
     if sparse:
-        K = csr_array(diags(degree_matrix(H, order=order)))
+        K = csr_array(diags_array(degree_matrix(H, order=order)))
```
maximelucas, line 109:
```suggestion
        K = diags_array(degree_matrix(H, order=order), format="csr")
```

maximelucas (approved):

maximelucas:
Great, thanks a lot! Thanks for bearing with me :) 
I just left a last tiny comment (the `diags_array` is already a sparse matrix), but approved the PR already. 
Once you made the tiny change I'll squash and merge (I don't think you can, but if you can feel free to do it yourself).
Congrats on your first approved PR!

kaiser-dan:
> Great, thanks a lot! Thanks for bearing with me :) I just left a last tiny comment (the `diags_array` is already a sparse matrix), but approved the PR already. Once you made the tiny change I'll squash and merge (I don't think you can, but if you can feel free to do it yourself). Congrats on your first approved PR!

Thank you! It's been fun, I hope to contribute as much as I can moving forward as well.

nwlandry:
Nice work, @kaiser-dan! And thanks @maximelucas for reviewing!

PR #650, opened 2025-01-27
Fix sparse diag by upgrading SciPy and Numpy

### LINKED ISSUE #649: sparse adjacency is wrong in "email-enron" dataset
First we load the dataset, all good:

```python
H = xgi.load_xgi_data(dataset="email-enron")
H.cleanup(isolates=True, singletons=False, connected=True, relabel=False, multiedges=False)
print(H) # --> Hypergraph named email-Enron with 143 nodes and 1459 hyperedges
print(xgi.is_connected(H)) # --> True
```
Then, we compute the adjacency matrix at order 1, sparse and not sparse: 

```python
weighted = True
order = 1

sp = False 
adj = xgi.adjacency_matrix(H, order=order, sparse=sp, weighted=weighted, index=False)

sp = True 
adj_sp = xgi.adjacency_matrix(H, order=order, sparse=sp, weighted=weighted, index=False)
adj_sp = adj_sp.todense()
```

Now, if we look at the diagonal of the non-sparse matrix, it's all zeros as expected.
But, in the sparse matrix we have many non-zero values, which is wrong:
```python
print(np.diag(adj_sp))
# array([ 0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,
#        0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,
#        0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,
#       0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,
#        0,  0,  0,  0,  0,  0,  0,  0,  0,  9,  6,  5, 13, 10, 15,  7, 15,
#       7,  1,  6,  3,  4,  9,  4,  4,  2, 16, 14,  8, 20, 13, 12, 18, 13,
#       15, 11, 13,  8, 10, 13, 13,  6,  5,  8, 10, 17,  8,  4,  3,  6,  3,
#        7, 10,  2, 17,  6,  8,  9, 19,  2,  3,  4,  4, 16,  9, 16,  8,  8,
#        3, 22,  7,  6,  2,  4,  3])
```
It's not just the diagonal: even with `weighted=True`, the sparse matrix contains values above 1:

Image

which should not happen since at order 1, two nodes share at most one pairwise edge (multiedges were removed by the cleanup above). In fact, the non-sparse adjacency matrix at order 1 contains only 0s and 1s, regardless of the value of `weighted`. 

So I'm confused. I could reproduce this only at order 1. Also, I quickly tried a few datasets and random structures, and could only reproduce this problem in the "email-enron" dataset. 

This has consequences down the line, for example my sparse Laplacian got negative eigenvalues (which should not happen, and does not happen in the non-sparse version).

maximelucas:
Fixes #649. 

The problems seems to be due to a  [bug](https://github.com/scipy/scipy/issues/21791) from SciPy that made `setdiag()` not set the whole diagonal to 0 in some cases for sparse arrays. It's been fixed by this [PR](https://github.com/scipy/scipy/pull/21792) since, and included in the 1.15 SciPy release. 

I just upped our SciPy version in the requirements and added a test.


maximelucas:
Tests for Python 3.9 fail because SciPy 15.0 requires Python 3.10-3.13 (and NumPy 1.23.5 or greater.)

Python 3.9's end of life is planned for [Oct 2025](https://devguide.python.org/versions/).
Would dropping it already be a problem? @nwlandry

nwlandry:
> Tests for Python 3.9 fail because SciPy [1.15](https://docs.scipy.org/doc/scipy-1.15.0/release/1.15.0-notes.html) requires Python 3.10-3.13 (and NumPy 1.23.5 or greater.)
> 
> Python 3.9's end of life is planned for [Oct 2025](https://devguide.python.org/versions/). Would dropping it already be a problem? @nwlandry

I think that it doesn't make sense to defer this essential fix to align with Python's release schedule. I am good with you removing v3.9 (just make a note in the PR title as well maybe?)

nwlandry (approved):
Looks great, @maximelucas! Good to squash and merge!

maximelucas:
Agreed. I also upgraded NumPy as required by SciPy. 

This was an important bug to fix, but for some reasons I only ever noticed it in "email-enron", and even then, only the smaller eigenvalues were affected. So hopefully, the bug hasn't had too much of an impact.

Issue #652, opened 2025-01-31
Problem with erdos_renyi_hypergraph generator


cosimoagostinelli:
Hi,
I am trying to run the following lines:

for s in tqdm(range(10)):
    H_s = xgi.uniform_erdos_renyi_hypergraph(200, 2, 0.0185, seed=s)

but it looks like the code gets trapped in a never-ending loop for a specific value of 'seed'.
Could you please help me with this issue?
I had a look at the source code but I did not manage to find the problem.

Thank you!

tlarock:
Hi @cosimoagostinelli, thanks for raising this issue. I ran your code on my system and did not run into an infinite loop for any seed. Could you let us know:

(1) which value of `seed` is causing the infinite loop on your end?

(2) which version of `xgi` are you running?

and

(3) do you get the same issue if you remove the `tqdm` progress bar wrapper around the `range`? I didn't use it in my test because I didn't happen to have `tqdm` installed in the environment so just want to rule out that it is somehow an issue with that library.

Thanks!

tlarock:
Dealing with `index` in [this conditional](https://github.com/xgi-org/xgi/blob/d14e3e5ff7d49769812f28abfd6371e59ea1d631/xgi/generators/uniform.py#L403) causes the infinite loop in the case of `seed=5`. We end up with `index=19900`, which is exactly 200 choose 2. We then get an infinite loop in `_index_to_edge_comb` [here](https://github.com/xgi-org/xgi/blob/d14e3e5ff7d49769812f28abfd6371e59ea1d631/xgi/generators/uniform.py#L509).

If I change the conditional to strict inequality, it works. Similarly, if I change [this line](https://github.com/xgi-org/xgi/blob/d14e3e5ff7d49769812f28abfd6371e59ea1d631/xgi/generators/uniform.py#L505) so that it does not increment index inside of `_index_to_edge_comb`, it works.

I am not sure what is the best approach for fixing, since we need to ensure that `index` always makes sense for both `_index_to_edge_comb` and `_index_to_edge_prod` (which currently work fine, as far as we know).

I think in any case it makes sense to add some kind of validation on the value of `index` to `_index_to_edge_comb` if we can do so cheaply and perhaps a test that covers this edge case.

@nwlandry it looks like you pushed this code originally, could you have a look?

cosimoagostinelli:
Hi @tlarock , thank you for the quick reply.
[...]

PR #653, opened 2025-01-31
Fix fencepost errors in `uniform_erdos_renyi_hypergraph`

tlarock:
Addresses #652.

Subtracts 1 from the `max_index` variables in `uniform_erdos_renyi_hypergraph` to fit 0-indexing in `_index_to_edge_comb` and `_index_to_edge_prod`.

Adds warnings to the latter two functions if bad `index` parameters are passed. 

Adds the test case that alerted us the issue in the first place. Also, the original test case for `multiedges=True` triggered the new warning in `_index_to_edge_prod` before the fix, so it was indeed an issue for both options.

### DIFF
--- a/benchmarks/generators.py
+++ b/benchmarks/generators.py
@@ -10,3 +10,11 @@ def test_erdos_renyi(benchmark):
         xgi.random_hypergraph(100, [0.1, 0.001])
 
     benchmark.pedantic(erdos_renyi, rounds=rounds, iterations=iterations)
+
+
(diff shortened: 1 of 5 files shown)


nwlandry (approved):
Thanks, @tlarock! Looks good.

nwlandry:
@cosimoagostinelli --- can you verify that this fixes your issue? (just in case)

cosimoagostinelli:
Yes, now everything works!
Thank you very much @nwlandry and @tlarock.

PR #655, opened 2025-02-03
Fix `fast_random_hypergraph`

nwlandry:
`fast_random_hypergraph` suffers from the same errors that were pointed out in issue #652. This attempts to fix these errors. As a side note, this now converts sampling using the log method into `np.random.geometric()`. Not sure how it compares performance-wise, but would be interesting to see.


nwlandry:
The only thing that doesn't work is the `seed` argument for `numpy.random.geometric`. Any ideas @leotrs @tlarock @maximelucas?

xgi/utils/utilities.py
```diff
@@ -553,3 +556,14 @@ def binomial_sequence(k, N):
 
 def get_network_type(H):
     return str(type(H)).split(".")[-1].split("'")[0].lower()
+
+
+def geometric(p):
```
maximelucas, line 561: document parameter `p`

xgi/generators/random.py
```diff
@@ -88,18 +88,16 @@ def fast_random_hypergraph(n, ps, order=None, seed=None):
         if p == 1:
             H.add_edges_from([e for e in combinations(nodes, d + 1)])
         elif p > 0:
-            r = random.random()
-            index = floor(log(r) / log(1 - p)) - 1  # -1 b/c zero indexing
-            max_index = comb(n, d + 1, exact=True)
+            index = geometric(p) - 1  # -1 b/c zero indexing
+            max_index = comb(n, d + 1, exact=True) - 1
```
maximelucas, line 92: it's the -1 on this line that fixes the bug right?

nwlandry: Correct. This is in line with PR #653.

xgi/utils/utilities.py
nwlandry: Done.

nwlandry:
@maximelucas, I added documentation to the new method I made and I added a test. (I verified that the test fails on the main branch)

nwlandry:
Quick aside is that comparing `np.random.geometric` vs. `xgi.geometric`, the new xgi version version is 4x faster.

nwlandry:
Okay, I fixed a bunch of random errors and should be good to go @maximelucas!

maximelucas:
> Quick aside is that comparing `np.random.geometric` vs. `xgi.geometric`, the new xgi version version is 4x faster.

Yea often numpy functions are much faster when generating a bunch of random numbers but slower for just one.

tlarock:
> Quick aside is that comparing `np.random.geometric` vs. `xgi.geometric`, the new xgi version version is 4x faster.

Faster is great! Do they give the same distribution?

nwlandry:
I can do a KS test at some point!

nwlandry:
Okay, the distributions look similar, and the methods are the same (almost), but I think the difference that the KS test picked up on is due to numerical errors in the evaluation of the log function.

PR #661, opened 2025-02-10
New iterative generators

maximelucas:
Added two iterative deterministic generators:
- pseudo-fractal simplicial complex
- apollonian complex

Code adapted from the "Higher-order Laplacian Renormalization" [repo](https://github.com/nplresearch/higher_order_LRG).


maximelucas:
@Marconurisso @MartaMorandini Can you check again that these two functions look correct? I wrote tests only for `order=2` and `n_iter` up to 2, can you add basic ones for another value of `order` if it's worth it?

maximelucas:
I've now also added an NGF function, please check that it's correct and we should add a few simple tests.

PR #662, opened 2025-02-10
Added Zulip as a sponsor and added ref data

### LINKED ISSUE #660: Add a "Using XGI-DATA" section?
A good portion of papers use XGI-DATA, but not XGI. Do we want to add these papers somewhere on our website? Like on the XGI-DATA page?

nwlandry:
This fixes #660.


maximelucas:
Thanks a lot Nich. I like the XGI-DATA tags. 

About the new support section on the homepage, it's nice to show off support, but I think it takes too much space and the logos are big and don't fit so well with the rest so it's not super pretty. Also since there's only one per category, I wouldn't differentiate between financial and infrastructural.

I would personally opt for a more minimal version. Ideas:
1. Just one line of text mentioning NSF and Zulip.
2. Just one line with the two logos but much smaller (and in black and white or somth?)
3. Just "Check out who supports XGI" with a link to another page with more details
4. Nothing on home page but somthing on a separate page.

What do you think?

nwlandry:
What about a variant of 3? A "Support" section with the NSF and Zulip logos and a "see more" link linking to a separate page

nwlandry:
I think I implemented all of your suggestions, @maximelucas!

PR #665, opened 2025-02-14
Feature/spectral clustering

### LINKED ISSUE #604: Add community detection algorithms

kaiser-dan:
# Summary
---
Partially addresses #604.

Add `xgi.communities` module.
Add [`spectral_clustering`](https://github.com/kaiser-dan/xgi/blob/feature/spectral_clustering/xgi/communities/spectral.py#L16) method to cluster a given hypergraph into K-many groups.

# Description
---
Added a new module to begin resolving #604 and adding community detection capabilities.
Added a simple version of spectral clustering, clustering the given hypergraph into a specified number of communities based on a single K-means application of the hypergraph embedding given by the spectra as computed with `normalized_hypergraph_laplacian` - a heuristic suggested in [1]. 

# Notes
---
~~If the user is agnostic to the number of groups, there is reason to believe to [spectral gap](https://en.wikipedia.org/wiki/Spectral_gap) may give a nice default number of groups. However, I am unsure how to handle the case of a hypergraph having _no_ discernible community structure. The spectral gap would be uninformative there. For this reason, I have not yet added that functionality and instead an error is raised if `k` is unspecified.~~

I've removed the ability to leave `k`, the number of groups, unspecified. It does still have a default `k=2`, however, the usage is such that the user must know the number of groups they wish to cluster the hypergraph into in order to effectively use this functionality.

# Concerns
---
~~The tests do depend on a random seed being fixed - I am unsure how the underlying OS affects this. Furthermore, I am unsure if the tests I've written, specifically [this one](https://github.com/kaiser-dan/xgi/blob/feature/spectral_clustering/tests/communities/test_spectral.py#L89), are an ideal way to test the correctness of the clustering method.~~

None.

# References
---
[1] D. Zhou, J. Huang, and B. Schölkopf, “Learning with Hypergraphs: Clustering, Classification, and Embedding,” in Advances in Neural Information Processing Systems, MIT Press, 2006. Accessed: Nov. 10, 2023. [Online]. Available: https://proceedings.neurips.cc/paper_files/paper/2006/hash/dff8e9c2ac33381546d96deea9922999-Abstract.html


xgi/communities/spectral.py
```diff
@@ -0,0 +1,115 @@
...
+
+__all__ = [
+    "spectral_clustering",
+]
+
+MAX_ITERATIONS = 10_000
+
+
+def spectral_clustering(H, k=None):
+    """Cluster into k-many groups using spectral techniques.
```
maximelucas, line 17:
```suggestion
    """Cluster hypergraph into k-many groups of nodes using spectral techniques.
```

xgi/communities/spectral.py
```diff
@@ -0,0 +1,115 @@
...
+    # Apply k-means clustering
+    _clusters = _kmeans(X, k)
+
+    # Remap to node ids
+    clusters = {rowdict[id]: cluster for id, cluster in _clusters.items()}
+
+    return clusters
+
+
+def _kmeans(X, k, seed=37):
```
maximelucas, line 73: Add docstring. Also, is there a reason to specify the seed by default instead of `seed=None` like in other functions?

xgi/communities/spectral.py
```diff
@@ -0,0 +1,115 @@
...
+                "The number of desired clusters cannot exceed the number of nodes!"
+            )
+
+    # Compute normalize Laplacian and its spectra
+    L, rowdict = normalized_hypergraph_laplacian(H, index=True)
+    evals, eigs = eigsh(L, k=k, which="SA")
+
+    # Form metric space representation
+    X = np.array(eigs)
+    print(X.shape, X)
```
maximelucas, line 62: shouldn't be there anymore?

xgi/communities/spectral.py
```diff
@@ -0,0 +1,115 @@
...
+    """Cluster into k-many groups using spectral techniques.
+
+    Compute a spectral clustering according to the heuristic suggested in [1].
+
+    Parameters
+    ----------
+    H : Hypergraph
+        Hypergraph
+    k : int, optional
+        Number of clusters to find. If unspecified, computes spectral gap.
```
maximelucas, line 26: How would this work, the spectral is not an integer in general?

xgi/communities/spectral.py
```diff
@@ -0,0 +1,115 @@
...
+    ----------
+    .. [1] Zhou, D., Huang, J., & Schölkopf, B. (2006).
+        Learning with Hypergraphs: Clustering, Classification, and Embedding
+        Advances in Neural Information Processing Systems.
+
+    """
+    if k is None:
+        raise NotImplementedError(
+            "Choosing a number of clusters organically is currently unsupported. Please specify an integer value for paramater 'k'!"
+        )
```
maximelucas, line 49: I'd say either we implement this `k=None` default option, or we remove the default value and these lines (until it's maybe implemented).

xgi/communities/spectral.py
```diff
@@ -0,0 +1,115 @@
...
+from ..core import Hypergraph
+from ..linalg.laplacian_matrix import normalized_hypergraph_laplacian
+
+from ..exception import XGIError
+
+__all__ = [
+    "spectral_clustering",
+]
+
+MAX_ITERATIONS = 10_000
```
maximelucas, line 13: Unless this parameter is planned to be used in other functions, I'd define it in _kmeans(), the only function using it? Maybe even as a parameter of the function?

maximelucas:
Thanks! @thomasrobiglio is probably best placed to review the method, I just made quick general comments more about the formatting

thomasrobiglio:
Thank you @kaiser-dan! I will have a look over the week

kaiser-dan:
> Thank you @kaiser-dan! I will have a look over the week

Sure thing, I will also be working on some better tests when I get a chance. Unclear when that will be, unfortunately.

xgi/communities/spectral.py
kaiser-dan: Ah, sorry I meant the location of the spectral gap. I will expand on this later this week with updates to the PR.

kaiser-dan: Removed in latest commits

xgi/communities/spectral.py
kaiser-dan: Added as kwarg, added kwarg propagation

kaiser-dan:
Okie I have updated the code, including a core functionality fix, add a less boring unit test, and addressed @maximelucas's review. 

@thomasrobiglio let me know if you get a chance to review and if anything pops out to you! Furthermore, I would appreciate if @maximelucas gets a chance to re-review it? :eyes:

xgi/communities/spectral.py
```diff
@@ -43,34 +41,48 @@ def spectral_clustering(H, k=None):
...
-def _kmeans(X, k, seed=37):
+def _kmeans(X, k, max_iter=1_000, seed=None):
+    """Simple k-means clustering of vectors X.
+
+    Uses Forgy method for selecting initial centroids.
+
+    Parameters
+    ----------
+    X : array-like
+        Vectors to cluster.
```
maximelucas, line 73: specify shape to avoid ambiguity, i guess it's (n,d) for n vectors in d dimensions?

thomasrobiglio (approved):
Thank you @kaiser-dan, I think this is overall in good shape and makes a very strong contribution to xgi 😄 

I have added a couple of very simple comments. 

Maybe (this can be also done in a future PR) we can have a simple example of this in the [recipes](https://xgi.readthedocs.io/en/stable/api/tutorials/recipes.html). This could showcase something like you have in the tests in which you have an assortative SBM and you recover the partitions, maybe with a drawing coloring the nodes in the different clusters.

xgi/communities/spectral.py
```diff
@@ -0,0 +1,126 @@
...
+
+__all__ = [
+    "spectral_clustering",
+]
+
+
+def spectral_clustering(H, k=2, **kwargs):
+    """Cluster hypergraph into k-many groups of nodes using spectral techniques.
+
+    Compute a spectral clustering according to the heuristic suggested in [1].
```
thomasrobiglio, line 15: I would suggest to make this more explanatory, see below for a suggestion (please check that everything is correct and feel free to change it however you think it's better)

```suggestion
    """Computes a spectral clustering in :math:`k` partitions of the input 
    hypergraph according to the heuristic presented in [1].
    
    This is done by computing the normalized Laplacian of the input hypergraph 
    and then applying :math:`k`-means clustering on the hypergraph vertices 
    represented in the :math:`k`-dimensional Euclidian space generated 
    by the eigenvectors corresponding to the :math:`k` largest eigenvalues of the 
    normalized Laplacian.
     
 ```

Comment (kaiser-dan) [reply xgi/communities/spectral.py:15]: This is mostly the procedure, albeit with the smallest eigenvectors not the largest. I am adding it to the PR now

**Comment by kaiser-dan** [comment]
Thanks, @thomasrobiglio!

> Maybe (this can be also done in a future PR) we can have a simple example of this in the [recipes](https://xgi.readthedocs.io/en/stable/api/tutorials/recipes.html). This could showcase something like you have in the tests in which you have an assortative SBM and you recover the partitions, maybe with a drawing coloring the nodes in the different clusters.

I was thinking about adding this, but was also wondering if a focused tutorial, e.g. as in the statistics tutorial, might be useful as well? At least, when modularity and perhaps some additional community detection methods are added. 

I can add the recipe later today or tomorrow, regardless.

**Comment by thomasrobiglio** [comment]
> I was thinking about adding this, but was also wondering if a focused tutorial, e.g. as in the statistics tutorial, might be useful as well? At least, when modularity and perhaps some additional community detection methods are added.

Indeed, in the end, a focus tutorial will be the way to go once we implement different methods. 
I'd say we can open an issue on this topic once we close this PR. 
A good way to proceed might be to have a recipe to showcase this method that we will then integrate in a focus tutorial once we have >1 methods. 

This PR is good to go from my side, let's wait for @maximelucas to give his blessing then you can merge!

File: xgi/communities/spectral.py [hunk 9]
```diff
@@ -0,0 +1,131 @@
...
+
+from ..exception import XGIError
+from ..linalg.laplacian_matrix import normalized_hypergraph_laplacian
+
+__all__ = [
+    "spectral_clustering",
+]
+
+
+def spectral_clustering(H, k=2, **kwargs):
```
maximelucas, line 12: since k_means only has two arguments, I would specify them here explicitly so we can also add them to the docstrings and it's easier for the user to find them

xgi/communities/spectral.py
```diff
@@ -0,0 +1,131 @@
...
+    -------
+    dict
+        A dictionary mapping node ids to their clusters. Clusters begin at 0.
+
+    Raises
+    ------
+    XGIError
+        If more groups are specified than nodes in the hypergraph.
+
+
```
maximelucas, line 41: unnecessary blank line :) run isort and black if possible, otherwise we can do it later

xgi/communities/spectral.py
```diff
@@ -0,0 +1,131 @@
...
+        Maximum number of cluster updates to compute, default 10,000.
+    seed : int, optional
+        Seed used to initialize clusters, optional.
+
+    Returns
+    -------
+    dict
+        A dictionary mapping node ids to their clusters. Clusters begin at 0.
+    """
+    rng = np.random.default_rng(seed=seed)
```
maximelucas, line 91: is this the new preferred way to set the seed? I think we've used `np.random.seed(seed)` in the past, just checking to see if we should update the others in another PR

xgi/communities/spectral.py
kaiser-dan: Hmm, intriguingly it does not seem to change anything. I did this previously in https://github.com/xgi-org/xgi/pull/665/commits/dec548dbe84b5ea9252b0db29e07320c2d1dbc1c , yet it does not remove it. 

I've added a couple quick screenshots below to show this.
image
image

kaiser-dan:
> Just left two minor comments. Also, you should make your new function available in the online docs :) To do this, add a new folder "communities" [here](https://github.com/xgi-org/xgi/tree/main/docs/source/api) with an .rst inside, e.g. like [here](https://github.com/xgi-org/xgi/blob/main/docs/source/api/generators/xgi.generators.uniform.rst).

Okie I have added the doc files, though this was my first time using Sphinx so I am a little unsure how to parse the output. Opening the built html files seems to look fine, but there was a bunch of complaints from Sphinx about multiple files. They do not appear to be related to the `communities` module, so perhaps that is a sign I have added the docs correctly?


> Also I agree with Thomas, maybe in the meantime it would be nice to have an example shown as a recipe so people can see your function in action! It can be very minimal. It can be added at the end of this [notebook](https://github.com/xgi-org/xgi/blob/main/tutorials/recipes/recipes.ipynb) (right @thomasrobiglio ?). Or it could be showcased in the [Gallery](https://xgi.readthedocs.io/en/stable/auto_examples/index.html) too, what do you think? For this you'd need to add a minimal `.py` example [here](https://github.com/xgi-org/xgi/tree/main/examples/advanced).

Unfortunately, I was unable to get to the gallery/recipe today. I plan on doing that before the weekend and would definitely appreciate some feedback on that once its up!

> Almost there!

:)

Also, I have not yet made the adjustment to the `**kwarg` removal but I have not forgotten.

xgi/communities/spectral.py
maximelucas: Ah yes, I forgot but black doesn't touch the docstrings, you can just remove this one by hand then :)

xgi/communities/spectral.py
maximelucas: Sounds good, maybe open an issue so we don't slow down this PR with this

docs/source/api/communities/xgi.communities.spectral.rst
```diff
@@ -0,0 +1,11 @@
...
+========================
+
+.. currentmodule:: xgi.communities.spectral
+
+.. automodule:: xgi.communities.spectral
+
+   .. rubric:: Functions
+
+   .. autofunction:: spectral_clustering
+   .. autofunction:: _kmeans
```
maximelucas, line 11: I think by convention in Python functions starting with `_` are "private" and not meant to be documented. And since it's not in the `__all__` it might not even work?

maximelucas:
> Okie I have added the doc files, though this was my first time using Sphinx so I am a little unsure how to parse the output. Opening the built html files seems to look fine, but there was a bunch of complaints from Sphinx about multiple files. They do not appear to be related to the `communities` module, so perhaps that is a sign I have added the docs correctly?

Yes you may have messages about other files but it should be fine. One way of checking visually that the docs are fine is by clicking here 
image
on the "docs" test which will bring you to the online docs associated to your PR.
This shows that `communities` doesn't appear yet because I forgot to tell you to add it twice to the [`api_reference.rst` ](https://github.com/xgi-org/xgi/blob/main/docs/source/api_reference.rst) (because you created a whole new module).

Also, you have a ton of changes in automated files under `docs/auto_examples/`. In theory, this file should be in your [.gitignore](https://github.com/xgi-org/xgi/blob/main/.gitignore) and those changes should not be tracked and appear.

kaiser-dan:
Okie @maximelucas sorry for the delay! I have updated the docs according to the latest. I think I got all of the discussed changes now?

xgi/communities/spectral.py
maximelucas: Only this one left!

kaiser-dan: @maximelucas Whoops sorry! I have modified this now.

nwlandry:
The ubuntu tests are unrelated and can be ignored--it's been a consistent issue with `load_xgi_data` tests when we hit the rate limit. Thanks so much for all your work on this @kaiser-dan! I think that this might call for a major release since this is an entirely new capability! 👀

maximelucas (approved):

maximelucas:
Thanks and congrats @kaiser-dan !

PR #668, opened 2025-03-20
Added a recipe for community detection

nwlandry:
Added a recipe for the new community detection algorithm.


nwlandry:
@kaiser-dan --- if you have suggestions for networks to use for the recipe, feel free to chime in!

PR #671, opened 2025-04-22
Improved the hypergraph equality method

### LINKED ISSUE #194: EdgeView comparison only checks edge ID, not contents
This is Bad.

```python3
In [11]: H1 = xgi.Hypergraph([{0, 1, 2}, {0}, {1, 2}])

In [12]: H2 = xgi.Hypergraph([{1, 2, 3}, {0}, {1, 3}])

In [13]: H1 == H2
Out[13]: False

In [14]: H1.edges == H2.edges
Out[14]: True
```

nwlandry:
This method fixes #194.

### DIFF
--- a/xgi/algorithms/properties.py
+++ b/xgi/algorithms/properties.py
@@ -33,19 +33,22 @@ def equal(H1, H2, compare_ids=True, compare_attrs=True):
         edges2_with_counts = defaultdict(lambda: 0)
         for e in H2.edges.members():
             edges2_with_counts[frozenset(e)] += 1
-
+        print(dict(edges1_with_counts), dict(edges2_with_counts))
         if edges1_with_counts != edges2_with_counts:
+            print("hi")
             return False
 
     if compare_attrs:
-        if H1._hypergraph != H2._hypergraph:
+        if H1._net_attr != H2._net_attr:
             return False
 
-        if H1._node_attrs != H2._node_attrs:
+        if H1._node_attr != H2._node_attr:
             return False
 
-        if H1._edge_attrs != H2._edge_attrs:
-            return False
+        # there's no sense in comparing edge attrs if we're not matching on IDs
+        if compare_ids:
+            if H1._edge_attr != H2._edge_attr:
+                return False
 
     return True
 
--- a/xgi/algorithms/properties.py
+++ b/xgi/algorithms/properties.py
@@ -24,6 +24,11 @@ __all__ = [
 def equal(H1, H2, compare_edge_ids=True, compare_attrs=True):
     """Evaluates whether two networks are equal.
 
+    If `compare_edge_ids=False`, then only the edges and their
+    counts (there might be multi-edges!) are compared. Similarly,
+    when comparing edge attributes in this fashion, the sorted list
+    of edge attributes associated with each edge is compared.
+
     Parameters
     ----------
     H1 : Hypergraph, DiHypergraph, or SimplicialComplex
--- a/xgi/algorithms/properties.py
+++ b/xgi/algorithms/properties.py
@@ -21,8 +21,34 @@ __all__ = [
 ]
 
 
-def equal(H1, H2, compare_ids=True, compare_attrs=True):
-    if compare_ids:
+def equal(H1, H2, compare_edge_ids=True, compare_attrs=True):
+    """Evaluates whether two networks are equal.
+
+    Important caveat: when `compare_edge_ids` is False, even
+    if `compare_attrs` is True, it will not compare the network
+    attributes and the node attributes, but not the edge
+    attributes, because there are no edge IDs to match on.
+
+    Parameters
+    ----------
+    H1 : Hypergraph, DiHypergraph, or SimplicialComplex
+        Network 1
+    H2 : Hypergraph, DiHypergraph, or SimplicialComplex
+        Network 2
+    compare_edge_ds : bool, optional
+        Whether or not to compare edge IDs, by default True
+    compare_attrs : bool, optional
+        Whether or not to compare attributes, by default True.
+        Note that if `compare_edge_ids` is False, even if `compare_attrs`
+        is True, then all other 
+        
+
+    Returns
+    -------
+    bool
+        Whether the two networks are equal
+    """
+    if compare_edge_ids:
         if H1._edge != H2._edge:
             return False
     else:
@@ -33,9 +59,7 @@ def equal(H1, H2, compare_ids=True, compare_attrs=True):
         edges2_with_counts = defaultdict(lambda: 0)
         for e in H2.edges.members():
             edges2_with_counts[frozenset(e)] += 1
-        print(dict(edges1_with_counts), dict(edges2_with_counts))
         if edges1_with_counts != edges2_with_counts:
-            print("hi")
             return False
 
     if compare_attrs:
@@ -46,7 +70,7 @@ def equal(H1, H2, compare_ids=True, compare_attrs=True):
             return False
 
         # there's no sense in comparing edge attrs if we're not matching on IDs
-        if compare_ids:
+        if compare_edge_ids:
             if H1._edge_attr != H2._edge_attr:
                 return False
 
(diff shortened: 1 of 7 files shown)


nwlandry:
@tlarock let me know if this is along the lines of what you were thinking. The only think that I didn't know how to handle is what if users want to compare edge attributes, but don't want to compare on ids. That shouldn't be allowed, right?

nwlandry:
If this looks good to you, then I will add some unit tests.

tlarock:
Okay so with #646 we can compare IDView and IDStat objects for equality by comparing their underlying dictionaries.

Since `_edge` containers are `IDDict`s, we can compare them directly, which is what happens when we have `compare_ids=True`. This will return `False` if the sets of hyperedge ids are not the same.

If instead we want to compare the members, rather than the IDs, we construct the multi-hypergraph as a dictionary of frozen sets representing hyperedges pointing to integers representing edge occurences, then compare those. So two hypergraphs are equal if they contain the same hyperedges and those hyperedges appear the same number of times.

Finally, if we want to also compare attributes, we use the same approach, comparing all of `_hypergraph`, `_node_attrs`, and `_edge_attrs`.

Assuming this is correct, I think it sounds like a good approach to addressing #194 ! I'm not 100% sure that defaulting to just comparing edge IDs is the best, but also not feeling super strongly the other way at the moment.

Note: I actually don't know what `_hypergraph` is, could you point me to it?

nwlandry:
> Okay so with #646 we can compare IDView and IDStat objects for equality by comparing their underlying dictionaries.
> 
> Since `_edge` containers are `IDDict`s, we can compare them directly, which is what happens when we have `compare_ids=True`. This will return `False` if the sets of hyperedge ids are not the same.
> 
> If instead we want to compare the members, rather than the IDs, we construct the multi-hypergraph as a dictionary of frozen sets representing hyperedges pointing to integers representing edge occurences, then compare those. So two hypergraphs are equal if they contain the same hyperedges and those hyperedges appear the same number of times.
> 
> Finally, if we want to also compare attributes, we use the same approach, comparing all of `_hypergraph`, `_node_attrs`, and `_edge_attrs`.
> 
> Assuming this is correct, I think it sounds like a good approach to addressing #194 ! I'm not 100% sure that defaulting to just comparing edge IDs is the best, but also not feeling super strongly the other way at the moment.
> 
> Note: I actually don't know what `_hypergraph` is, could you point me to it?

Good catch! This should actually be `_net_attr`. I agree---it doesn't seem super satisfying, but I suppose that it could work for now.

nwlandry:
Okay, all the unit tests should be there, @tlarock!

xgi/algorithms/properties.py
```diff
@@ -33,19 +33,22 @@ def equal(H1, H2, compare_ids=True, compare_attrs=True):
         edges2_with_counts = defaultdict(lambda: 0)
         for e in H2.edges.members():
             edges2_with_counts[frozenset(e)] += 1
-
+        print(dict(edges1_with_counts), dict(edges2_with_counts))
```
tlarock, line 36: Remove

xgi/algorithms/properties.py
```diff
@@ -33,19 +33,22 @@ def equal(H1, H2, compare_ids=True, compare_attrs=True):
...
         if edges1_with_counts != edges2_with_counts:
+            print("hi")
```
tlarock, line 38: Remove

xgi/algorithms/properties.py
```diff
@@ -33,19 +33,22 @@ def equal(H1, H2, compare_ids=True, compare_attrs=True):
...
+        if H1._net_attr != H2._net_attr:
             return False
 
-        if H1._node_attrs != H2._node_attrs:
+        if H1._node_attr != H2._node_attr:
             return False
 
-        if H1._edge_attrs != H2._edge_attrs:
-            return False
+        # there's no sense in comparing edge attrs if we're not matching on IDs
```
tlarock, line 48: Can you explain this in more detail? As a user I would be surprised that `compare_attrs=True` actually silently doesn't compare edge attributes based on another function argument. If doing it this way is the best option, we should document it well.

tlarock (changes requested):
A few stray print statements need to be taken out and I had one question.

We also need a docstring for the `equal` function, and we need to document how equality comparisons work more broadly (that might be a larger conversation beyond this PR). It should be made clear that the `equal` function is the source of all hypergraph equality, as well as exactly what kinds of equality can be determined via the function.

xgi/algorithms/properties.py
nwlandry: lol sorry about that!!!

xgi/algorithms/properties.py
nwlandry: thanks!

xgi/algorithms/properties.py
nwlandry: It was unclear to me how to match edge attributes with each other if we are explicitly stating that we are ignoring the edge IDs in the first place. If you have a suggestion, happy to hear it, but my interim solution will be to (1) rename `compare_ids` to `compare_edge_ids` and (2) add a few sentences to the doc string.

tlarock: Totally! Is it possible or would it make sense to use the defaultdict/frozenset strategy again for attributes, or not really? That's the only possibility that comes to mind.

tlarock: hmmm won't every entry in this `defaultdict` just have a list of length 1? What is the data type of `H1._edge_attr[e]` and can we just use it as the value type for the defaultdict?

Also, what exactly will need to get sorted before comparison? Didn't quite understand that bit.

I agree it is a bit convoluted, but if we can do it correctly, then I think it makes sense to implement it rather than have a somewhat convoluted caveat in the docs!

nwlandry: The issue would occur with multiedges. so the idea is that if I have three multi-edges `{1, 2, 3}` with IDs 0, 1, and 2 and associated attributes `color="red"`, `color="blue"`, and `color="yellow"`, then if I'm comparing to a hypergraph with the same edges but different edge IDs, how can I ensure that they are equal. And I guess that the way that makes sense to me to to construct a dict where keys are frozensets of edges and the values are lists of all associated attribute dictionaries. But the problem is that these lists may be out of order, so that's why I was thinking we might need to sort them. Of course this is not an issue if multi-edges aren't present, but since the `Hypergraph` object supports multi-edges, I figured that we should make sure this algo works with them.

tlarock: Thanks, that makes sense! I think it would be a good idea to implement this.

nwlandry: > hmmm won't every entry in this `defaultdict` just have a list of length 1? What is the data type of `H1._edge_attr[e]` and can we just use it as the value type for the defaultdict?
> 
> Also, what exactly will need to get sorted before comparison? Didn't quite understand that bit.
> 
> I agree it is a bit convoluted, but if we can do it correctly, then I think it makes sense to implement it rather than have a somewhat convoluted caveat in the docs!

okay, I have implemented a working version of what you suggested. It's a little gross, but it works!

tlarock: I am thinking about whether we can we use the invariant established by the `edges*_with_counts` comparison to simplify into just one loop in the attribute comparison. My idea is that we already checked that the two hypergraphs have the same set of hyperedges appearing the same number of times, so maybe we can just use one loop over the hyperedges for this part and still achieve the same goal? Let me know if that makes any sense to you haha

tlarock: Do we maintain a hyperedge members -> hyperedge ids data structure? Can't remember off the top of my head, if we don't then the above idea won't be convenient since we need the IDs to get the attributes, so we can just forget it.

nwlandry: I don't think that we do. If it's okay with you, I'll keep the two loops separate, because I don't know that it will save all that much time.

tlarock (approved):

PR #672, opened 2025-05-01
Profile import

### LINKED ISSUE #651: import xgi takes a long time
In a Jupyter Notebook (In editable mode) it can take up to 10s to run `import xgi`. Maybe lazy loading would solve this problem? NetworkX seems to have some of this [functionality](https://github.com/networkx/networkx/blob/main/networkx/lazy_imports.py).

nwlandry:
This PR
* Adds `tuna` to the list of testing dependencies with a code snippet in the `HOW_TO_CONTRIBUTE.md` file to profile the import of XGI
* Makes package imports much more specific instead of importing a whole library only to use one function.
* Moves many slow package imports inside the functions to speed up the initial load of the library.

Fixes #651.

Lingering issue: @maximelucas is it true that we are importing seaborn for one color palette only? This import costs 0.5s.


nwlandry:
@leotrs --- would love to know if you have thoughts on best practices here.

nwlandry:
Starting with a fresh branch

PR #676, opened 2025-05-15
Speedup imports

nwlandry:
This PR
* Removes seaborn as a a dependency
* Adds tuna to help profile imports


kaiser-dan:
I've made a few additional miscellaneous fixes to this PR:

- `mako` and `mako_r` are also `seaborn` palettes alongside the already addressed `crest`. Unlike the earlier changes, I did not define this color palette but instead removed it from all the examples. I replaced it with `plasma`.
- The updates to importing, I suspect from `tuna`'s insights, had some incomplete changes to calling the `factorial` function
- The node drawing function wanted `x` and `y` passed as positional arugments for the downstream `Axes3D.scatter` call

UPDATE:
I've checked the CI logs and I believe the failing tests are 429 request errors from the `xgi_data` tests and don't have to do with any of the other functionality.

PR #677, opened 2025-07-28
fix JSON error

nwlandry:
This PR converts NaNs to None in the `write_hif` function so that we no longer get JSON errors.


maximelucas:
Nice, thanks Nich! 
Should we open an issue to make sure we re-write all XGI-data files with the fix?

nwlandry:
> Nice, thanks Nich! Should we open an issue to make sure we re-write all XGI-data files with the fix?
> 
> But wait, just to check, what was the problem with the nans? It looks like JSON can handle nans: https://stackoverflow.com/questions/6601812/sending-nan-in-json

From my reading of the article, Javascript can handle NaN, but JSON cannot. I will open an issue on XGI-DATA!

PR #683, opened 2026-02-19
feat: Add nested hypergraph generators (Kim et al. 2023, Barrett et al. 2025)

jg-you:
Adds two generative models that explicitly parameterize nestedness: `random_nested_hypergraph` and `simplicial_chung_lu_hypergraph`.

Barrett et al.: Implementation of Algorithms 3 and 4 from the paper. Couldn't find much to re-use between this and  `chung_lu_hypergraph` since the methods are fundamentally different.

Kim et al.: Implements the algorithm described in Section II [here](https://arxiv.org/pdf/2303.00224). Duplicate edges are prevented during facet generation (step 1) and after rewiring (step 3) using frozenset sets. I wasn't sure whether `epsilon` was fixed or could vary by order, so I made this argument a list or a float to support both cases.

### Notes
- Tried to match the codebase conventions for parameter names: `d` for edge size, `p` for probability, `epsilon` for retention, `k1`/`k2` for degree/size sequences.
- Seed reproducibility tests only assert same-seed determinism. There's a bug in the existing code, e.g., [here](https://github.com/jg-you/xgi/blob/9bddfced00d4d601510e06dd8fba7d17dc60b57e/tests/generators/test_random.py#L43), where different seeds could lead to the edges being identical nonetheless. Did not reproduce the pattern in the new tests as a result.


xgi/generators/random.py
```diff
@@ -469,3 +471,221 @@ def watts_strogatz_hypergraph(n, d, k, l, p, seed=None):
...
+    # Build the edge-size sequence (random permutation) — Algorithm 4, step 1
+    size_sequence = list(k2.values())
+    random.shuffle(size_sequence)
+
+    edges = []  # list of frozensets
+
+    for k in size_sequence:
+        if random.random() < p:
+            # Simplicial edge — Algorithm 3
+            edges_not_k = [e for e in edges if len(e) != k]
```
nwlandry: This line will lead to quadratic complexity if I understand correctly. Do you think that it would be feasible to make a dictionary by edge sizes and use this to avoid iterating through the entire edge list each time?

jg-you:
> Ok I reviewed `random_nested_hypergraph`. Left a few comments, mainly to adhere to the new guidelines for random number generators, which we recently adopted #689.

Ty, all implemented except for the one comment I have a question about -- level of optimization desired.
Main tradeoff being code complexity vs speed.

nwlandry:
Hey @jg-you --- sorry it took me so long to reply! If the optimization is presorting the edges by size, then I would say absolutely, since that is relatively straightforward and reduces the time complexity dramatically. Thanks!

jg-you:
> Hey @jg-you --- sorry it took me so long to reply! If the optimization is presorting the edges by size, then I would say absolutely, since that is relatively straightforward and reduces the time complexity dramatically. Thanks!

Hey oops, I missed this thread.
Fixed the infinite loop with a ValueError and pushed a mini optimization for samling in the dense case.

Should be good to merge!

leotrs (approved):
Thanks JG, looks great. Sorry for the long wait on our end.

maximelucas:
Thanks @jg-you !

maximelucas (approved):

PR #685, opened 2026-02-27
Fix scipy FutureWarning in laplacian_matrix

leotrs:
## Summary
- Fix `FutureWarning` from scipy's `diags_array` about implicit int-to-float dtype casting
- Explicitly cast degree vectors to `float` before passing to `diags_array` in `laplacian_matrix.py`
- Ensure edge weights are always `float` (not `int`) in weighted laplacian computation

## Context
scipy >= 1.14 warns when `diags_array` receives integer data without an explicit dtype, since a future version will preserve integer dtype instead of casting to float. Our laplacian code relies on float arithmetic, so we now cast explicitly.

## Test plan
- [x] All 393 tests pass
- [x] No more `FutureWarning` from scipy `diags_array`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/xgi/linalg/laplacian_matrix.py
+++ b/xgi/linalg/laplacian_matrix.py
@@ -101,10 +101,11 @@ def laplacian(H, order=1, sparse=False, rescale_per_node=False, index=False):
         L = csr_array((0, 0)) if sparse else np.empty((0, 0))
         return (L, {}) if index else L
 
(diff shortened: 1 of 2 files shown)


kaiser-dan (approved):
Similar scenario to #687 , we should wait to merge into `dev` until #690 is closed.

PR #701, opened 2026-03-31
docs: introduce stats earlier, add cheat sheet, link quickstart

leotrs:
## Summary
- **Revise "XGI in 5 minutes"**: Added a new section after the histograms that introduces the stats interface — stat objects, output formats (`asdict`, `mean`, `max`), and `filterby`. Users now encounter the stats system within their first 5 minutes.
- **Add stats cheat sheet**: New one-page reference covering all stats, output formats, summary methods, filter modes, custom stats, multi-stats, and visualization integration.
- **Link quickstart notebook**: The comprehensive `quickstart.ipynb` (81 cells) was orphaned — not linked from the Getting Started index. Now listed as the 4th getting started tutorial.

## Motivation
The stats system is XGI's most powerful feature but users didn't encounter it until Focus Tutorial 6 (the 6th of 7 feature tutorials). These changes ensure users discover it early and have a quick reference to come back to.

Part of a broader docs revamp effort tracked in xgi-igm.

## Test plan
- [x] "XGI in 5 minutes" notebook runs without errors
- [x] Docs build cleanly (no RST warnings)
- [x] Full test suite passes (392 passed, 6 skipped)
- [ ] Review rendered cheat sheet on ReadTheDocs preview

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/source/api/stats.rst
+++ b/docs/source/api/stats.rst
@@ -4,6 +4,14 @@ stats package
 
 .. automodule:: xgi.stats
 
+.. seealso::
+
+   :doc:`/stats_cheatsheet`
+      Quick reference for output formats, filtering, custom stats, and more.
+
+   :doc:`/api/tutorials/focus_6`
+      Step-by-step tutorial on the stats interface.
+
 .. admonition:: Available statistics
    
    .. rubric:: Hypergraphs and simplicial complexes
--- a/docs/source/api/tutorials/getting_started.rst
+++ b/docs/source/api/tutorials/getting_started.rst
@@ -10,6 +10,7 @@ Getting started
    getting_started_1
    getting_started_2
    getting_started_3
+   getting_started_4
 
 
 
--- a/docs/source/api/tutorials/getting_started_4.nblink
+++ b/docs/source/api/tutorials/getting_started_4.nblink
@@ -1,0 +1,3 @@
+{
+    "path": "../../../../tutorials/getting_started/quickstart.ipynb"
+}
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -13,6 +13,7 @@
 
    installing
    user_guides
+   stats_cheatsheet
    api_reference
    xgi-data
    auto_examples/index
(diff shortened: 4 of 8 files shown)


maximelucas:
Thanks Leo!

- **5 minutes**: nice idea to add stats there! The new cell don't show the output (the others do), can we fix that?

- **stats cheat sheet**: great idea, a few comments
    - I wouldn't keep it in the menu bar, it takes a lot of space and doesn't feel as important as the other (few) menu items. It feels like cheat sheet like this should fit naturally somewhere in the user guides.  image
    - Can we link from somewhere to make it discoverable? Probably from where stats are referenced in tutos, and also in the reference page on stats? For example in pandas, the `pivot` [reference](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.pivot.html) links to a [user guide](https://pandas.pydata.org/docs/user_guide/reshaping.html#reshaping) explaining all the ways of reshaping a dataframe with more visual examples.
    
    - Our current [reference page for stats](https://xgi.readthedocs.io/en/stable/api/stats.html) is quite poor, and we should think of the content in the [cheat sheet ](https://xgi--701.org.readthedocs.build/en/701/stats_cheatsheet.html)and the reference together. Some things that are currently in the cheatsheet should be in the reference (too/instead), like the output formats and different ways of calling the functions. I don't have a clear idea yet, but I think it's worth thinking about.

leotrs:
Thanks for the feedback Max! Addressed points 1 and 2:

- Notebook cells now have outputs (switched to using the smaller seeded `H` so the dict fits on screen and the filter actually filters)
- Cheat sheet moved out of the top nav and into the User Guides page
- Added cross-links: stats reference page → cheat sheet + tutorial, and focus tutorial 6 → cheat sheet

Re point 3 (overlap between cheat sheet and stats reference page): the reference page is autogenerated, so the way to beef it up is through better docstrings. We'll tackle this when we do the docstring forwarding work (making `H.nodes.degree?` show the actual degree docstring instead of the generic NodeStat one). That's tracked separately.

maximelucas (approved):

PR #706, opened 2026-04-02
Optimize line graph construction

Ved235:
This changes the implementation in `xgi.convert.to_line_graph()` to avoid scanning every pair of hyperedges. Instead of using `combinations(H._edge, 2)` and checking every pair, it now only considers pairs that actually overlap.

On synthetic hypergraphs generated using codex, there was a significant performance gain:

- `2k`: `436.47 ms` -> `25.54 ms` 
- `4k`: `1936.89 ms` -> `98.47 ms`
- `6k`: `4975.49 ms` -> `219.49 ms`

### Validation
```bash
pytest -q
392 passed, 6 skipped, 117 warnings in 17.99s
```

### DIFF
--- a/xgi/convert/line_graph.py
+++ b/xgi/convert/line_graph.py
@@ -1,5 +1,6 @@
 """Method for converting to a line graph."""
 
+from collections import defaultdict
 from itertools import combinations
 
 import networkx as nx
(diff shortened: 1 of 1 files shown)


kaiser-dan (approved):
Hey @Ved235 , thanks for this work! It is a welcome change with some impressive speedups.

I've looked through the code and it seems straight-forward enough, I'll approve the PR now but merge it into the `dev` branch instead later.

nwlandry:
@kaiser-dan --- is there a timeline for this being merged?

PR #708, opened 2026-04-08
feat: forward stat function docstrings to stat objects

### LINKED ISSUE #357: stats features are hard to find and understand from docs
If a new user starts looking `H.nodes.degree` they will have a hard time finding it in the docs: 
- if they look for it in classes, `Hypergraph.py` doesn't say anything about stats and has no pointers: https://xgi.readthedocs.io/en/latest/api/classes/xgi.classes.hypergraph.Hypergraph.html#xgi.classes.hypergraph.Hypergraph.edges
- if they know to look in the stats packages (how would they?), they might end up here: https://xgi.readthedocs.io/en/latest/api/stats.html, which clearly contains more info. But the new user has to make a conceptual jump by understanding that information about `H.nodes.degree` is actually described in `xgi.stats.nodestats.degree(H)`, which looks pretty different. And there is no example to make that particular link explicit (there are somewhere else above on the page).

I'm not sure how yet, but I'm realising we might need to make the link between these things much more explicit for new users. So far, the best "explanation" of how stats work and what they are is probably the tutorial.

_Originally posted by @maximelucas in https://github.com/xgi-org/xgi/issues/202#issuecomment-1297023988_

leotrs:
## Summary

- `H.nodes.degree?` now shows the actual degree docstring (parameters, description) instead of the generic `NodeStat` class docstring
- One-line change: `self.__doc__ = func.__doc__` in `IDStat.__init__`
- Works for all stat types: node, edge, directed, custom, and with args

Closes #357 (the last remaining point — discoverability was addressed by #699)

## Test plan

- [x] New test `test_stat_docstring_forwarding` covers node stats, edge stats, args preservation, custom stats, and directed hypergraph stats
- [x] Full test suite passes (396 passed)
- [x] Manually verified in IPython

### DIFF
--- a/tests/core/test_views.py
+++ b/tests/core/test_views.py
@@ -594,3 +594,26 @@ def test_custom_stat_still_works():
 
     H = xgi.Hypergraph([[1, 2]])
     assert H.nodes.my_custom_stat.asdict() == {1: 42, 2: 42}
+
+
+def test_stat_docstring_forwarding():
+    """Stat objects should expose the underlying function's docstring."""
+    H = xgi.Hypergraph([[1, 2, 3], [3, 4, 5]])
+
+    assert "degree" in H.nodes.degree.__doc__.lower()
+    assert "order" in H.edges.order.__doc__.lower()
+
+    # Calling with args should preserve the docstring
+    assert H.nodes.degree(order=2).__doc__ == H.nodes.degree.__doc__
+
+    # Custom stats should forward their docstring too
+    @xgi.nodestat_func
+    def my_stat(net, bunch):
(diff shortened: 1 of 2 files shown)


leotrs:
As requested by @maximelucas recently

nwlandry (approved):
This is awesome, @leotrs! Good to merge, but this makes me wonder if we should remind users somewhere in the docs (maybe quickstart.ipynb or something?) that using the ? is helpful? But this is a separate issue that we could talk about tomorrow in the team meeting.

maximelucas (approved):

PR #710, opened 2026-04-26
docs: automated the generation of the "Using XGI" page

nwlandry:
This is a better way of rendering the Using XGI page, where the citations are sorted as YAML files.


maximelucas:
Hey Nich, thanks! 

For future reference, could you write somewhere what the updated process is for someone adding a project in there?
I'm guessing, still add it in same format to same file, but then run the formatting script?

A few things, looking at the [result](https://xgi--710.org.readthedocs.build/en/710/using_xgi.html): 
1. There's minor but systematic problems in the rendering of bold/italic, for example
    image
    image  
    
2. Software list is now horizontal, was that on purpose?
    image

nwlandry:
Hey @maximelucas thanks for your comments! They should all be incorporated!

maximelucas:
Looks good thanks! Already approving. 

Tiny things: maybe put Software first (need to scroll a lot to see them now), and could you add [hypersync](https://github.com/maximelucas/hypersync)?

maximelucas (approved):

Issue #718, opened 2026-05-19
Adding edges overwrite existing ones


LennyDZ:
When I create a hypergraph from an adj. matrix. And then add edge using `add_edge([v1, v2, ...])`, the method automatically set a edge id that seems to overlap an existing one, which results in an existing edge being overwritten.

This isn't intuitive, and I suppose not expected. The snippet below reproduce the issue:

```python
# Add edges [{0, 1}, {1, 2}, {2, 3}]
adj_matrix = np.array([[1, 1, 0, 0], [0, 1, 1, 0], [0, 0, 1, 1]])

xgi_graph = xgi.convert.from_incidence_matrix(adj_matrix.T)
print(xgi_graph.nodes) # [0, 1, 2, 3]
print(xgi_graph.edges.members()) # [{0, 1}, {1, 2}, {2, 3}]

xgi_graph.add_edge([0, 3]) # edge (0, 3) overwrite edge (0, 1)
print(xgi_graph.nodes) # [0, 1, 2, 3]
print(xgi_graph.edges.members()) # [{0, 3}, {1, 2}, {2, 3}]
```

maximelucas:
Thanks @LennyDZ for catching and reporting this! It seems to be because `from_incidence_matrix` calls `add_node_to_edge` which should update the hypergraph's edge uid counter, but doesn't. As a result, `add_edge` adds an edges with the first available uid, which is (wrongly) 0. I'll make a PR for this.

kaiser-dan:
Thanks for showing us @LennyDZ and the fix @maximelucas ! Closed with #719

PR #719, opened 2026-05-21
fix: issue #718

maximelucas:
`add_node_to_edge()` now correctly updates the edge uid count. 

In addition to `from_incidence_matrix()`, the function `add_node_to_edge()`  is also called by `from_bipartite_graph()`, `from_hif_dict()`, `from_bipartite_pandas_dataframe`, and `from_bipartite_edgelist`. 

The fix was a single line, and I added a test to check that it works.
This solves #718.


kaiser-dan (approved):
Another simple and concise fix, thanks @maximelucas ! Approved and merging now

PR #721, opened 2026-05-21
fix: account for idx=0 when adding edges

### LINKED ISSUE #720: `idx=0` silently fails in `add_edge` and `add_simplex` due to truthiness check on edge ID
When a user provides `idx=0` as an explicit edge ID, three methods mishandle it because they test `if idx:`/ `if not idx:` instead of checking    against `None`. Since 0 is falsy in Python, the condition takes the wrong branch.
   
  Affected methods:
  - `Hypergraph.add_edge`: if idx: skips the call to update_uid_counter, so the internal counter stays at 0 and the next
   auto-generated edge will silently overwrite edge 0.
  - `DiHypergraph.add_edge`: same issue.
  - `SimplicialComplex.add_simplex`: if not idx is True when idx=0, so the user-provided ID is silently replaced
  by an auto-generated one.

MWE:
```python
  import xgi
  H = xgi.Hypergraph()
  H.add_edge([1, 2], idx=0)
  H.add_edge([2, 3])           # overwrites edge 0!
  print(H.edges.members(dtype=dict))  # {0: {2, 3}}  — edge {1,2} is gone
```

  Fix:  Replace truthiness checks with explicit None checks
  - `if idx:` → `if idx is not None:`
  - `if not idx else idx` → `if idx is None else idx`

maximelucas:
Fixes #720 . `if idx:` / `if not idx:` evaluates incorrectly when `idx=0`. Replaced with explicit `None` checks in `Hypergraph.add_edge`, `DiHypergraph.add_edge`,  and `SimplicialComplex.add_simplex`.

### DIFF
--- a/tests/core/test_hypergraph.py
+++ b/tests/core/test_hypergraph.py
@@ -1000,3 +1000,27 @@ def test_cleanup():
     assert cleanH["name"] == "test"
 
     assert cleanH._edge == xgi.dual_dict(cleanH._node)
+
+
+def test_fix_720():
+    # idx=0 must be treated as a valid user-provided ID, not as "no ID given"
+
+    # Hypergraph.add_edge: counter must advance past 0 so the next auto-ID doesn't collide
+    H = xgi.Hypergraph()
+    H.add_edge([1, 2], idx=0)
+    H.add_edge([2, 3])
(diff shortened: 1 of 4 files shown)


kaiser-dan (approved):
Clear and concise, great catch and fix @maximelucas ! Approved

PR #722, opened 2026-05-22
Update issue templates

kaiser-dan:
Proposing adding some standard issue templates for bug reports and feature requests to help encourage new issues to contain enough information for rapid diagnosis/planning.

Note I am proposing to merge directly to main. They add markdown documents that are read by GitHub but not used in the xgi functionality itself, so we can save time bypassing dev?

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -27,6 +27,9 @@ If applicable, add screenshots to help explain your problem.
 **Desktop (please complete the following information):**
  - OS: [e.g. iOS]
  - Version [e.g. 22]
+ - Python version [e.g. 3.13.1]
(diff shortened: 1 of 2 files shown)


thomasrobiglio:
Looks good @kaiser-dan... for the bug report, do you think we should add a specific item asking if the user is running the latest version of XGI and if they observe the same undesired behavior with the current git version?

Also the same for which Python version they are running.

kaiser-dan:
> Looks good @kaiser-dan... for the bug report, do you think we should add a specific item asking if the user is running the latest version of XGI and if they observe the same undesired behavior with the current git version?
> 
> Also the same for which Python version they are running.

Yes I do, I've added two fields for Python and xgi version to make a bug even easier to track. Good idea, thank you!

thomasrobiglio (approved):

PR #724, opened 2026-05-27
fix: accept np.random.Generator as seed in spring layouts (#712)

### LINKED ISSUE #712: Error when using seed=rng in layout
Running the following 

```python
import xgi
import numpy as np

seed = 42
rng = np.random.default_rng(seed)

H = xgi.random_hypergraph(20, [0.15, 0.01], seed=rng)

pos = xgi.barycenter_spring_layout(H, seed=rng) # ---> this should work but it gives an error
```

gives the error below. The doc says it should work since @leotrs 's PR #689. Replacing `seed=seed` doesn't give an error, but doesn't do the same.

I think the error is because it's passing `seed` to networkx's `nx.spring_layout` which [says](https://networkx.org/documentation/stable/reference/generated/networkx.drawing.layout.spring_layout.html) it should accept an RNG but doesn't? In fact, this throws the same error:

```python
import networkx as nx
G = nx.fast_gnp_random_graph(10, p=0.1)
pos = nx.spring_layout(G, seed=rng)
```

So until networkx fixes the issue, maybe we remove this from our docs for the layout functions?

---
### Error 

```python
AttributeError                            Traceback (most recent call last)
Cell In[3], line 9
      5 rng = np.random.default_rng(seed)
      7 H = xgi.random_hypergraph(20, [0.15, 0.01], seed=rng)
----> 9 pos = xgi.barycenter_spring_layout(H, seed=rng)

File ~/WORK/SCIENCE/xgi/xgi/drawing/layout.py:343, in barycenter_spring_layout(H, return_phantom_graph, seed, k, **kwargs)
    339 G = _augmented_projection(H)
    341 # Creating a dictionary for the position of the nodes with the standard spring
    342 # layout
--> 343 pos_with_phantom_nodes = nx.spring_layout(G, seed=seed, k=k, **kwargs)
    345 # Retaining only the positions of the real nodes
    346 pos = {k: pos_with_phantom_nodes[k] for k in list(H.nodes)}

File <class 'networkx.utils.decorators.argmap'> compilation 4:4, in argmap_spring_layout_1(G, k, pos, fixed, iterations, threshold, weight, scale, center, dim, seed)
      2 import collections
      3 import gzip
----> 4 import inspect
      5 import itertools
      6 import re

File ~/.pyenv/versions/3.11.9/envs/venv_xgi/lib/python3.11/site-packages/networkx/drawing/layout.py:492, in spring_layout(G, k, pos, fixed, iterations, threshold, weight, scale, center, dim, seed)
    490         nnodes, _ = A.shape
    491         k = dom_size / np.sqrt(nnodes)
--> 492     pos = _fruchterman_reingold(
    493         A, k, pos_arr, fixed, iterations, threshold, dim, seed
    494     )
    495 if fixed is None and scale is not None:
    496     pos = rescale_layout(pos, scale=scale) + center

File <class 'networkx.utils.decorators.argmap'> compilation 12:4, in argmap__fruchterman_reingold_9(A, k, pos, fixed, iterations, threshold, dim, seed)
      2 import collections
      3 import gzip
----> 4 import inspect
      5 import itertools
      6 import re

File ~/.pyenv/versions/3.11.9/envs/venv_xgi/lib/python3.11/site-packages/networkx/drawing/layout.py:520, in _fruchterman_reingold(A, k, pos, fixed, iterations, threshold, dim, seed)
    516     raise nx.NetworkXError(msg) from err
    518 if pos is None:
    519     # random initial positions
--> 520     pos = np.asarray(seed.rand(nnodes, dim), dtype=A.dtype)
    521 else:
    522     # make sure positions are of same type as matrix
    523     pos = pos.astype(A.dtype)

AttributeError: 'numpy.random._generator.Generator' object has no attribute 'rand'
```

leotrs:
## Summary

Fixes #712. NetworkX's `spring_layout` calls `seed.rand(...)`, which only exists on the legacy `RandomState`. After #689 made our APIs accept `np.random.Generator` for `seed`, the four xgi layout functions that wrap `nx.spring_layout` started erroring when a `Generator` was passed.

The fix is a 5-line helper that draws a fresh integer from the rng (so the user's `Generator` still advances) and passes that to NetworkX. When NetworkX eventually accepts `Generator` natively, the helper can be deleted.

Applied at all four call sites:
- `pairwise_spring_layout`
- `barycenter_spring_layout`
- `weighted_barycenter_spring_layout`
- `bipartite_spring_layout`

## Test plan

- [x] New regression test `test_spring_layouts_accept_rng` covers all four functions with a `Generator`, an `int`, and `None`
- [x] Full test suite passes (410 passed, 6 skipped)
- [x] Manually verified the original repro from #712 now works

### DIFF
--- a/tests/drawing/test_layout.py
+++ b/tests/drawing/test_layout.py
@@ -270,3 +270,41 @@ def test_spring_layouts_accept_rng(edgelist1):
     xgi.barycenter_spring_layout(H, seed=42)
     # None still works
     xgi.barycenter_spring_layout(H, seed=None)
+
+
+def test_spring_layouts_rng_semantics(edgelist1):
+    """Document expected reproducibility semantics for the seed argument.
+
+    The conventions mirror sklearn / scipy:
+
+    * Reusing the same `Generator` instance advances its state, so two
+      consecutive calls produce different layouts.
+    * Two fresh `Generator`s constructed from the same seed produce identical
+      layouts.
+    * Passing an int seed reproduces itself, but is not equivalent to passing a
+      `Generator` constructed from that same int (the conversion path differs).
+    """
+    H = xgi.Hypergraph(edgelist1)
+
+    # Reusing one rng across calls: state advances → different outputs
+    rng = np.random.default_rng(42)
+    pos_a = xgi.barycenter_spring_layout(H, seed=rng)
+    pos_b = xgi.barycenter_spring_layout(H, seed=rng)
+    assert any(not np.allclose(pos_a[n], pos_b[n]) for n in pos_a)
+
+    # Two fresh rngs from the same seed → identical outputs
+    pos_c = xgi.barycenter_spring_layout(H, seed=np.random.default_rng(42))
+    pos_d = xgi.barycenter_spring_layout(H, seed=np.random.default_rng(42))
+    for n in pos_c:
+        assert np.allclose(pos_c[n], pos_d[n])
+
+    # Same int seed reused → identical outputs
+    pos_e = xgi.barycenter_spring_layout(H, seed=42)
+    pos_f = xgi.barycenter_spring_layout(H, seed=42)
+    for n in pos_e:
+        assert np.allclose(pos_e[n], pos_f[n])
+
+    # int seed vs rng built from same seed → NOT equivalent (different paths)
+    pos_g = xgi.barycenter_spring_layout(H, seed=42)
+    pos_h = xgi.barycenter_spring_layout(H, seed=np.random.default_rng(42))
+    assert any(not np.allclose(pos_g[n], pos_h[n]) for n in pos_g)
--- a/tests/drawing/test_layout.py
+++ b/tests/drawing/test_layout.py
@@ -253,3 +253,20 @@ def test_edge_positions_from_barycenters(edgelist1):
     for idx, e in H.edges.members(dtype=dict).items():
         mean_pos = np.mean([node_pos[n] for n in e], axis=0)
         assert np.allclose(edge_pos[idx], mean_pos)
+
+
+def test_spring_layouts_accept_rng(edgelist1):
+    """Spring layouts should accept np.random.Generator as seed (issue #712)."""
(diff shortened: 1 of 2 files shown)


maximelucas:
Thanks Leo. Quick questions:
1. Do we expect `pos1` and `pos2` to be the same? If yes maybe add a line in the test checking this.
```python
    pos_1 = xgi.pairwise_spring_layout(H, seed=rng)
    pos_2 = xgi.pairwise_spring_layout(H, seed=rng)
    pos_1==pos2 # ?
```
2. Same question here:
```python
    rng = np.random.default_rng(42)

    pos_1 = xgi.barycenter_spring_layout(H, seed=rng)
    pos_2 = xgi.barycenter_spring_layout(H, seed=42)
    pos_1==pos2 # ?
```

leotrs:
Thanks Max, good clarifying questions. The short answer to both is **no**, and the reason is the same in both cases.

**Q1** (`pos_1 = layout(H, seed=rng); pos_2 = layout(H, seed=rng)`): the rng's state advances between calls (that's the whole point of passing a single `Generator` — it tracks position in a random stream). So `pos_1 != pos_2`. To get identical layouts, either pass the same int seed twice, or construct two fresh `default_rng(same_seed)` instances. Matches sklearn / scipy convention.

**Q2** (`seed=rng` vs `seed=42` where the rng was built with `default_rng(42)`): also different. With this PR, `seed=rng` draws a fresh int via `rng.integers(0, 2**32 - 1)` and passes that to NetworkX, whereas `seed=42` is passed directly. The rng-derived int has no relation to the `42` used to construct the rng — it's just where the rng's stream happens to land first.

Pushed a follow-up commit (9d581c7) with a `test_spring_layouts_rng_semantics` test that pins down all four cases as explicit assertions, so the semantics are documented in code.

maximelucas:
Thanks Leo! We should probably have a "Randomness" page in the documentation at some point explaining all this, or referring to numpy docs.

maximelucas (approved):

PR #725, opened 2026-05-27
feat: add empties option to cleanup methods (#570)

### LINKED ISSUE #570: add method for removing empty edges
In addition, add this as an option for the `cleanup` method.

leotrs:
## Summary

Closes #570. Adds an `empties` parameter to `cleanup()` on `Hypergraph`, `DiHypergraph`, and `SimplicialComplex`. Defaults to `False` (matching the convention of the other cleanup flags like `singletons` and `multiedges`), so empty edges are removed by default.

Empty edges aren't trivial to create through normal use, but they can arise from manual data manipulation or from external data sources, and there was no clean way to strip them out.

`H.edges.empty()` already existed as a view filter, so removing them outside `cleanup` is also straightforward:

```python
H.remove_edges_from(H.edges.empty())
```

## Test plan

- [x] Added `test_cleanup_empties` regression test
- [x] Full test suite passes (410 passed, 6 skipped)

### DIFF
--- a/tests/core/test_hypergraph.py
+++ b/tests/core/test_hypergraph.py
@@ -1000,3 +1000,22 @@ def test_cleanup():
     assert cleanH["name"] == "test"
 
     assert cleanH._edge == xgi.dual_dict(cleanH._node)
+
+
+def test_cleanup_empties():
+    """cleanup should remove empty edges by default and keep them with empties=True."""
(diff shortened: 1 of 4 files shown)


maximelucas (approved):

maximelucas:
@leotrs @kaiser-dan same, merged into main

leotrs:
Same — see #727. All recovered.

PR #726, opened 2026-05-28
docs: deduplicate stats docstrings and link to algorithms (#621)

### LINKED ISSUE #621: Improve stats docstrings
Related to #166.

We should not copy-paste docs in the stats interface, but reference the docs in the algorithms module. We should link stats with algorithms with the `See Also` field.

leotrs:
## Summary

Closes #621. Several stats (in `xgi.stats.nodestats` and `xgi.stats.edgestats`) are thin wrappers around `xgi.algorithms` functions but were carrying their own copies of formulas, references, and notes. Those copies had drifted from the originals.

This PR makes the algorithms module the single source of truth:
- Wrapper stat docstrings now point at the corresponding algorithm via `See Also` rather than re-explaining
- Each algorithm gets a short note mentioning its stat accessor (e.g., \"Also accessible via the stats interface as `H.nodes.clustering_coefficient`\") so users find both calling conventions from either direction

Combined with the docstring forwarding from #708, `H.nodes.clustering_coefficient?` now shows the short pointer doc instead of duplicated content, and one click takes you to the authoritative algorithm doc.

Affects: `clustering_coefficient`, `local_clustering_coefficient`, `two_node_clustering_coefficient`, `clique_eigenvector_centrality`, `h_eigenvector_centrality`, `z_eigenvector_centrality`, `node_edge_centrality` (node and edge), `katz_centrality`, and the three local simpliciality stats.

## Test plan

- [x] Full test suite passes (409 passed, 6 skipped)
- [x] Verified forwarded docstrings render correctly via `H.nodes.<stat>.__doc__`

### DIFF
--- a/xgi/algorithms/centrality.py
+++ b/xgi/algorithms/centrality.py
@@ -39,6 +39,11 @@ def clique_eigenvector_centrality(H, tol=1e-6):
         Centrality, where keys are node IDs and values are centralities. The
         centralities are 1-normalized.
 
+    Notes
+    -----
+    Also accessible via the stats interface as
+    ``H.nodes.clique_eigenvector_centrality``.
+
     See Also
     --------
     h_eigenvector_centrality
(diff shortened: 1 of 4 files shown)


maximelucas:
Amazing!! Thanks Leo

maximelucas (approved):

maximelucas:
@leotrs @kaiser-dan same, this was merged into `main`

leotrs:
Acknowledged — see my reply on #727. All recovered via backmerge.

PR #727, opened 2026-05-28
docs: clarify that H.edges[idx] returns attributes, not members

leotrs:
## Summary

A recurring point of confusion is that `H.edges[0]` returns the (empty) attribute dict rather than the members of edge 0. The behavior is technically correct, but the docstring didn't direct users to the actual member-access methods.

This PR keeps the behavior the same (per our v1.0 stance of hardening rather than changing the API) and improves the docstring:

- Upfront note saying explicitly that this returns **attributes**, not members
- Concrete examples showing both the empty default and a case with attributes set
- `See Also` pointing at `EdgeView.members` and `NodeView.memberships`

## Test plan

- [x] Doctests pass (run with `--doctest-modules`)
- [x] Full test suite passes (415 passed, 6 skipped)

### DIFF
--- a/xgi/core/views.py
+++ b/xgi/core/views.py
@@ -596,6 +596,9 @@ class NodeView(IDView):
     `tutorial
     <https://xgi.readthedocs.io/en/stable/api/tutorials/focus_6.html>`_.
 
+    Indexing with ``H.nodes[id]`` returns the node's **attribute dictionary**,
+    not the edges it belongs to. Use :meth:`memberships` for the latter.
+
     """
 
     _id_kind = "node"
@@ -707,6 +710,9 @@ class EdgeView(IDView):
     package are also accessible via the `EdgeView` class.  For more details, see the
     `tutorial
     <https://xgi.readthedocs.io/en/stable/api/tutorials/focus_6.html>`_.
+
+    Indexing with ``H.edges[id]`` returns the edge's **attribute dictionary**,
+    not its members. Use :meth:`members` for the latter.
 
     """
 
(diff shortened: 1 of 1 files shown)


maximelucas:
Great idea to clarify this, thanks Leo. 
I wonder though in what context the user will actually see the docs of the `.getitem()` method?
Like do we get there by typing `H.nodes` and then in notebook `shift+tab` to see the docs, or ..?
This is just to make sure that users will see this explanation easily and often.

leotrs:
Good point Max. Strictly speaking the `__getitem__` docstring only shows up via `H.edges.__getitem__?` or on the API reference page, not via the more natural `H.edges[0]?` (which evaluates first and shows the dict's docstring instead).

Pushed a follow-up that also adds a short attr-vs-member note to the `NodeView` and `EdgeView` class-level docstrings. Now `H.nodes?` / `H.edges?` (a common discovery move) surfaces the distinction too.

maximelucas:
Great!

maximelucas (approved):

maximelucas:
Oops, this was merged into `main` rather than `dev`. @leotrs @kaiser-dan ?
Btw, by default a new PR merges into `main`, can we change a setting so that the default is into `dev`?

leotrs:
Sorry about that, totally my mistake — I missed the dev/main convention. Just retargeted #728 and #729 to dev, backmerged main into dev to recover the stranded commits, and opened #734 to document the workflow in CLAUDE.md so it doesn't happen again. Going to change the default branch to dev now so PRs target dev by default.

PR #728, opened 2026-05-28
feat: add H.nodes.pos stat accessor for node positions (#624)

### LINKED ISSUE #624: Function for extracting node positions from attributes
Related to #623 and #179.

It would be great to have a function to do the following:
```python
H = xgi.ring_lattice(n, d, k, l, store_pos=True)
pos = xgi.pos_from_attributes(H, attr="pos")
xgi.draw(H, pos=pos)
```

leotrs:
## Summary

Closes #624. Adds `pos` as a first-class stat on `NodeView` and `DiNodeView` — a discoverable shortcut for the conventional `\"pos\"` node attribute that xgi's drawing functions use.

Why promote `pos` to a stat? It's already a privileged attribute name in the library (the whole reason the issue exists), and the descriptor pattern from #699 means it now appears in `dir()`, tab completion, and IDE autocomplete. So the natural call site becomes:

```python
xgi.draw(H, pos=H.nodes.pos.asdict())
```

This is what the original issue asked for, achieved without adding a new top-level helper. Nodes without a `\"pos\"` attribute map to `None` (same as `attrs(\"pos\")` does today).

## Changes

- `xgi/stats/nodestats.py`: new `pos` function (3-line body, full docstring)
- `xgi/stats/dinodestats.py`: same, for `DiHypergraph`
- `xgi/core/views.py`: register on `NodeView` and `DiNodeView`
- `xgi/core/views.pyi`: type stubs
- `xgi/drawing/draw.py`: point the `pos` parameter docstring at `H.nodes.pos.asdict()`
- `docs/source/api/stats.rst`: list it under both node-stats sections
- `docs/source/stats_cheatsheet.rst`: add a row
- `tutorials/recipes/recipes.ipynb`: new recipe showing the layout-store-redraw pattern

## Test plan

- [x] New `test_pos_stat` covers `Hypergraph` + `DiHypergraph`, with and without the attribute set, and equivalence to `attrs(\"pos\")`
- [x] Full test suite passes (411 passed, 6 skipped)
- [x] Doctest on the new `pos` function passes

### DIFF
--- a/docs/source/api/stats.rst
+++ b/docs/source/api/stats.rst
@@ -28,6 +28,7 @@ stats package
       ~xgi.stats.nodestats.h_eigenvector_centrality
       ~xgi.stats.nodestats.local_clustering_coefficient
       ~xgi.stats.nodestats.node_edge_centrality
+      ~xgi.stats.nodestats.pos
       ~xgi.stats.nodestats.two_node_clustering_coefficient
 
    *Statistics of edges*
@@ -78,6 +79,7 @@ stats package
       ~xgi.stats.dinodestats.degree
       ~xgi.stats.dinodestats.in_degree
       ~xgi.stats.dinodestats.out_degree
+      ~xgi.stats.dinodestats.pos
 
    *Statistics of edges*
 
--- a/docs/source/stats_cheatsheet.rst
+++ b/docs/source/stats_cheatsheet.rst
@@ -96,6 +96,8 @@ Available node statistics
      - ``H.nodes.katz_centrality``
    * - ``attrs``
      - ``H.nodes.attrs('color')``
+   * - ``pos``
+     - ``H.nodes.pos`` (shortcut for ``H.nodes.attrs('pos')``)
 
 Available edge statistics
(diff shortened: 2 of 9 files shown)


nwlandry:
This seems cool, @leotrs! Any idea why the tests are failing? But other than that, I think it's good to go. (Probably best to keep the lattice generative model re-write separate?)

leotrs:
@nwlandry the failing test is `drawing/test_draw.py::test_issue_515` with a `RuntimeWarning: overflow encountered in dot` — same intermittent `np.dot` flake @kaiser-dan flagged on #737. Not caused by this PR (nothing here touches the drawing math path). I'll rerun.

On the lattice re-write: agreed, keeping it separate. The lattice work you're thinking of is #729 (`with_positions` on `ring_lattice`); this PR is only the `pos` stat accessor and the drawing convenience. No lattice generator code is touched here.

nwlandry:
As mentioned in #729, I fixed the flaky error, so squashing and merging now.

PR #729, opened 2026-05-28
feat: add with_positions option to ring_lattice (#623)

### LINKED ISSUE #623: Store lattice positions as node attributes
Related to #179.

When generating triangular lattices, it should be possible to store the lattice positions as attributes for ease of plotting. This should be an optional argument in the lattice generator functions such as [`ring_lattice`](https://github.com/xgi-org/xgi/blob/8a6bac907be8bca03df4812c60b7723fbc6c3dd9/xgi/generators/lattice.py#L17).
```python
xgi.ring_lattice(n, d, k, l, store_pos=True)
```

leotrs:
## Summary

Closes #623. Adds a `with_positions=False` keyword argument to `ring_lattice`. When True, stores each node's coordinate on a unit circle as the `\"pos\"` node attribute. Naming follows the NetworkX convention (e.g., `triangular_lattice_graph(with_positions=True)`) as Max suggested in the issue.

Combined with #728 (`H.nodes.pos` accessor), the natural workflow becomes:

```python
H = xgi.ring_lattice(8, 3, 2, 0, with_positions=True)
xgi.draw(H, pos=H.nodes.pos.asdict())
```

`ring_lattice` is currently the only lattice generator in xgi, so this is a one-function change. Future lattice generators (e.g., triangular) should follow the same convention.

## Test plan

- [x] `test_ring_lattice_with_positions` covers both default (no pos) and `with_positions=True` (verifies unit-circle coordinates)
- [x] Full test suite passes (411 passed, 6 skipped)

### DIFF
--- a/tests/generators/test_lattice.py
+++ b/tests/generators/test_lattice.py
@@ -28,3 +28,19 @@ def test_ring_lattice():
     # k < 0 test
     with pytest.raises(ValueError):
         xgi.ring_lattice(5, 2, -1, 0)
+
+
+def test_ring_lattice_with_positions():
+    from math import cos, pi, sin
+
+    # default: no pos stored
+    H = xgi.ring_lattice(5, 2, 2, 0)
+    assert H.nodes.attrs("pos").asdict() == {i: None for i in range(5)}
+
+    # with_positions=True: each node has a unit-circle coordinate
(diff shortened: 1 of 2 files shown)


nwlandry:
Ah, this PR answered my question on PR #728. Thanks for this! Looks great!

nwlandry:
Fixed the plotting error! Merging now.

PR #732, opened 2026-05-29
Fixes kwarg validation bugs in `draw()` reported in #730 and #731.

### LINKED ISSUE #500: change size of node labels?
it is not directly related to the node color, but more node size and node labels, in the current https://xgi.readthedocs.io/en/stable/api/drawing/xgi.drawing.draw.html#module-xgi.drawing.draw it seems not clear to me if you also allow to specify the changed size of the text of node lables. 
and if yes, to document it on the drawing function maybe? 
thanks!

_Originally posted by @Liyubov in https://github.com/xgi-org/xgi/issues/499#issuecomment-1885724024_

### LINKED ISSUE #730: `draw()` kwargs validation raises false errors and uses wrong exception type
`draw()` forwards `**kwargs` to `settings`,  `draw_node_labels()` and `draw_hyperedge_labels()`, but the validation logic in `draw_nodes()` and `draw_hyperedges()` has two bugs that cause incorrect errors or confusing behavior.

---

**Bug A — False `ValueError` when mixing node and hyperedge labels**

Passing a valid hyperedge-label kwarg alongside `node_labels=True` triggers an error, because the node-label validation block sees an unrecognized key and raises before the hyperedge-label block ever runs.

```python
import xgi

H = xgi.Hypergraph([[1, 2, 3], [3, 4]])
xgi.draw(H, node_labels=True, hyperedge_labels=True, font_size_edges=14)
# ValueError: Received invalid argument(s): font_size_edges
# Expected: draws fine with larger hyperedge label font
```

---

**Bug B — `ValueError` when passing size-rescaling kwargs alongside labels**

Size-rescaling keys like `min_node_size` are valid `draw()` kwargs, but they are not in `draw_node_labels()`'s signature. The validation block incorrectly flags them as invalid.

```python
import xgi

H = xgi.Hypergraph([[1, 2, 3], [3, 4]])
xgi.draw(H, node_labels=True, min_node_size=3)
# ValueError: Received invalid argument(s): min_node_size
# Expected: draws fine with rescaled node sizes
```

---

**Bug C — Wrong exception type**

Python convention for unexpected keyword arguments is `TypeError`. The current `ValueError` is inconsistent with what users expect and with how Python itself behaves.

```python
import xgi

H = xgi.Hypergraph([[1, 2, 3]])
try:
    xgi.draw(H, node_labels=True, not_a_real_kwarg=99)
except ValueError:
    print("got ValueError")   # current (unexpected)
except TypeError:
    print("got TypeError")    # expected, consistent with Python conventions
```

---

**Suggested fix:** split `**kwargs` into its three buckets (settings keys, node label keys, edge label keys) at the top of `draw()` itself, validate once against the union of all known keys, and raise `TypeError` for unknowns. This removes the need for per-function validation entirely. #680 flagged that the documentation wasn't explicit on how kwargs were used.

### LINKED ISSUE #731: Bug - `draw()` silently ignores invalid label kwargs when labels are disabled
The kwargs validation in `draw_nodes()` and `draw_hyperedges()` only runs when `node_labels` / `hyperedge_labels` is truthy. When labels are disabled (the default), any kwarg — including typos — is silently ignored, giving no feedback to the user.

```python
import xgi

H = xgi.Hypergraph([[1, 2, 3], [3, 4]])

# Typo in kwarg name, labels disabled (default)
xgi.draw(H, node_labels=False, font_siz_nodes=20)
# No error, no warning. The typo is silently swallowed.

# Now enable labels — suddenly the typo surfaces
xgi.draw(H, node_labels=True, font_siz_nodes=20)
# ValueError: Received invalid argument(s): font_siz_nodes
```

The user gets no feedback in the first call, which makes this class of mistake hard to diagnose — especially since `node_labels=False` is the default, so the labels-off path is by far the most common.

**Suggested fix:** validate `**kwargs` unconditionally, regardless of whether labels are enabled. Ideally this is done once at the top of `draw()` so the check is never accidentally skipped. See related issue #[N] for a proposed refactor that addresses both issues together.

Related to #730 and #680

maximelucas:
**Root cause:** `**kwargs` in `draw()` served three purposes simultaneously:  size-rescaling settings (`min_node_size` etc.), node-label styling, and  edge-label styling — with validation only running conditionally when labels  were enabled, and only against one label function's signature at a time. The docstrings were also unclear.

  **Bugs fixed:**
  - **#731:** Typos in kwargs were silently ignored when `node_labels=False`   (the default), making them impossible to catch without enabling labels.
  - **#730 A:** Passing a valid edge-label kwarg (e.g. `font_size_edges`) while   `node_labels=True` raised a false error, because the node-label validator   ran first and rejected kwargs intended for `draw_hyperedge_labels`.
  - **#730B:** Passing a settings kwarg (e.g. `min_node_size`) while    `node_labels=True` raised a false error, because `min_node_size` is not in     `draw_node_labels`'s signature.
  - **#730C:** Invalid kwargs raised `ValueError`; Python convention for    unexpected keyword arguments is `TypeError`.

  **Fix:** Split `**kwargs` into three non-overlapping buckets at the top of  `draw()` — settings keys, node-label keys, edge-label keys — validate once  against their union (unconditionally, before anything else), and route each  bucket only to the function that needs it. Removed the now-redundant  per-function validation blocks in `draw_nodes()` and `draw_hyperedges()`. Made docstrings explicit about it.

  ## Test plan

  - [x] New test `test_draw_kwargs_validation` covers all four bug scenarios    plus positive cases confirming valid kwargs from all three buckets work
  - [x] All existing drawing tests pass


leotrs (approved):
Nice fix Max. The three-bucket split + single upfront validation is much cleaner than the per-function conditional checks. Tests are thorough.

PR #735, opened 2026-05-31
Optimize `node_swap`

### LINKED ISSUE #642: make node_swap faster
The `node_swap` function is slow for many swaps but can be made much faster with a dictionary mapping

maximelucas:
Optimize `node_swap` by replacing the three-loop temp-ID swap with a  single-pass dictionary mapping. Closes #642 

  **Old approach:** used a temporary node ID to swap in three sequential  passes — replace `nid1 → temp`, then `nid2 → nid1`, then `temp → nid2` —  plus a `deepcopy` of the full edge dict.

  **New approach:** builds a `{nid1: nid2, nid2: nid1}` mapping and remaps  each edge's members in a single dict comprehension. Handles the edge case  where both nodes appear in the same edge correctly (the swap is applied
  atomically per member, leaving the set unchanged). Removes the `id_temp` parameter and the `deepcopy` import entirely.

  **Timing** 

Single swap on an `xgi.random_hypergraph` averaged over runs.

  | nodes | edges  | before (ms) | after (ms) | speedup |
  |------:|-------:|------------:|-----------:|--------:|
  |    50 |    324 |        3.13 |       2.08 |   1.51x |
  |   200 |  7 533 |       70.69 |      46.71 |   1.51x |
  |   500 | 24 396 |      248.70 |      160.23 |  1.55x |
  | 1 000 | 93 195 |      958.04 |      643.82 |  1.49x |

The ~1.5x speedup is consistent across sizes. 

**Tests**

Added some tests for uncovered cases.

### DIFF
--- a/tests/generators/test_randomizing.py
+++ b/tests/generators/test_randomizing.py
@@ -79,6 +79,7 @@ def test_shuffle_hyperedges():
 def test_node_swap(edgelist8):
 
     H0 = xgi.Hypergraph(edgelist8)
+    edges_orig = dict(H0._edge)
 
     H = xgi.node_swap(H0, 2, 5)  # all orders
     edges_new = {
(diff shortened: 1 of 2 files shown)


leotrs (approved):
Nice cleanup Max. The dict-mapping comprehension is much easier to read than the three-pass temp-ID swap, the deepcopy goes away, and the new tests cover the previously implicit edge cases (self-inverse on both-in-one-edge, original-not-mutated, attribute preservation).

One thing worth flagging: removing the `id_temp` parameter is a backward-incompatible change. Anyone passing `node_swap(H, a, b, id_temp=...)` will get a `TypeError`. The arg was always implementation-detail leakage so dropping it is the right call, but worth mentioning in the changelog so it's an intentional v1.0 break rather than a silent one.

PR #736, opened 2026-06-01
fix: only proxy actual stats through Hypergraph.__getattr__ (#408)

### LINKED ISSUE #408: All attributes of views are being exposed via Hypergraph
The stats package was designed so that users could immediately access stats from the main hypergraph object:
```python3
>>> H.degree()
{1: 1, 2: 1, 3: 1, 4: 1, 5: 1, 6: 2, 8: 1, 7: 1}
```

However, this was done by exposing all of IDView's methods (which include stats) but also include others that we don't want to expose via Hypergraph, such as:
```python3
>>> H.filterby()
TypeError: IDView.filterby() missing 2 required positional arguments: 'stat' and 'val'

>>> H.neighbors()
TypeError: IDView.neighbors() missing 1 required positional argument: 'id'

>>> H.memberships()
AttributeError: 'dict' object has no attribute 'asdict'
```

leotrs:
## Summary

Closes #408. `Hypergraph.__getattr__` (and `DiHypergraph.__getattr__`) blindly delegated every attribute lookup to the underlying views and wrapped whatever it found in a function that calls `.asdict()`. That works for actual stats but produces confusing errors when users hit view methods that happen to share a name with no stat:

```python
>>> H.filterby()
TypeError: IDView.filterby() missing 2 required positional arguments: 'stat' and 'val'

>>> H.memberships()
AttributeError: 'dict' object has no attribute 'asdict'

>>> H.neighbors()
TypeError: IDView.neighbors() missing 1 required positional argument: 'idx'
```

## Fix

Tighten the delegation check from `stat is None` to `not isinstance(stat, IDStat)`. Built-in stats (registered via `_stat_property`) and user-defined stats (via `@nodestat_func` etc.) both return `IDStat` instances, so they continue to work. Non-stat view methods (`filterby`, `neighbors`, `memberships`, `isolates`, `singletons`, `empty`, `members`, `dimembers`, `head`, `tail`, ...) now correctly raise `AttributeError` at lookup time with the existing helpful message.

Same one-line change in both `Hypergraph` and `DiHypergraph`.

**Safety check:** grepped `xgi/`, `tests/`, and `tutorials/` for `H.filterby(`, `H.neighbors(`, `H.memberships(`, etc. — zero hits. Nothing in the repo relied on the broken proxies.

## Test plan

- [x] Two new regression tests (`test_getattr_only_proxies_stats` for both `Hypergraph` and `DiHypergraph`) covering: built-in stats still work, user-defined stats still work, every non-stat view method raises `AttributeError`, unknown attributes raise `AttributeError`
- [x] Full test suite passes (420 passed, 6 skipped)

### DIFF
--- a/tests/core/test_dihypergraph.py
+++ b/tests/core/test_dihypergraph.py
@@ -779,3 +779,44 @@ def test_cleanup(dihypergraph1):
     assert edge_in == xgi.dual_dict(node_out)
     assert edge_out == xgi.dual_dict(node_in)
     assert cleanDH["name"] == "test"
+
+
+def test_getattr_only_proxies_stats():
+    """DiHypergraph.__getattr__ must only delegate to actual stats, not arbitrary
+    view methods. See issue #408."""
+    DH = xgi.DiHypergraph([([1, 2], [3, 4]), ([2], [5, 6])])
+
+    # Built-in stats remain accessible
+    assert DH.degree(1) == 1
+    assert DH.in_degree(3) == 1
+    assert DH.out_degree(1) == 1
+    assert DH.size(0) == 4
+
+    # User-defined stats remain accessible
+    @xgi.dinodestat_func
+    def my_stat(net, bunch):
+        return {n: 7 for n in bunch}
+
+    assert all(v == 7 for v in DH.my_stat().values())
+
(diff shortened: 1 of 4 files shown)


nwlandry (approved):
Thanks @leotrs!

PR #737, opened 2026-06-04
docs: add Randomness page explaining seed semantics

leotrs:
## Summary

Adds a new \`randomness.rst\` page to the User Guides documenting how XGI's \`seed\` arguments behave. Per Max's suggestion on PR #724.

The page covers:
- Accepted types: \`None\`, \`int\`, \`numpy.random.Generator\`
- The four reproducibility cases (incl. the surprising one: \`seed=42\` ≠ \`seed=default_rng(42)\`)
- Why and when to pass a single \`Generator\` to a multi-step pipeline
- Links to NumPy's upstream rng docs

Also wires it into the User Guides toctree and adds a card for discoverability.

## Test plan

- [x] Docs-only change, full test suite still passes (418 passed, 6 skipped)
- [x] No tests added (docs page)

### DIFF
--- a/docs/source/randomness.rst
+++ b/docs/source/randomness.rst
@@ -1,0 +1,103 @@
+Randomness and reproducibility
+==============================
+
+Most stochastic functions in XGI accept a ``seed`` argument so that results can
+be made reproducible. This page documents what to pass and what to expect.
+
+Accepted types for ``seed``
+---------------------------
+
+XGI accepts three kinds of values:
+
+- ``None`` (default): a fresh, non-deterministic random state is used.
+- ``int``: a deterministic seed. Each call seeded with the same integer
+  produces identical output.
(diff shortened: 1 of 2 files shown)


maximelucas:
Looks great. Maybe add links to that page from randomness related pages? Like stochastic functions/generators.

kaiser-dan:
Okie it looks like there is an RuntimeError computing an `np.dot` in some unit test. It also seems to happen only intermittently, since #738 retested recently and did not fail.

I'm happy to merge this anyways, it is a documentation change and updating unit tests in this PR may be misleading.

leotrs:
@kaiser-dan thanks for the check and for the willingness to merge past the flaky test. I just pushed one more small commit to address @maximelucas's ask (cross-links from the generators and drawing landing pages to the new Randomness page). Once CI on that comes back, I'll merge.

PR #738, opened 2026-06-04
refactor: mark read_json/write_json as internal helpers (#697)

leotrs:
## Summary

Closes #697 (Option 1 path).

\`read_json\`/\`write_json\` had already been pulled from the top-level public API in #692 but still lived under public names and emitted \`DeprecationWarning\`s. The advice in those warnings ('use \`read_hif\`/\`write_hif\` instead') was misleading for the only remaining caller — \`xgi_data.py\` — because xgi-data still stores datasets in the legacy hypergraph-dict JSON format, not HIF, and the internal caller had nowhere else to go.

This PR completes the cleanup:
- Rename \`read_json\` → \`_read_json\`, \`write_json\` → \`_write_json\` so the private status is obvious.
- Drop the \`DeprecationWarning\` calls.
- \`__all__\` explicitly empty.
- Update the one internal call site in \`xgi/readwrite/xgi_data.py\`.
- Update \`See Also\` references in \`hypergraph_dict.py\` to point at the HIF readers/writers.
- Tests adjusted via import alias so the existing coverage is preserved.

Once xgi-data finishes migrating to HIF (tracked separately by Nich/Dan), these helpers can be deleted entirely.

## Test plan

- [x] Full test suite passes (418 passed, 6 skipped)
- [x] No deprecation warning emitted by xgi-data calls anymore

### DIFF
--- a/tests/readwrite/test_json.py
+++ b/tests/readwrite/test_json.py
@@ -5,7 +5,7 @@ import pytest
 
 import xgi
 from xgi.exception import XGIError
-from xgi.readwrite.json import read_json, write_json
+from xgi.readwrite.json import _read_json as read_json, _write_json as write_json
 
 json_string1 = """
 {
--- a/tests/readwrite/test_xgi_data.py
+++ b/tests/readwrite/test_xgi_data.py
@@ -7,7 +7,7 @@ import pytest
 
 from xgi import download_xgi_data, load_xgi_data
 from xgi.exception import XGIError
-from xgi.readwrite.json import read_json
+from xgi.readwrite.json import _read_json as read_json
 
 
 @pytest.mark.skipif(
--- a/xgi/convert/hypergraph_dict.py
+++ b/xgi/convert/hypergraph_dict.py
@@ -31,8 +31,8 @@ def to_hypergraph_dict(H):
 
     See Also
     --------
-    ~xgi.readwrite.json.read_json
-    ~xgi.readwrite.json.write_json
+    ~xgi.readwrite.hif.read_hif
+    ~xgi.readwrite.hif.write_hif
     """
     data = {}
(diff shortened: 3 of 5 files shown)


nwlandry:
I think (?) that I finished the update to HIF in XGI-DATA. Once that's merged, I think that these private functions could just be removed altogether. Unclear whether an update to this PR or a fresh PR is better, but I defer to you on this. (I suppose that we could rewrite the xgi_data functions to simplify the logic now that we've converted to HIF)

leotrs:
@nwlandry thanks, and good news that xgi-data is fully HIF now.

I'd keep this PR narrow (just the rename/internal-marking) and do the full removal + \`xgi_data.py\` simplification as a follow-up, for two reasons:

1. **Overlap with #756.** Your data-integrations PR already touches the same code path: it removes the \`read\` option from \`load_xgi_data\`, drops the old XGI JSON format from the load path, and reshapes \`request_json_from_url\` → \`request_from_url\`. Rebasing a "delete the JSON helpers + rewrite xgi_data on HIF" change on top of that is going to be much cleaner than doing both PRs in flight at the same time.

2. **This PR is already reviewed and green.** It does exactly what its title says. Expanding scope now would re-open the review and delay the small win.

Proposed sequencing:
- Merge #738 as-is (mark internal, no behavior change).
- #756 lands next (or first, either order works since they're compatible).
- Then a small follow-up PR: delete \`_read_json\`/\`_write_json\`, rewrite \`xgi_data.py\`'s local-cache path to use HIF read/write.

Happy to file an issue for step 3 if useful, so it doesn't get lost.

nwlandry:
No, that makes sense @leotrs! Thanks. I think it's ready to merge!

PR #741, opened 2026-06-29
feat: ship an agent skill alongside the xgi package

leotrs:
## Summary

Adds \`xgi/skill/SKILL.md\` as a SKILL.md-format agent skill (the open standard supported by Claude Code, Codex, Cursor, Gemini CLI, and skill-manager tools like \`agent-skills-sdk\`). Ships with the package via \`pyproject.toml\` package-data, so \`pip install xgi\` brings it along automatically and skill-aware tools discover it via the standard mechanism.

## What's in the skill

One file (\`SKILL.md\`) with YAML frontmatter (name, description, activation rule) followed by:

- **Basics** — the three core classes, how to build/inspect them
- **Where each capability lives** — a table mapping intent → submodule → online docs link
- **The stats interface** — XGI's main differentiator, the most-missed surface
- **Gotchas** — \`H.edges[idx]\` vs \`H.edges.members(0)\`, view-only methods that don't proxy through the hypergraph, rng semantics, deprecated I/O, exception types, class-specific differences
- **Idiomatic patterns** — build→compute→visualize, layout reuse, conversions, reproducibility, algorithms, cleanup

Plus thin scaffolding:
- \`README.md\` — meta doc for the team
- \`__init__.py\` — exposes \`SKILL_DIR\` / \`SKILL_FILE\` for programmatic discovery

Deliberately not included: a hand-maintained mirror of \`xgi.__all__\`. Agents can introspect at runtime, and the skill links into the online docs for the full API surface.

## Why

Hackathon feedback (June 2026) showed roughly half of XGI users used the library like NetworkX and never reached the stats / generators / algorithms / drawing surface. AI coding assistants amplify this discoverability problem: without an explicit skill, they default to the most NetworkX-like API and miss XGI's actual capabilities. This closes the gap on the agent side, in parallel with the human-facing docs work.

## Test plan

- [x] Files import correctly (\`from xgi import skill; skill.SKILL_FILE.exists()\`)
- [x] Full test suite passes (418 passed, 6 skipped)
- [ ] Future: CI check that runs the code snippets in \`SKILL.md\` and fails if they break (would catch documentation drift)


leotrs:
Following today's mtg: one way to address the discoverability problem is to tell AI agents how to use XGI.

nwlandry:
I like this, @leotrs! I think it is a useful summary for IRL users as well! Only minor nitpick: I assume the .beads folder should not be committed? Other than that, ready to merge IMO!

nwlandry (approved):

leotrs:
Good catch. Removed the entire `.beads/` folder in 765d572 and added `.beads/` to `.gitignore`. It's local task-tracking state that was committed by accident earlier — shouldn't have been in the repo at all.

PR #747, opened 2026-07-24
Make networks copyable and picklable on Python 3.14

### LINKED ISSUE #746: Bug - copy() doesn't work.
**Describe the bug**
I get a "TypeError: cannot pickle 'itertools.count' object" error.

**To Reproduce**
Run
```python
H1 = xgi.Hypergraph([[1,2,3]])
H2 = H1.copy()
```

**Expected behavior**
This should copy the hypergraph to a new object.

**Desktop:**
 - OS: Mac
 - Version: Sonoma 14.8.7 (23J520)
 - Python version 3.14
 - xgi version: 0.10.2

DMZ22:
Closes #746.

Confirming your diagnosis — and the blast radius is wider than the report. On CPython 3.14 this hits **all three network classes**, and all three of `copy()`, `deepcopy()` and `pickle`, not just `Hypergraph.copy()`:

| | `copy()` | `deepcopy()` | `pickle` |
|---|---|---|---|
| `Hypergraph` | ✗ | ✗ | ✗ |
| `DiHypergraph` | ✗ | ✗ | ✗ |
| `SimplicialComplex` | ✗ | ✗ | ✗ |

all with `TypeError: cannot pickle 'itertools.count' object`. Because `cleanup()`, `merge_duplicate_edges()` and `random_edge_shuffle()` all go through `copy()`, they fail too. **`pytest tests/core/` on `dev` was 17 failed / 125 passed under 3.14**; with this change it is 145 passed.

### Approach

*(Revised after review — see the discussion below. The first version kept `itertools.count` and worked around its non-peekability; @leotrs rightly pushed back, and @nwlandry agreed.)*

`_edge_uid` is now a plain `int`. That removes the problem rather than handling it:

- no helpers — the previous `uid_counter_value` / `uid_counter_from_value` are gone
- `__getstate__`, `__setstate__` and `copy()` are plain assignments
- `next(self._edge_uid)` became `uid = self._edge_uid; self._edge_uid += 1` at all seven call sites
- `update_uid_counter` sets the counter directly
- **`itertools.count` is no longer imported anywhere in the package**, which is the check that it is genuinely gone rather than relocated

Net effect on this PR: **58 insertions against 122 deletions.**

One deliberate deviation from the suggested `H._edge_uid = max(H._edge_uid, idx + 1)`: `idx` is not always a number. `add_edge` accepts string ids and tuples arrive from merged edges, so the `str` / `tuple` / `float(idx).is_integer()` guards are load-bearing and stay — only the body collapsed to a single assignment. `tests/core/test_simplicialcomplex.py` covers both (`["a", "b", 100, 101, 102, 103]`).

### Tests

A regression test per class asserts that `copy()`, `copy.copy()`, `deepcopy()` and a pickle round-trip all succeed, and that the clone's **next edge uid is preserved** — a clone must not reissue an existing id.

Two consequences of moving to an int, both worth flagging:

- five existing tests in `test_simplicialcomplex.py` asserted on the counter via `next(H._edge_uid)` and are now `H._edge_uid == N`. That the tests reached into the counter at all is arguably an argument for this design.
- my own `clone._edge_uid is not H._edge_uid` assertions were meaningless once the value is a small interned int, so they now assert the original's counter is unmoved after the clone advances its own.

Full suite: **430 passed / 6 skipped**. `tests/drawing/test_draw.py::test_issue_515` failed on one earlier run with a numpy overflow and passed on others, both with and without this change — it looks flaky rather than related. `isort` and `black` report no changes on the files I touched.

### DIFF
--- a/tests/core/test_dihypergraph.py
+++ b/tests/core/test_dihypergraph.py
@@ -830,4 +830,5 @@ def test_uid_counter_survives_copy_and_pickle(diedgelist1):
     for clone in (H.copy(), copy(H), deepcopy(H), pickle.loads(pickle.dumps(H))):
         clone.add_edge([[1], [2]])
         assert expected in clone.edges
-        assert clone._edge_uid is not H._edge_uid
+        # the clone owns its counter: advancing it must not move the original's
+        assert H._edge_uid == expected
--- a/tests/core/test_dihypergraph.py
+++ b/tests/core/test_dihypergraph.py
@@ -820,3 +820,14 @@ def test_getattr_only_proxies_stats():
     # Sanity: unknown attribute still raises
     with pytest.raises(AttributeError):
         DH.this_does_not_exist
+
+
+def test_uid_counter_survives_copy_and_pickle(diedgelist1):
+    """copy/deepcopy/pickle must work and must preserve the next edge uid (see #746)."""
+    H = xgi.DiHypergraph(diedgelist1)
+    expected = H.num_edges
+
+    for clone in (H.copy(), copy(H), deepcopy(H), pickle.loads(pickle.dumps(H))):
+        clone.add_edge([[1], [2]])
+        assert expected in clone.edges
+        assert clone._edge_uid is not H._edge_uid
--- a/tests/core/test_hypergraph.py
+++ b/tests/core/test_hypergraph.py
@@ -1092,18 +1092,9 @@ def test_uid_counter_survives_copy_and_pickle(edgelist1):
     for clone in (H.copy(), copy(H), deepcopy(H), pickle.loads(pickle.dumps(H))):
         clone.add_edge([1, 2])
         assert expected in clone.edges
-        assert clone._edge_uid is not H._edge_uid
+        # the clone owns its counter: advancing it must not move the original's
+        assert H._edge_uid == expected
 
     # the original is untouched by having been copied
     H.add_edge([1, 2])
     assert expected in H.edges
-
-
-def test_uid_counter_value_does_not_advance_the_counter(edgelist1):
-    H = xgi.Hypergraph(edgelist1)
-
-    first = xgi.utils.uid_counter_value(H)
-    second = xgi.utils.uid_counter_value(H)
-
-    assert first == second
-    assert next(H._edge_uid) == first
--- a/tests/core/test_hypergraph.py
+++ b/tests/core/test_hypergraph.py
@@ -1082,3 +1082,28 @@ def test_getattr_only_proxies_stats():
     # Sanity: unknown attribute still raises
     with pytest.raises(AttributeError):
         H.this_does_not_exist
+
+
+def test_uid_counter_survives_copy_and_pickle(edgelist1):
+    """copy/deepcopy/pickle must work and must preserve the next edge uid (see #746)."""
+    H = xgi.Hypergraph(edgelist1)
+    expected = H.num_edges
+
+    for clone in (H.copy(), copy(H), deepcopy(H), pickle.loads(pickle.dumps(H))):
+        clone.add_edge([1, 2])
+        assert expected in clone.edges
+        assert clone._edge_uid is not H._edge_uid
+
+    # the original is untouched by having been copied
+    H.add_edge([1, 2])
+    assert expected in H.edges
+
+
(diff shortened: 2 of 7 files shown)


leotrs:
Thanks for the deep diagnosis and the thorough tests. The scope expansion is useful too.

I'd like to push back on the mechanism though. The `value = next(...); self._edge_uid = count(value)` mechanism in `uid_counter_value` is clever, but it's clever precisely because it's working around `itertools.count` being non-peekable. 

I think we should just drop `itertools.count` and store `_edge_uid` as a plain int. That would:

- eliminate uid_counter_value / uid_counter_from_value entirely (no need to peek what's already visible),
- collapse __getstate__ / __setstate__ / copy() to plain assignments (no helpers, no dance),
- turn update_uid_counter into H._edge_uid = max(H._edge_uid, idx + 1),
- turn next(self._edge_uid) into uid = self._edge_uid; self._edge_uid += 1 at each of the ~7 call sites in hypergraph.py / dihypergraph.py / simplicialcomplex.py.

The whole PR gets smaller and the pickle/copy issue disappears rather than being worked around.

nwlandry:
Thanks for the contribution, @DMZ22! I think I agree with @leotrs here that we can sidestep the whole issue by making this an int and doing the steps he suggested.

DMZ22:
Agreed on all counts — you're right that the peek was clever in the wrong way. Pushed in `bebcea8`, done exactly as you laid it out:

- `_edge_uid` is a plain `int`, initialised to `0`
- `uid_counter_value` / `uid_counter_from_value` are gone
- `__getstate__` / `__setstate__` / `copy()` are plain assignments
- `next(self._edge_uid)` became `uid = self._edge_uid; self._edge_uid += 1` at all seven call sites
- `itertools.count` is no longer imported anywhere in the package, which is a good sign the workaround is genuinely gone rather than relocated

The diff is now **58 insertions against 122 deletions** — a net reduction of 64 lines, where the previous version added a helper module's worth of machinery.

One deviation from your sketch, and I'd rather flag it than bury it. I did not reduce `update_uid_counter` all the way to `H._edge_uid = max(H._edge_uid, idx + 1)`, because `idx` is not always a number:

```python
if (
    not isinstance(idx, str)
    and not isinstance(idx, tuple)
    and float(idx).is_integer()
    and H._edge_uid <= idx
):
    H._edge_uid = int(idx) + 1
```

The `str` and `tuple` guards are load-bearing — `add_edge` accepts string ids, and the existing comment notes that tuples arrive from merged edges. `max(0, "a" + 1)` would raise, and `tests/core/test_simplicialcomplex.py` covers both cases (`["a", "b", 100, 101, 102, 103]`). The `float(idx).is_integer()` check also still matters for the `np.int32`-style types the docstring mentions. So the body collapsed to a single assignment, but the guard stayed.

Two other things worth knowing:

Five existing tests in `test_simplicialcomplex.py` asserted on the counter via `next(H._edge_uid)`; they are now `H._edge_uid == N`. That the tests were reaching into the counter at all is arguably an argument for your design.

My own tests asserted `clone._edge_uid is not H._edge_uid` to prove the clone owned its counter. That check is meaningless for small ints, since they're interned, so it is replaced by asserting the original's counter is unmoved after the clone advances its own.

Full suite is 430 passed / 6 skipped. Note `tests/drawing/test_draw.py::test_issue_515` failed on one earlier run with a numpy overflow and passed on this one, both with and without my change — it looks flaky rather than related.

leotrs (approved):
Clean rewrite. Approving.

leotrs:
Merged, thanks for the contribution. One process note for future PRs.

We don't have formal guidelines for AI use in contributions yet, but we'll be putting some together soon. In the meantime, here's the informal expectation:

The comments and PR descriptions on this thread read as copy-pasted from an AI chat assistant. Using AI tools to help with contributions is fine (most of us do!) but please write the discussion parts in your own voice, and disclose AI assistance where it's substantial. Two reasons:

1. Reviewers rely on the discussion to understand your intent, weigh trade-offs, and ask follow-up questions. If we're actually talking to a model, that back-and-forth loses value.
2. It's harder to trust the diagnostic sections when we can't tell whether they came from you running the code or from a model asserting they did.

Formal guidelines coming; we'll post them when they're ready.

